PR #13, opened 2022-04-05
Run linting as part of gha workflow

ankitade:
* __->__ #13

Summary:
Run the precommit hooks on the files changed by the current PR. The action will just fail for now, so user will have to make the changes themselves. Will look into adding annotations of the changes next.

Test plan:
Added an unused import to cnn_lstm and the run failed.  https://github.com/facebookresearch/multimodal/runs/5879178261?check_suite_focus=true

Removed it and the action passed
https://github.com/facebookresearch/multimodal/runs/5879701467?check_suite_focus=true

Differential Revision: [D35490879](https://our.internmc.facebook.com/intern/diff/D35490879)


.github/workflows/pre-commit.yaml
```diff
@@ -1 +1,23 @@
...
+        with:
+          python-version: "3.8"
+      - name: Check out repos
+        uses: actions/checkout@v2
+      - name: Install lint utilities
+        run: |
+          pip install pre-commit
+          pre-commit install-hooks
+      - name: Run precommit hooks for linting
+        run: pre-commit run --all-files
```
langong347, line 23: As a routine check, do we need to run pre-commit on all files? should we just run changes only?

ankitade: this is only trial pr to see if workflow runs. will probably replace precommit with actual lints

ankitade: hm it might be noisy though, if some unrelated changes get merged, as this is not diff blocking yet. just like it failed for me. 
so the one used by torchrec is apparently deprecated as per readme so didnt go with that

langong347: Q: If the pre-commit CI fails, will it give the developer a warning page or auto format his/her code right away?

Sorry didn't notice the DEPRECATED notice.

ankitade:
@ankitade has imported this pull request.  If you are a Facebook employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D35490879).

.github/workflows/lint.yaml
```diff
@@ -0,0 +1,31 @@
...
+  pull_request:
+
+jobs:
+  lint:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Setup Python
+        uses: actions/setup-python@v2
+        with:
+          python-version: "3.8"
```
ebsmothers, line 15: nit: do we also want to run on 3.9? Maybe not necessary for this initial version, but something to keep in mind

ebsmothers (approved):

.github/workflows/lint.yaml
ankitade: hm idk running it x times will be slow right. let me see whats a "good practice" here

.github/workflows/pre-commit.yaml
ankitade: updated the summary. have a followup

PR #22, opened 2022-04-08
[TorchMultimodal][refactor] Support type checking with mypy

ebsmothers:
* (to be filled)

Summary:
Adding support for type checking using mypy.
Cleaning up a bunch of files to fix existing type check errors.
Note that for now we are only type checking non-FLAVA code inside torchmultimodal.
Also adding some instructions to CONTRIBUTING.md

Test plan:
```
mypy
Success: no issues found in 28 source files
```

```
python3 -m pytest -vv
...
42 passed, 2 skipped, 26 warnings in 22.87s
```

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -75,6 +75,24 @@ flake8 `git diff main --name-only`
 ufmt format `git diff main --name-only`
 ```
 
+### Type checking
+
+TorchMultimodal uses mypy for type checking. You can install mypy with the command
+
+```
+python3 -m pip install mypy
+```
+
+To perform type checking:
+
+```
+# on the whole repo
+mypy
+
+## only on modified files
+mypy `git diff main --name-only`
+```
+
 
 ### Unit Tests
 Please add unit tests for adding a new feature or a bug-fix. To run a specific test:
--- a/mypy.ini
+++ b/mypy.ini
@@ -1,0 +1,25 @@
+[mypy]
+
+files = torchmultimodal
+
+warn_unused_configs = True
+warn_redundant_casts = True
+show_error_codes = True
+show_column_numbers = True
+check_untyped_defs = True
+pretty = True
+strict_optional = False
+allow_redefinition = True
+namespace_packages = True
+install_types = True
+
+exclude = models/flava.py|modules/losses/flava.py
+
+[mypy-PIL.*]
+ignore_missing_imports = True
+
+[mypy-torchtext.*]
+ignore_missing_imports = True
+
+[mypy-torchvision.*]
+ignore_missing_imports = True
--- a/torchmultimodal/modules/encoders/clip_text_encoder.py
+++ b/torchmultimodal/modules/encoders/clip_text_encoder.py
@@ -87,7 +87,7 @@ class CLIPTextEncoder(nn.Module):
 
     def build_attention_mask(self):
         mask = torch.full((self.context_length, self.context_length), True).triu(1)
-        return mask.to(dtype=bool)
+        return mask.to(device=None, dtype=torch.bool)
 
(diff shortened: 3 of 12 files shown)


mypy.ini
```diff
@@ -0,0 +1,25 @@
...
+show_error_codes = True
+show_column_numbers = True
+check_untyped_defs = True
+pretty = True
+strict_optional = False
+allow_redefinition = True
+namespace_packages = True
+install_types = True
+
+exclude = models/flava.py|modules/losses/flava.py
```
langong347, line 16: Thanks! Mind filing a BE task on fixing FLAVA type annotation for tracking?

mypy.ini
```diff
@@ -0,0 +1,25 @@
...
+
+[mypy-PIL.*]
+ignore_missing_imports = True
```
langong347, line 19: Is this because pillow is installed with torchvision?
Wonder if we only need this for pytorch domain libs?

langong347 (approved):

torchmultimodal/modules/encoders/clip_text_encoder.py
```diff
@@ -87,7 +87,7 @@ def initialize_parameters(self):
 
     def build_attention_mask(self):
```
ankitade, line 88: no return annotation needed?

torchmultimodal/transforms/text_transforms.py
```diff
@@ -30,7 +30,7 @@ def __call__(
         self, l: Union[List[str], List[List[str]]]
     ) -> Union[List[int], List[List[int]]]:
         if isinstance(l[0], str):
-            return [int(x) for x in l]
+            return [int(x) for x in l]  # type: ignore
```
ankitade, line 33: why ignore for only this one?

CONTRIBUTING.md
```diff
@@ -75,6 +75,24 @@ flake8 `git diff main --name-only`
 ufmt format `git diff main --name-only`
 ```
 
+### Type checking
+
+TorchMultimodal uses mypy for type checking. You can install mypy with the command
+
+```
+python3 -m pip install mypy
```
ankitade, line 83: can u just do pip install like others?

mypy.ini
```diff
@@ -0,0 +1,25 @@
...
+
+[mypy-torchtext.*]
```
ankitade, line 21: why do we need these?

torchmultimodal/transforms/text_transforms.py
ebsmothers: There's an assumption here that every element of l is either str or List[str]. So we only check that the first one is a str. But then mypy complains because it thinks the other elements could still be List[str]. Personally I would rather ignore the error than iterate over the whole list just to make mypy happy

CONTRIBUTING.md
ebsmothers: Maybe, but this is from the [mypy docs](https://mypy.readthedocs.io/en/stable/getting_started.html), so personally I would stick with the officially-supported command

torchmultimodal/modules/encoders/clip_text_encoder.py
ebsmothers: It is probably due to [this property](https://mypy.readthedocs.io/en/stable/common_issues.html#no-errors-reported-for-obviously-wrong-code) of mypy where functions without any annotations are ignored completely. Not sure why this is, but we should see if there is some solution for this behavior.

ebsmothers:
@ebsmothers has imported this pull request.  If you are a Facebook employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D35584888).

PR #24, opened 2022-04-12
Run unit tests in gha

ankitade:
Summary:
As title

Test plan:
successful run on this PR itself (https://github.com/facebookresearch/multimodal/runs/6016844312?check_suite_focus=true)

### DIFF
--- a/.github/workflows/unit_test.yaml
+++ b/.github/workflows/unit_test.yaml
@@ -1,1 +1,32 @@
 name: unit test
+
+on:
+  push:
+    branches: [ main ]
+  pull_request:
+
+jobs:
+  unit_tests:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out repo
+        uses: actions/checkout@v2
+      - name: Setup conda env
+        uses: conda-incubator/setup-miniconda@v2
+        with:
+          miniconda-version: "latest"
+          activate-environment: test
+          python-version: "3.8"
+      - name: Install dependencies
+        shell: bash -l {0}
+        run: |
+          set -eux
+          conda activate test
+          conda install pytorch torchvision torchtext cpuonly -c pytorch-nightly
+          pip install -e ".[dev]"
+      - name: Run unit tests
+        shell: bash -l {0}
+        run: |
+          set -eux
+          conda activate test
+          pytest test -vv
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,1 +1,2 @@
 packaging
+iopath


.github/workflows/unit_test.yaml
```diff
@@ -1 +1,30 @@
...
+        with:
+          miniconda-version: "latest"
+          activate-environment: test
+          python-version: "3.8"
+      - name: Install dependencies
+        shell: bash -l {0}
+        run: |
+          conda activate test
+          conda install pytorch torchvision torchtext cpuonly -c pytorch-nightly
+          pip install -r requirements.txt
```
ebsmothers, line 25: Given the changes in #26, should we also add `pip install -r dev-requirements.txt` for pytest installation?

.github/workflows/unit_test.yaml
```diff
@@ -1 +1,30 @@
...
+      - name: Run unit tests
+        shell: bash -l {0}
+        run: |
+          conda activate test
+          pytest test
```
ebsmothers, line 30: Just wanna confirm, will this signal still pass with our skipped unit tests?

ankitade: its already skipping (see the tp link)

.github/workflows/unit_test.yaml
```diff
@@ -1 +1,30 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out repo
+        uses: actions/checkout@v2
+      - name: Setup conda env
+        uses: conda-incubator/setup-miniconda@v2
+        with:
+          miniconda-version: "latest"
+          activate-environment: test
+          python-version: "3.8"
```
langong347, line 19: Can we add 3.9 in CI too? (since we claimed supporting >=3.8 in readme).

.github/workflows/unit_test.yaml
langong347: nit: it's more explicit to use `pip3` than `pip` even though there's no ambiguity here given the conda env is created for python 3.8.

.github/workflows/unit_test.yaml
langong347: Can we increase the versobity `pytest -vv`?

langong347: Can we increase the verbosity `pytest -vv`?

.github/workflows/unit_test.yaml
langong347: So the tests are running without `python setup.py install` our own library?

.github/workflows/unit_test.yaml
langong347: Can we increase the verbosity `pytest -vv`?

.github/workflows/unit_test.yaml
langong347: > Given the changes in #26, should we also add `pip install -r dev-requirements.txt` for pytest installation?

I think it should be `pip install -e .[dev]` which installs our own library as well as the dev requirements (thus replacing `python setup.py develop`)

.github/workflows/unit_test.yaml
```diff
@@ -1 +1,30 @@
...
+      - name: Install dependencies
+        shell: bash -l {0}
+        run: |
+          conda activate test
```
langong347, line 23: How about adding `set -eux` for all job steps that contain bash commands? I've seen this a lot in automation (e.g., [pytorch](https://github.com/search?q=org%3Apytorch+%22set+-eux%22&type=Code)) which guarantees that we will immediately exist the shell when an error occurs. Since each job step generates a new process and a new shell, we will need to add it for all of them.

.github/workflows/unit_test.yaml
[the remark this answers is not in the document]
ankitade: yeah I saw that but is it necessary for things like test and linters once the build ci tests for all versions. its going to run x times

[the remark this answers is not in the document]
langong347: We need to build from source in the test CI so in this regard "build" and "test" are in the same workflow. 
There's also CI for nightly builds that generates wheels / binaries but the workflow is scheduled, e.g., midnight every day, and typically it includes running unit tests as well.
So I think the CI category is defined not around "build" and "test" but rather "diff time" vs. "nightlies".

For linter it's probably ok to test for just one version -- assuming the linter (and type checker) should remain relatively stable as to their rules.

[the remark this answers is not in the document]
ankitade: I thought if we have a separate workflow which builds from source for each python, os etc configuration on PRs, its ok to run tests (and linters etc) on a hardcoded configuration? 
or are you suggesting just using this one for testing builds also on PRs (not the recurring build and release workflow)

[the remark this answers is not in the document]
ankitade: lets do a followup as part of the build CI task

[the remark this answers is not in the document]
langong347: RE: separate build workflow --> this I was referring to in my comment as "CI for nightly builds". In addition, there may also be CI for release / stable builds. Those are dedicated build CIs which will test ALL combinations of supported python, os and etc.

For now since we don't do any binary or release builds, "build CI" is simply the build from source step in the test CI before the `pytest` step. So whatever CI matrix we plug in there will be applied to both "build" and "test" steps. 

RE: Avoid hardcoding the configuration: we can [read it as a JSON file](https://stackoverflow.com/questions/61919141/read-json-file-in-github-actions)

PR #26, opened 2022-04-13
Read dependencies from requirement files for setup

ankitade:
Summary:
- setup.py now has install_requires which reads from requirements.txt
- extra requires for dev specific dependencies coming from dev-requirements.txt (moved pytest to it)
- pytorch and torch* dependencies still need to be installed manually since we depend on nightly builds and there seems to be no principled way of getting them in setup.py

Test plan:
 - In a fresh conda env, followed Readme and tried to import flava. Checked no dev dependencies were installed in the env
 - In another env, followed updated contribution guide and ran pytest. Verified mypy, pytest and precommit were installed

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -33,7 +33,7 @@ outlined on that page and do not file a public issue.
 
 Same as in [README](README.md) with the exception of:
 ```
-python setup.py develop
+pip install -e ".[dev]"
 ```
 
 ## Development Process
@@ -54,7 +54,7 @@ We actively welcome your pull requests.
 TorchMultimodal uses pre-commit hooks to ensure style consistency and prevent common mistakes. Enable it by:
 
 ```
-pip install pre-commit && pre-commit install
+pre-commit install
 ```
 
 After this pre-commit hooks will be run before every commit.
@@ -63,7 +63,6 @@ Ideally, flake and ufmt should be run via pre-commit hooks.
 But if for some reason you want to run them separately follow this:
 
 ```
-pip install flake8==4.0.1 ufmt==1.3.0 black==21.4b2 usort==0.6.4
 flake8 (examples|test|torchmultimodal)
 ufmt format (examples|test|torchmultimodal)
 ```
@@ -71,19 +70,14 @@ ufmt format (examples|test|torchmultimodal)
 Alternatively, you can run on only those files you have modified, e.g.
 
 ```
+pip install flake8==4.0.1 ufmt==1.3.0 black==21.4b2 usort==0.6.4
 flake8 `git diff main --name-only`
 ufmt format `git diff main --name-only`
 ```
 
 ### Type checking
 
-TorchMultimodal uses mypy for type checking. You can install mypy with the command
-
-```
-python3 -m pip install mypy
-```
-
-To perform type checking:
+TorchMultimodal uses mypy for type checking. To perform type checking:
 
 ```
 # on the whole repo
--- a/README.md
+++ b/README.md
@@ -30,9 +30,8 @@ TorchMultimodal requires Python >= 3.8. The library can be installed with or wit
     ```
     git clone --recursive https://github.com/facebookresearch/multimodal.git torchmultimodal
     cd torchmultimodal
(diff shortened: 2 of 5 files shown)


dev-requirements.txt
```diff
@@ -0,0 +1,3 @@
+pytest
+mypy
+pre-commit
```
ebsmothers, line 3: Should we also add flake8/ufmt/black/usort to dev requirements?

CONTRIBUTING.md
```diff
@@ -33,7 +33,7 @@ outlined on that page and do not file a public issue.
 
 Same as in [README](README.md) with the exception of:
 ```
-python setup.py develop
+pip install -e ".[dev]"
```
ebsmothers, line 36: We may also want to change the instructions in [README.md](https://github.com/facebookresearch/multimodal/blob/main/README.md#building-from-source)

CONTRIBUTING.md
```diff
@@ -77,13 +76,7 @@ ufmt format `git diff main --name-only`
...
 ### Type checking
 
-TorchMultimodal uses mypy for type checking. You can install mypy with the command
-
-```
-python3 -m pip install mypy
-```
-
-To perform type checking:
+TorchMultimodal uses mypy for type checking.To perform type checking:
```
langong347, line 79: Adding a space here?
`TorchMultimodal uses mypy for type checking.(space)To perform type checking:`

README.md
```diff
@@ -30,7 +30,6 @@ TorchMultimodal requires Python >= 3.8. The library can be installed with or wit
     ```
     git clone --recursive https://github.com/facebookresearch/multimodal.git torchmultimodal
     cd torchmultimodal
-    pip install -r requirements.txt
 
     python setup.py install
```
langong347, line 34: Since we now read requirements into setuptools, we can `pip install torchmultimodal`?

langong347:
We should consider adding version numbers in requirements file at least for packages where reproducible behaviors are important.

dev-requirements.txt
ankitade: it should come from precommit right?

ankitade:
this is pretty much reproducing the manual readme right now so idk which ones we want to pin

dev-requirements.txt
ebsmothers: Yeah pre-commit will run these regardless, I am thinking more if developers want to run the checks themselves. But no strong preference here

langong347: pre-commit will run more than just linting and autoformatting including license header, AST and etc.
If the user wants to run linting/formatter separately, they will need to install flake8/umft/black/usort themselves.
The versions are also important to ensure consistency with fb internal. We documented the versions in contributing guide some time ago.

CONTRIBUTING.md
```diff
@@ -33,7 +33,7 @@ outlined on that page and do not file a public issue.
 
 Same as in [README](README.md) with the exception of:
 ```
-python setup.py develop
+pip install -e .
```
langong347, line 36: Does this now replace the two-step approach `python setup.py [mode]` + `pip install -r [mode]-requirements`?

Should we consider sticking with one flavor in our documentation (in README.md it still tells the users to do `python setup.py install`)?

langong347:
> this is pretty much reproducing the manual readme right now so idk which ones we want to pin

Versions are added in requirements files. All packages should have versions to guarantee reproducibility across developers and (especially) in CI.

But I've also seen repos not being strict with versions (i.e., only package names in requirements). That should be fine until we become more sensitive to versions.

For now I'm just referring to the linter/formatter (see my comments earlier) as those are known issues.

CONTRIBUTING.md
ankitade: strange, i meant to change it in requirements.txt and not in contributing but maybe i wasnt paying attention

dev-requirements.txt
ankitade: just added back the install instructions for ufmt etc in contributing

ebsmothers (approved):

PR #28, opened 2022-04-17
Run mypy as part of gha

ankitade:
Summary:
- Run mypy on all files on PRs. I tried doing it for files changes by PR but then it starts looking at the config files also :(
- Add some ignores based for iopath changes

Test plan:
- made a typing error in cnn_lstm -> type checkinng workflow fails https://github.com/facebookresearch/multimodal/runs/6068075411?check_suite_focus=true
- Fixed the error -> workflow passes on this PR https://github.com/facebookresearch/multimodal/runs/6068955875?check_suite_focus=true

### DIFF
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -10,7 +10,6 @@ jobs:
         uses: actions/setup-python@v2
         with:
           python-version: "3.8"
-
       - name: Check out repos
         uses: actions/checkout@v2
         with:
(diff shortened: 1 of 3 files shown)


.github/workflows/lint.yaml
```diff
@@ -26,3 +25,27 @@ jobs:
           output: ' '
       - name: Lint modified files
         run: pre-commit run --files ${{ steps.file_changes.outputs.files }}
+  type_checking:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out repo
+        uses: actions/checkout@v2
+      - name: Setup conda env
```
langong347, line 33: N00b Q: do we need to install the library in order to run mypy? Is it possible to just check out the python interpreter, source code and run on the PR changes (like in linter)?

ankitade: it complains about imports if you don't do installation of torch or might happen with other dependencies (repros locally too)

mypy.ini
```diff
@@ -24,3 +24,9 @@ ignore_missing_imports = True
 
 [mypy-torchvision.*]
 ignore_missing_imports = True
+
+[mypy-torchmultimodal.fb.*]
```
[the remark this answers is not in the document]
ankitade: technically this is for all fb and next one for file_io

langong347:
LGTM.

langong347 (approved):

PR #29, opened 2022-04-26
Added quantisation layer, commitment loss, and associated unit tests

RdoubleA:
### Summary:
Vector Quantized Variational Autoencoder (VQVAE) is widely used to generate multimodal data (audio, image, or video samples). A well-known model implementing this is OpenAI's DALL-E model, which can generate images from text captions. The core component of the VQVAE is the quantisation layer, which is a discrete embedding of vectors that codifies the outputs of an encoder. Here, this layer is implemented as its own module, along with commitment loss which is used to fine-tune the encoder.

### Test plan:
Unit tests were created and added in this PR. Testing was done with PyTest.

`pytest -vv test/modules/losses/test_commitment.py`

image

`pytest -vv test/modules/layers/test_quantisation.py`

image


torchmultimodal/modules/losses/vqvae.py
```diff
@@ -0,0 +1,28 @@
...
+    def __init__(self, commitment_cost: float = 1.0, **kwargs: Any):
+        super().__init__()
+        self.mse_loss = nn.MSELoss(reduction="mean")
+        self.commitment_cost = commitment_cost
+
+    def forward(self, quantised: Tensor, encoded: Tensor):
+        # Quantised vectors must be detached because commitment loss only lets gradient flow through encoder output
+        loss = self.mse_loss(quantised.detach(), encoded) * self.commitment_cost
+
+        return CommitmentLossOutput(loss=loss)
```
ankitade, line 28: you could just return the Tensor here since its only one return value

torchmultimodal/modules/layers/quantisation.py
```diff
@@ -0,0 +1,61 @@
...
+    """
+
+    def __init__(self, num_embeddings, embedding_dim):
+        super().__init__()
+        self.embedding_dim = embedding_dim
+        self.num_embeddings = num_embeddings
+
+        self.embedding = nn.Embedding(self.num_embeddings, self.embedding_dim)
+        self.embedding.weight.data.normal_()
+        self.quantised_vectors = None
```
ankitade, line 24: does this need to be a class variable?

torchmultimodal/modules/layers/quantisation.py
```diff
@@ -0,0 +1,61 @@
...
+        )
+        encodings.scatter_(1, encoding_indices, 1)
+
+        # Quantise
+        quantised_flattened = torch.matmul(encodings, self.embedding.weight)
+        self.quantised_vectors = quantised_flattened
+
+        return quantised_flattened
+
+    def get_quantised_vectors(self):
```
ankitade, line 55: looks like this is for some sort of caching? would be better to leave this out of the core module

torchmultimodal/modules/layers/quantisation.py
```diff
@@ -0,0 +1,61 @@
...
+from torch import nn
+
+
+class Quantisation(nn.Module):
+    """
+    Embedding layer that takes in a collection of flattened vectors and finds closest embedding vectors
+    to each flattened vector and outputs those selected embedding vectors. Also known as vector quantisation.
+    """
+
+    def __init__(self, num_embeddings, embedding_dim):
```
ankitade, line 17: some type annotations

torchmultimodal/modules/layers/quantisation.py
```diff
@@ -0,0 +1,61 @@
...
+
+        # Encoding - select closest embedding vectors
+        encoding_indices = torch.argmin(distances, dim=1).unsqueeze(1)
+        encodings = torch.zeros(
+            encoding_indices.shape[0], self.num_embeddings, device=x_flat.device
+        )
+        encodings.scatter_(1, encoding_indices, 1)
+
+        # Quantise
+        quantised_flattened = torch.matmul(encodings, self.embedding.weight)
```
ankitade, line 50: is this same as doing embedding look up of the arg min indices? if yes, writing this as self.embedding(..) might be more clear

torchmultimodal/modules/losses/vqvae.py
```diff
@@ -0,0 +1,28 @@
...
+
+
+class CommitmentLoss(nn.Module):
+    def __init__(self, commitment_cost: float = 1.0, **kwargs: Any):
+        super().__init__()
+        self.mse_loss = nn.MSELoss(reduction="mean")
+        self.commitment_cost = commitment_cost
+
+    def forward(self, quantised: Tensor, encoded: Tensor):
+        # Quantised vectors must be detached because commitment loss only lets gradient flow through encoder output
```
ankitade, line 25: noob q, how are the embeddings learnt ? is there another loss fn?

test/modules/layers/test_quantisation.py
```diff
@@ -0,0 +1,119 @@
...
+        self.encoded = torch.randn((2, self.embedding_dim, 3, 3))
+        self.embedding_weights = torch.randn((self.num_embeddings, self.embedding_dim))
+
+    def test_quantised_output(self):
+        vq = Quantisation(
+            num_embeddings=self.num_embeddings, embedding_dim=self.embedding_dim
+        )
+        vq.embedding = nn.Embedding.from_pretrained(self.embedding_weights)
+        x, permuted_shape = flatten_to_channel_vectors(self.encoded, 1)
+        output = vq(x)
```
ankitade, line 37: why do we need these processing? cant we hard code the tensor format to begin with?

ankitade (commented):
overall good start. some questions / suggestions

torchmultimodal/modules/layers/quantisation.py
```diff
@@ -0,0 +1,61 @@
...
+        # Calculate distances from each encoder output vector to each embedding vector, ||x - emb||^2
+        w_t = self.embedding.weight.t()
+        distances = (
+            torch.sum(x_flat ** 2, dim=1, keepdim=True)
+            - 2 * torch.matmul(x_flat, w_t)
+            + torch.sum(w_t ** 2, dim=0, keepdim=True)
+        )
+
+        # Encoding - select closest embedding vectors
+        encoding_indices = torch.argmin(distances, dim=1).unsqueeze(1)
```
langong347, line 43: Not sure if I understand the steps below this line. I suppose here we want to:
- find the embeddings that are closest to the input tensors (aka "quantization")
- find the corresponding indices (may not seem obvious why we need this but a downstream module may use them for other purposes)

In addition, the embeddings need to be "moved" towards the encoder outputs (similar to KL divergence loss in Kingma's VAE paper in the continuous space). I think OpenAI's [implementation](https://github.com/openai/jukebox/blob/08efbbc1d4ed1a3cef96e08a931944c8b4d63bb3/jukebox/vqvae/bottleneck.py#L173) is a bit hand-waving that omits the burden to actually compute this loss but instead merely "update" the embeddings by the delta with the encoder output. But it's ok to stick with openAI for this exercise. And if we add this logic, please also include in the docstring description.

torchmultimodal/modules/layers/quantisation.py
langong347: Not caching per se but just allow other modules to retrieve the (quantized) embedding vectors to compute commit loss at a later time.

This design is to avoid returning the loss directly by `forward`.

torchmultimodal/modules/losses/vqvae.py
```diff
@@ -0,0 +1,28 @@
...
+
+from torch import nn, Tensor
+
+
+@dataclass
+class CommitmentLossOutput:
+    loss: Tensor
+
+
+class CommitmentLoss(nn.Module):
```
langong347, line 18: Add docstring for this class.

torchmultimodal/modules/losses/vqvae.py
langong347: The loss consists of 3 parts: 
- reconstruction (delta between input and output tensors)
- VQ objective loss (to move the quantized vectors closer to output of the encoder)
- commitment loss (to regularize the quantized vectors)

Eq 3) in the [paper](https://papers.nips.cc/paper/2017/file/7a98af17e63a0ac09ce2e96d03992fbc-Paper.pdf)

test/modules/losses/test_commitment.py
```diff
@@ -0,0 +1,35 @@
...
+
+
+class TestCommitment(unittest.TestCase):
+    """
+    Test the Commitment Loss
+    """
+
+    def setUp(self):
+        torch.set_printoptions(precision=10)
+        torch.manual_seed(4)
```
langong347, line 20: The same here. Use `test_utils`.

test/modules/losses/test_commitment.py
```diff
@@ -0,0 +1,35 @@
...
+        self.quantized = torch.randn((2, 3))
+        self.encoded = torch.randn((2, 3))
+
+    def test_loss_value(self):
+        commitment = CommitmentLoss()
+        loss = commitment(self.quantized, self.encoded)
+
+        actual = loss.loss.item()
+        expected = 1.2070025206
```
langong347, line 29: Personally I prefer manual inputs and computation of the expected for simple logic like L2 norm. Otherwise, it just becomes a regression test (checks if future values deviate from the current snapshot) rather than unit test (which checks for correctness).

test/modules/layers/test_quantisation.py
langong347: The quantization layer assumes the input tensor is already "flattened" into a shape where the last dimension is the number of embeddings (which is all what matters here, more details see comments in the module). This allows generalization to all modalities.

torchmultimodal/utils/preprocess.py
```diff
@@ -0,0 +1,48 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+
+#
+# preprocess.py
```
langong347, line 9: Since the logic here is made modality agnostic and the only consumer is the quantization layer (for now), I'm wondering if we need a separate module for them. Alternatively, we can also consider `torchmultimodal/transforms/`, or the kitchen sink `torchmultimodal/utils/common.py`.

@ankitade Any thoughts here?

test/modules/layers/test_quantisation.py
```diff
@@ -0,0 +1,119 @@
...
+    """
+    Test the Quantisation class
+    """
+
+    def setUp(self):
+        torch.set_printoptions(precision=10)
+        torch.manual_seed(4)
+        self.num_embeddings = 4
+        self.embedding_dim = 5
+        self.encoded = torch.randn((2, self.embedding_dim, 3, 3))
```
langong347, line 28: The same comment here about using manual calculation, e.g., how do I know we are finding the closest matching embeddings by looking at those outputs here?

torchmultimodal/modules/losses/vqvae.py
RdoubleA: Yup, the commitment loss trains the encoder and the VQ objective loss (I've also seen it called codebook loss) is also very similar except it only allows gradient flow to the embedding to train the embedding weights. 

According to the paper, both are used instead of just one combined loss because technically there's no gradient for the argmin of distance to embedding, so they use a straight through estimator which blocks gradients for the embedding.

Which reminds me that I forgot to do the "straight through" estimator... will fix that

torchmultimodal/utils/preprocess.py
```diff
@@ -0,0 +1,48 @@
...
+
+    return x_flat, permuted_shape
+
+
+def reshape_from_channel_vectors(x_flat, permuted_shape, orig_channel_dim):
+    """
+    The inverse of flatten_to_channel_vectors
+    """
+    # Reshape flattened vectors to permuted shape
+    x_out = x_flat.view(permuted_shape)
```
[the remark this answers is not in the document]
RdoubleA: yes exactly, we need the permuted shape instead of original for that reason. this makes this functions a bit awkward to use in isolation... so may want to revisit this later.

test/modules/losses/test_commitment.py
RdoubleA: not sure what you mean by manual, do you mean using hand-picked values instead of random values with a fixed rng seed?

torchmultimodal/modules/layers/quantisation.py
```diff
@@ -0,0 +1,61 @@
...
+            x_shape[-1] == self.embedding_dim
+        ), f"Expected {x_shape[-1]} to be embedding size of {self.embedding_dim}"
+
+        # Calculate distances from each encoder output vector to each embedding vector, ||x - emb||^2
+        w_t = self.embedding.weight.t()
+        distances = (
+            torch.sum(x_flat ** 2, dim=1, keepdim=True)
+            - 2 * torch.matmul(x_flat, w_t)
+            + torch.sum(w_t ** 2, dim=0, keepdim=True)
+        )
```
ebsmothers, line 40: Can you do this instead? (Possibly modulo a transpose of w_t)

```suggestion
        distances = torch.cdist(x_flat,w_t,p=2.0) ** 2
```

RdoubleA: Yep I just realized cdist does the same thing

torchmultimodal/utils/preprocess.py
RdoubleA: @langong347 thoughts on point 3?

test/modules/layers/test_quantisation.py
RdoubleA: can we assume that all modalities' input data will have the dimensions batch x channel x n dims? I understood that to be consistent across pytorch

torchmultimodal/utils/preprocess.py
langong347: Agreed on (3) we can move them under `Quantisation` class.

torchmultimodal/modules/layers/quantisation.py
```diff
@@ -0,0 +1,61 @@
...
+    to each flattened vector and outputs those selected embedding vectors. Also known as vector quantisation.
+    """
+
+    def __init__(self, num_embeddings, embedding_dim):
+        super().__init__()
+        self.embedding_dim = embedding_dim
+        self.num_embeddings = num_embeddings
+
+        self.embedding = nn.Embedding(self.num_embeddings, self.embedding_dim)
+        self.embedding.weight.data.normal_()
```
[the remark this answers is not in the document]
RdoubleA: good point, I think in the paper they define a [discrete uniform distribution](https://en.wikipedia.org/wiki/Discrete_uniform_distribution) over the embedding vectors. In this semi-official implementation in [Google Colab](https://colab.research.google.com/github/zalandoresearch/pytorch-vq-vae/blob/master/vq-vae.ipynb) they used `self._embedding.weight.data.uniform_(-1/self._num_embeddings, 1/self._num_embeddings)`. I will do the same.

torchmultimodal/utils/preprocess.py
RdoubleA: also wanted to add that if the flattening and unflattening is done in the quantisation layer it's more straightforward because I don't have to feed the `permuted_shape` to unflatten it anymore

torchmultimodal/modules/layers/quantisation.py
ankitade: We can follow up offline (no need to block this PR). But at first glance, the concern is how does the caller of this function know the quantized embeddings are for which input?

torchmultimodal/utils/preprocess.py
ankitade: 3 sounds good. forward will call these utils iiuc

test/modules/layers/test_quantisation.py
langong347: How about text data where "channel" is embedding dim? Is it always the second in the shape array? 
The question about input tensor shape is orthogonal to our delegating the flattening / unflattening to separate functions anyways.

torchmultimodal/modules/layers/quantisation.py
langong347: The "input" will be the current output from the encoder when you string encoder, quantized layer and decoder together to become the full vqvae model (or a lightning module). Basically, the commit loss is computed "on-the-fly" --- let me know if it still doesn't make sense.

torchmultimodal/modules/layers/quantisation.py
```diff
@@ -0,0 +1,73 @@
...
+
+        self.embedding = nn.Embedding(self.num_embeddings, self.embedding_dim)
+        self.embedding.weight.data.uniform_(
+            -1 / self.num_embeddings, 1 / self.num_embeddings
+        )
+        self.quantised_vectors = None
+
+    def forward(self, x: Tensor):
+        # Rearrange from batch x channel x n dims to batch x n dims x channel
+        new_dims = (0,) + tuple(range(2, len(x.shape))) + (1,)
```
langong347, line 36: I'd still separate out the routines about data processing to be separate functions for this class. It will also make unit testing easier. Say, if something goes wrong with `forward`, how do I know if it's coming from data processing or the quantization step?

langong347 (changes requested):

test/modules/layers/test_quantisation.py
langong347: That's fine as long as we keep throwing the exception inside `forward` that verifies last dim of the permuted shape is equal to embedding dim of the quantized vector. My concern is that we don't always have control over what happens inside the encoder despite the convention.

torchmultimodal/modules/layers/quantisation.py
```diff
@@ -0,0 +1,73 @@
...
+        # Quantise
+        quantised_permuted = self.embedding(encoding_indices).view(permuted_shape)
+
+        # Straight through estimator
+        quantised_permuted = x_permuted + (quantised_permuted - x_permuted).detach()
+
+        # Rearrange back to batch x channel x n dims
+        old_dims = (0,) + (len(x.shape) - 1,) + tuple(range(1, len(x.shape) - 1))
+        quantised = quantised_permuted.permute(old_dims).contiguous()
+        self.quantised_vectors = quantised
```
langong347, line 63: >the quantised vectors are used for commitment, and that's the chosen embedding vectors based on the encoder output. it is different from the embedding itself, which can only be updated via VQ loss / codebook loss

In the forward pass, it doesn't matter. But the commitment loss shouldn't compute the gradient against `quantised_vectors` and that's why the latter needs to be "plucked" out of the graph (or `.detached`). 

Here you are caching it after it has been updated.

https://github.com/openai/jukebox/blob/08efbbc1d4ed1a3cef96e08a931944c8b4d63bb3/jukebox/vqvae/bottleneck.py#L170

langong347 (changes requested):

torchmultimodal/modules/layers/quantisation.py
```diff
@@ -0,0 +1,73 @@
...
+        ), f"Expected {x_flat.shape[-1]} to be embedding size of {self.embedding_dim}"
+
+        # Calculate distances from each encoder output vector to each embedding vector, ||x - emb||^2
+        distances = torch.cdist(x_flat, self.embedding.weight, p=2.0) ** 2
+
+        # Encoding - select closest embedding vectors
+        encoding_indices = torch.argmin(distances, dim=1)
+
+        # Quantise
+        quantised_permuted = self.embedding(encoding_indices).view(permuted_shape)
```
langong347, line 55: Just to be sure here: `encoding_indices` is equal to `x_flat.shape[0]` due to the pair-wise `cdist` such that we can reshape with `permuted_shape`?

torchmultimodal/modules/layers/quantisation.py
```diff
@@ -0,0 +1,73 @@
...
+        # Rearrange from batch x channel x n dims to batch x n dims x channel
+        new_dims = (0,) + tuple(range(2, len(x.shape))) + (1,)
+        x_permuted = x.permute(new_dims).contiguous()
+        permuted_shape = x_permuted.shape
+
+        # Flatten input
+        x_flat = x_permuted.view(-1, permuted_shape[-1])
+        # channel dimension should be embedding dim so that each element in encoder
+        # output volume gets associated with single embedding vector
+        assert (
```
langong347, line 44: add a unit test for exceptions too.

torchmultimodal/modules/layers/quantisation.py
```diff
@@ -0,0 +1,73 @@
...
+        embedding_dim (int): the dimensionality of the embedding vectors
+    """
+
+    def __init__(self, num_embeddings: int, embedding_dim: int):
+        super().__init__()
+        self.embedding_dim = embedding_dim
+        self.num_embeddings = num_embeddings
+
+        self.embedding = nn.Embedding(self.num_embeddings, self.embedding_dim)
+        self.embedding.weight.data.uniform_(
```
langong347, line 29: The original [paper](https://papers.nips.cc/paper/2017/file/7a98af17e63a0ac09ce2e96d03992fbc-Paper.pdf) assumes uniform distribution to be the prior for the latent vectors `p(z)` such that there's equal probability for each latent vector `z`. The posterior `p(z|x)` is deterministic by way of minimizing the L2 distance.

Constraining the values within `z` to be uniform is not prescribed per se but (I guess) it probably makes the convergence with the encoder output easier.

cc @ebsmothers

langong347:
Please include in the Test Plan:
- the exact command you used to run unit tests regarding your changes
- testing results

torchmultimodal/modules/layers/quantisation.py
RdoubleA: I've seen this in at least three separate implementations of the model. It probably makes more sense than initializing with normal distribution anyway.

torchmultimodal/modules/layers/quantisation.py
RdoubleA: right, I was a bit hesitant because of keeping track of the permuted shape, but simplicity in unit testing is probably more important

torchmultimodal/modules/layers/quantisation.py
RdoubleA: yes that is correct

torchmultimodal/modules/layers/quantisation.py
RdoubleA: yes I believe this aligns with the openai code both on the straight through estimator and the commitment loss

torchmultimodal/modules/layers/quantisation.py
RdoubleA: then I will keep the class variable for now

torchmultimodal/modules/layers/quantisation.py
```diff
@@ -5,25 +5,33 @@
...
-from torch import nn
+from torch import nn, Tensor
 
 
 class Quantisation(nn.Module):
-    """
-    Embedding layer that takes in a collection of flattened vectors and finds closest embedding vectors
-    to each flattened vector and outputs those selected embedding vectors. Also known as vector quantisation.
+    """Quantisation provides an embedding layer that takes in a collection of flattened vectors, usually the
+    output of an encoder architecture, and performs a nearest-neighbor lookup in the embedding space.
```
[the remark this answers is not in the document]
RdoubleA: not anymore, the separate process functions in forward should handle that. will update docstring

torchmultimodal/modules/layers/quantisation.py
ebsmothers: Noob question: where do we intend to get the second term in the paper's equation (3) from (the VQ objective)? Seems as written it could come from `CommitmentLoss` with the args reversed and `commitment_cost=1.0`, but based on the naming in that module I'm inferring that we don't intend to use it that way.

RdoubleA: could either implement as two separate loss modules, commitment and codebook, or one single module, vqvae loss. can't imagine a scenario you would want to use them separately, and beta / commitment_cost seems to only apply to commitment loss

langong347: To prevent the commit loss from taking a gradient against the quantised vector, the latter needs to be cached **before** it's updated by the straight through step:

```
quantised_flat = self.embedding(encoding_indices)
self.quantised_vector = self._postprocess(quantise_flat.detach(), permuted_shape)
quantised_flat = x_flat + (quantised_flat - x_flat).detach()

return quantised_flat
```

langong347:
Sorry I probably didn't make the purpose of Test Plan clear: You can just run tests related to your changes, e.g., `pytest -vv test.modules.layers.test_quantisation.py` something like this instead of executing all tests from the library.

test/modules/layers/test_quantisation.py
```diff
@@ -105,8 +78,47 @@ def test_quantised_shape(self):
         vq.embedding = nn.Embedding.from_pretrained(self.embedding_weights)
         output = vq(self.encoded)
         actual = torch.tensor(output.shape)
-        expected = torch.tensor([2, 5, 3, 3])
+        expected = torch.tensor([2, 5, 3])
 
         assert torch.equal(
```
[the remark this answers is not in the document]
RdoubleA: so I can remove this test then?

test/modules/layers/test_quantisation.py
```diff
@@ -105,8 +78,47 @@ def test_quantised_shape(self):
         vq.embedding = nn.Embedding.from_pretrained(self.embedding_weights)
         output = vq(self.encoded)
```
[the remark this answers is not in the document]
RdoubleA: I thought that pytest doesn't keep any variables across different tests, so I repeated the forward pass. or maybe I need to keep it in a class variable? also not sure if I even need `test_quantised_shape` since you mentioned the shapes are already compared in `test_quantised_output`

torchmultimodal/modules/layers/quantisation.py
```diff
@@ -5,19 +5,22 @@
...
-    images, videos, and audio data.
+    Vector quantisation was introduced in Oord et al. 2017 (https://arxiv.org/pdf/1711.00937.pdf)
+    to generate high-fidelity images, videos, and audio data.
 
     Args:
         num_embeddings (int): the number of vectors in the embedding space
         embedding_dim (int): the dimensionality of the embedding vectors
+
+    Inputs:
+        x (Tensor): Tensor containing a batch of encoder outputs. Expects dimensions to be batch x channel x n dims.
```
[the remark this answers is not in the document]
langong347: No, docstring length is not captured by the linter. Even within fbcode.

langong347 (approved):
Thanks for the nice work! Going through the threads in this PR I can see a few points to follow up:

1. Learning of the codebook needs to be implemented by adding VQ loss or as moving average of the encoder output.
2. Re-design of the commitment loss (or vqvae loss as a whole). 

Let me know your thoughts about those observation and if anything that might be missing.

RdoubleA:
> Thanks for the nice work! Going through the threads in this PR I can see a few points to follow up:
> 
> 1. Learning of the codebook needs to be implemented by adding VQ loss or as moving average of the encoder output.
> 2. Re-design of the commitment loss (or vqvae loss as a whole) and clean up API of the quantization layer.
> 
> Also I have a question for you:
> 
> * The commitment loss is computed with the quantized vectors detached [before](https://github.com/openai/jukebox/blob/08efbbc1d4ed1a3cef96e08a931944c8b4d63bb3/jukebox/vqvae/bottleneck.py#L170) returned by `forward` as in OpenAI and the Google example. In you current implementation, the quantized vectors are detached after (inside `CommitmentLoss`). Do you think the two steps are commutative?
> 
> Let me know your thoughts about those observation and if anything that might be missing.
> 
> Next step: click "import to fbsource" button below and wait for the internal CI to finish, fix any issues as needed.

I think as long as `quantized` is detached before computing loss then the gradients should be calculated correctly. Whether the loss is computed inside `forward` as in OpenAI's implementation or computed outside in a separate loss module as in `CommitmentLoss` the same tensors are being used, so it should have the same result.

PR #39, opened 2022-05-09
Refactored tensor equal assertion test into common utility

RdoubleA:
Summary:
The use of `torch.testing.assert_close(actual, expected)` was a prevalent pattern throughout the codebase, each having to create its own assertion failure message. This was consolidated into a common utility function in `test/test_utils.py` such that this assertion is only defined in a single location and all current/future unit tests simply have to import this to test if two tensors are equal without having to write the same assertion failure message.

Test plan:
Since this impacts potentially all unit tests, I ran the entire test suite with `pytest -vv`.

### DIFF
--- a/test/test_utils.py
+++ b/test/test_utils.py
@@ -71,9 +71,13 @@ def get_asset_path(file_name: str) -> str:
     return str(_ASSET_DIR.joinpath(file_name))
 
 
-def assert_tensors_equal(actual: Tensor, expected: Tensor):
+def assert_expected(
+    actual: Tensor, expected: Tensor, rtol: float = None, atol: float = None
+):
     torch.testing.assert_close(
         actual,
         expected,
+        rtol=rtol,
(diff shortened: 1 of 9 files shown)


test/test_utils.py
```diff
@@ -68,3 +69,11 @@ def set_rng_seed(seed):
 def get_asset_path(file_name: str) -> str:
     """Get the path to the file under assets directory."""
     return str(_ASSET_DIR.joinpath(file_name))
+
+
+def assert_tensors_equal(actual: Tensor, expected: Tensor):
```
langong347, line 74: Allow optional arguments so the caller can access `atol` `rtol`.
Just name it `assert_expected` as we may want to check both cases of strict equality and closeness.

For example: https://github.com/pytorch/vision/blob/main/test/test_models.py#L97

langong347 (approved):

PR #41, opened 2022-05-12
Added API to retrieve indices from the quantization layer

RdoubleA:
Summary:
Exposes the indices of the embedding vectors that were chosen in the quantisation layer during nearest neighbor lookup. These indices can be used in downstream tasks, such as a transformer decoder for cross-modal retrieval/generation.

Test plan:
Expanded existing unit tests in `test/modules/layers/test_quantisation.py::TestQuantisation::test_quantised_output` and ran `pytest test/modules/layers/test_quantisation.py`


torchmultimodal/modules/layers/quantisation.py
```diff
@@ -34,6 +34,7 @@ def __init__(self, num_embeddings: int, embedding_dim: int):
             -1 / self.num_embeddings, 1 / self.num_embeddings
         )
         self.quantised_vectors = None
+        self.codebook_indices = None
 
     def _preprocess(self, x: Tensor):
```
ankitade, line 39: missing return type. any idea why mypy isnt catching these @ebsmothers ?

torchmultimodal/modules/layers/quantisation.py
```diff
@@ -66,33 +67,34 @@ def quantise(self, x_flat: Tensor):
...
 
     def forward(self, x: Tensor):
         # Reshape and flatten encoder output for quantisation
         x_flat, permuted_shape = self._preprocess(x)
 
         # Quantisation via nearest neighbor lookup
-        quantised_flat = self.quantise(x_flat)
+        quantised_flat, codebook_indices = self.quantise(x_flat)
+        self.quantised_vectors = quantised_flat
+        self.codebook_indices = codebook_indices
```
ankitade, line 87: why do we set instance variables inside forward? i thought we can do away with get_quantised based on discussion with Lan or did I understand wrong?

test/modules/layers/test_quantisation.py
```diff
@@ -63,8 +68,12 @@ def test_quantised_output(self):
                 ],
             ]
         )
+        expected_flat = expected.permute(0, 2, 1).contiguous().view(-1, 5)
+        expected_codebook = torch.Tensor([2, 2, 0, 2, 1, 3]).type(torch.LongTensor)
 
         assert_expected(actual, expected)
+        assert_expected(actual_flat, expected_flat)
```
ankitade, line 75: is this redundant based on the previous check?

torchmultimodal/modules/layers/quantisation.py
```diff
@@ -66,33 +67,34 @@ def quantise(self, x_flat: Tensor):
...
 
         return quantised
 
-    def get_quantised_vectors(self):
+    def get_quantised(self):
         # Retrieve the previously quantised vectors without forward passing again
-        if self.quantised_vectors is None:
-            raise Exception(
+        if self.quantised_vectors is None or self.codebook_indices is None:
+            raise NameError(
```
ankitade, line 97: why nameError here, whats the wrong name :) ?
assertion might make more sense

torchmultimodal/modules/layers/quantisation.py
RdoubleA: My understanding was that it may be necessary to retrieve the quantised vectors without having to do a forward pass again, @langong347

test/modules/layers/test_quantisation.py
RdoubleA: normally yes but wanted to make sure that `get_quantised` returns the correct flattened quantised vectors

torchmultimodal/modules/layers/quantisation.py
RdoubleA: internal linter seems to prefer the `if...raise` pattern over `assert` and doesn't like when I test a function for an assertion to be raised. The preferred pattern is if, raise specific error, see [earlier discussion](https://www.internalfb.com/diff/D36030926?dst_version_fbid=450544480168607&transaction_fbid=3255484671399666). 

`NameError` was the closest (raised when a local or global name is not found), but maybe I could change this to `AssertionError`.

torchmultimodal/modules/layers/quantisation.py
ebsmothers: @ankitade I think we need to set some of [these](https://mypy.readthedocs.io/en/stable/command_line.html#untyped-definitions-and-calls) flags in our mypy config file. Let me add them in a separate PR

torchmultimodal/modules/layers/quantisation.py
RdoubleA: actually I understood this incorrectly, you are right the forward should just return the quantised vectors + codebook indices. will update this according to the design doc.

langong347 (approved):
LGTM. Just a nit about naming in the unit test.

PR #42, opened 2022-05-12
Move transformer classes out of flava.py into modules/layers/transformer.py

RdoubleA:
Summary:
Simplify `flava.py` by refactoring transformer classes out into their own file for a more modular design.

Test plan:
Added regression test for `TransformerEncoder`, and also ensured flava tests still pass, so I ran the whole test suite `pytest -vv`


test/modules/layers/test_transformer.py
```diff
@@ -0,0 +1,43 @@
...
+
+        actual_last_hidden_state_mean = torch.mean(output.last_hidden_state).item()
+        actual_hidden_states_mean = torch.mean(torch.cat(output.hidden_states)).item()
+        actual_attentions_mean = torch.mean(torch.cat(output.attentions)).item()
+
+        expected_last_hidden_state_mean = 0.4884
+        expected_hidden_states_mean = 0.4951
+        expected_attentions_mean = 0.2000
+
+        self.assertAlmostEqual(
```
ankitade, line 35: why assert only on the mean? we can reduce the hidden dim and check actual outputs?

RdoubleA: was thinking it would achieve the same thing, but I could mean across hidden dim instead and check against expected 5x5 as it will test against more values

torchmultimodal/modules/layers/transformer.py
```diff
@@ -0,0 +1,327 @@
...
+        initializer_range: float = 0.02,
+        use_cls_token: bool = True,
+        **kwargs: Any,
+    ):
+        super().__init__()
+        self.encoder = encoder
+        self.layernorm = layernorm
+        self.pooler = pooler
+        if use_cls_token:
+            self.cls_token = nn.Parameter(torch.zeros(1, 1, encoder.hidden_size))
```
RdoubleA, line 274: @ankitade mypy cannot find the right overload for this `torch.zeros` call, it thinks this call has signature `(int, int, Union[Tensor, Module])` but `hidden_size` is an int member of `TransformerEncoder`. This probably starting showing up once it was moved under `flava.py` because mypy ignores `flava.py`. Do we assume that the encoder here will always be an instance of `TransformerEncoder`? Or can we give it any encoder? `hidden_size` is not a member of `nn.Module` so it looks like we expect a custom encoder class, in which case maybe we should change the type annotation in the init function

torchmultimodal/models/flava.py
```diff
@@ -29,11 +34,7 @@
...
 EMBEDDING_OPTIONS = Literal["image", "text", "mm"]
-TransformerOutput = namedtuple(
-    "TransformerOutput",
-    ["last_hidden_state", "pooler_output", "hidden_states", "attentions"],
-    defaults=(None, None, None, None),
-)
+
 FLAVAOutput = namedtuple(
     "FLAVAOutput",
     ["image", "image_masked", "text", "text_masked", "multimodal", "multimodal_masked"],
```
RdoubleA, line 40: I'm thinking a cleaner way to define a named tuple is to do it just like `TwoTowerOutput` [here](https://github.com/facebookresearch/multimodal/blob/f03c241905ed276be4db0f40ce3fc535118ddd2c/torchmultimodal/architectures/two_tower.py#L14). That way it's consistent across the codebase and you don't have to define the type annotations separately. What are your thoughts? @ankitade

ankitade: yeah we can refactor this eventually. I am tempted to just do dataclasses though. The two tower one uses namedtuple because dataclasses are not torchscript friendly but thats not restriction for flava

torchmultimodal/modules/layers/transformer.py
ankitade: does doing something like this fix the issue ?
hidden_size: int = encoder.hidden_size

torchmultimodal/models/flava.py
RdoubleA: ok I will leave this for another PR

torchmultimodal/modules/layers/transformer.py
RdoubleA: nope, same issue

ankitade (commented):
realized flava tests are being skipped by default. can you double check else remove the skip annotation and rerun

test/modules/layers/test_transformer.py
```diff
@@ -0,0 +1,118 @@
...
+                    [[0.7162, 1.0436], [0.7712, 0.9802], [0.4622, 1.2150]],
+                    [[0.7426, 0.9244], [0.3179, 0.9238], [0.1016, 0.7398]],
+                ],
+                [
+                    [[0.4387, 2.2609], [0.4937, 2.1976], [0.1847, 2.4323]],
+                    [[0.4651, 2.1418], [0.0404, 2.1412], [-0.1759, 1.9571]],
+                ],
+            ]
+        )
+        expected_attentions = torch.Tensor(
```
apsdehal, line 55: Usually it is better to check for sum of the tensors to keep the code manageable.

torchmultimodal/models/flava.py
```diff
@@ -324,23 +325,6 @@ def to_2tuple(x):
     return (x, x)
 
 
-def _init_weights(module, initializer_range):
```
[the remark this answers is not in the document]
RdoubleA: I've moved this to `init_transformer_weights` in `transformer.py`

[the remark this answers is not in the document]
RdoubleA: I've moved this to `init_transformer_weights` in `transformer.py`

torchmultimodal/modules/layers/transformer.py
```diff
@@ -0,0 +1,327 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
```
apsdehal, line 5: Can you maintain the comment on the top of the file: https://github.com/facebookresearch/multimodal/pull/42/files#diff-88df49f19a5f0b27d93c6c37c2d3238bfefd7cd62373645609dc772a1143b126L7-L8?

torchmultimodal/modules/layers/transformer.py
```diff
@@ -0,0 +1,327 @@
...
+    defaults=(None, None, None, None),
+)
+
+
+class TransformerSelfAttention(nn.Module):
+    def __init__(
+        self,
+        hidden_size: int = 768,
+        num_attention_heads: int = 12,
+        attention_probs_dropout_prob: float = 0.0,
```
apsdehal, line 28: Probably also take `**kwargs` at this point since you are refactoring.

torchmultimodal/models/flava.py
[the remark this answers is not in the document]
apsdehal: I think what Ankita meant is that this function is still used inside this file and will need to be imported here.

torchmultimodal/modules/layers/transformer.py
```diff
@@ -0,0 +1,327 @@
...
+        layer_output = self.output(layer_output)
+        layer_output = self.dropout(layer_output)
+        layer_output += hidden_states
+
+        outputs = (layer_output,) + outputs
+
+        return outputs
+
+
+class TransformerEncoder(nn.Module):
```
apsdehal, line 196: Just to be clear, this is a pre-norm transformer and not a generally used postnorm architecture. Depending on how torchmm plans to support them, you can make it more specific.

torchmultimodal/modules/layers/transformer.py
```diff
@@ -0,0 +1,327 @@
...
+from torchmultimodal.modules.layers.normalizations import Fp32LayerNorm
+
+TransformerOutput = namedtuple(
+    "TransformerOutput",
+    ["last_hidden_state", "pooler_output", "hidden_states", "attentions"],
+    defaults=(None, None, None, None),
+)
+
+
+class TransformerSelfAttention(nn.Module):
```
apsdehal, line 23: This refactor will probably confused users more because as of now CLIP implementation is using Transformer from torchtext and then now we have these Transformer implementation. I would suggest possibly renaming them to `FlavaTransformer`, `FlavaSelfAttention` and so on or figure out the long term unification strategy.

apsdehal: This is also one of the reasons why I kept it inside the flava model file itself to avoid contamination of lib's modules.

torchmultimodal/modules/layers/transformer.py
apsdehal: Convert to int before passing by `int(encoder.hidden_size)`?

apsdehal (changes requested):

test/modules/layers/test_transformer.py
RdoubleA: @ankitade should I go back to reducing the tensor completely or keep it like this as we discussed earlier?

torchmultimodal/modules/layers/transformer.py
RdoubleA: I will rename to flava transformer for now until we figure out long term unification

torchmultimodal/modules/layers/transformer.py
RdoubleA: do we need `**kwargs` here?

test/modules/layers/test_transformer.py
ankitade: Its fine to leave them here. Its good to have granular checking at this point for guardrails for refactoring

torchmultimodal/modules/layers/transformer.py
ankitade: @apsdehal yes we have plans on have a generic transformer implementation.
 This task is simply a first step to move out the transformer and write tests to ensure we dont break anything during later refactoring. Happy to walk you through our plans  in our 1:1

torchmultimodal/modules/layers/transformer.py
RdoubleA: unfortunately, still the same issue

torchmultimodal/modules/layers/transformer.py
```diff
@@ -0,0 +1,330 @@
...
+
+class FLAVASelfAttention(nn.Module):
+    def __init__(
+        self,
+        hidden_size: int = 768,
+        num_attention_heads: int = 12,
+        attention_probs_dropout_prob: float = 0.0,
+    ):
+        super().__init__()
+        if hidden_size % num_attention_heads != 0:
```
langong347, line 34: Add tests for exceptions.

torchmultimodal/modules/layers/transformer.py
```diff
@@ -0,0 +1,330 @@
...
+    "FLAVATransformerOutput",
+    ["last_hidden_state", "pooler_output", "hidden_states", "attentions"],
+    defaults=(None, None, None, None),
+)
+
+
+class FLAVASelfAttention(nn.Module):
+    def __init__(
+        self,
+        hidden_size: int = 768,
```
langong347, line 29: I assume these values are specific to FLAVA? @apsdehal

torchmultimodal/modules/layers/transformer.py
RdoubleA: added `#type: ignore` for now

ankitade (commented):
LGTM. I can do the merging once the fixes get merged. Thanks Rafi

PR #43, opened 2022-05-12
Add model and architecture for Omnivore model

YosuaMichael:
Summary:
[Omnivore](https://arxiv.org/abs/2201.08377) is a vision multimodal model that able to classify RGB image, video, and depth image (RGBD) with a shared trunk parameters.

With this PR, we aim to have the model and architecture class for Omnivore that able to load converted pretrained weight from original author: https://github.com/facebookresearch/omnivore and produce same result.

Test plan:
- Add notebook in `examples/omnivore/LoadOriginalPretrainedWeightAndCompare.ipynb` that load the original omnivore pretrained weight in torchhub and make sure the output is the same
- Add unittest


torchmultimodal/modules/encoders/swin_transformer_3d_encoder.py
```diff
@@ -0,0 +1,567 @@
...
+from typing import Optional, Callable, List, Tuple
+
+import torch
+import torch.nn.functional as F
+from torch import nn, Tensor
+from torchvision.models.vision_transformer import MLPBlock
+from torchvision.ops.stochastic_depth import StochasticDepth
+
+
+def _compute_pad_size_3d(
```
ankitade, line 14: why do u need this and next function to be global?

torchmultimodal/modules/encoders/swin_transformer_3d_encoder.py
```diff
@@ -0,0 +1,567 @@
...
+            x, shifts=(shift_size[0], shift_size[1], shift_size[2]), dims=(1, 2, 3)
+        )
+
+    # unpad features
+    x = x[:, :D, :H, :W, :].contiguous()
+    return x
+
+
+# Commented because mypy can't detect torch.fx
+# torch.fx.wrap("shifted_window_attention_3d")
```
ankitade, line 223: remove.

torchmultimodal/architectures/omnivore.py
```diff
@@ -0,0 +1,39 @@
...
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+import torch
+from torch import nn
+
+
+class OmnivoreArchitecture(nn.Module):
```
ankitade, line 11: I know you are following the other examples here. but omnivorearchitecture is essentially an encoder+ head. Idk if we need a separate "architecture" for it. we can discuss in design meeting

ankitade (commented):
- some tests are failing on CI, can you take a look
- I left some comments around passing in instantiated modules but realized you are following torchvision style. so if we plan to upstream this, its ok to leave them as is

torchmultimodal/modules/encoders/swin_transformer_3d_encoder.py
```diff
@@ -0,0 +1,567 @@
...
+        embed_dim: int,
+        depths: List[int],
+        num_heads: List[int],
+        window_size: Tuple[int, int, int] = (2, 7, 7),
+        mlp_ratio: float = 4.0,
+        dropout: float = 0.0,
+        attention_dropout: float = 0.0,
+        stochastic_depth_prob: float = 0.0,
+        norm_layer: Optional[Callable[..., nn.Module]] = None,
+        block: Optional[Callable[..., nn.Module]] = None,
```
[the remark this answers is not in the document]
YosuaMichael: Could you elaborate more or maybe give some examples on this? What do you mean by separate function which takes in flat params?

torchmultimodal/modules/encoders/swin_transformer_3d_encoder.py
```diff
@@ -0,0 +1,567 @@
...
+        pad_size = _compute_pad_size_3d((D, H, W), self.patch_size)
+        x = F.pad(x, (0, pad_size[2], 0, pad_size[1], 0, pad_size[0]))
+        x = self.proj(x)  # B C D Wh Ww
+        x = x.permute(0, 2, 3, 4, 1)  # B D Wh Ww C
+        if self.norm is not None:
+            x = self.norm(x)
+        return x
+
+
+class PatchEmbedOmnivore(nn.Module):
```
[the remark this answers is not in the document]
YosuaMichael: Ah sure, that is a good idea!

torchmultimodal/modules/encoders/swin_transformer_3d_encoder.py
```diff
@@ -0,0 +1,567 @@
...
+        patch_size (Tuple[int, int, int]): Patch token size. Default: (2, 4, 4).
+        in_chans (int): Number of input channels. Default: 3
+        embed_dim (int): Number of linear projection output channels. Default: 96.
+        norm_layer (nn.Module, optional): Normalization layer. Default: None
+    """
+
+    def __init__(
+        self,
+        patch_size: Tuple[int, int, int] = (2, 4, 4),
+        in_chans: int = 3,
```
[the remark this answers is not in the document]
YosuaMichael: noted

torchmultimodal/modules/encoders/swin_transformer_3d_encoder.py
YosuaMichael: For `_compute_pad_size_3d` it is needed on 2 different classes so we put it on global but private.
For `_compute_attention_mask_3d` we put this as function so we can cache it. We don't put it inside the function `shifted_window_attention_3d` because we avoid nested function as it is not supported by jit script.

torchmultimodal/architectures/omnivore.py
YosuaMichael: Yeah, the omnivore architecture is basically an encoder + multiple heads (using ModuleDict). Sure, it is a good idea to discuss on design meeting

test/models/test_omnivore.py
```diff
@@ -0,0 +1,33 @@
...
+        torch.manual_seed(42)
+        self.device = get_current_device()
+
+    def test_omnivore_swin_t_forward(self):
+        model = omnivore.omnivore_swin_t().to(self.device)
+        self.assertTrue(isinstance(model, torch.nn.Module))
+
+        image = torch.randn(1, 3, 1, 112, 112)  # B C D H W
+        image_score = model(image, input_type="image")
+        self.assertEqual(image_score.size(), torch.Size((1, 1000)))
```
langong347, line 25: please use our test utility to assert on values and shapes. (we need to assert values too)

torchmultimodal/models/omnivore.py
```diff
@@ -0,0 +1,96 @@
...
+        )
+
+        self.depth_patch_embed = PatchEmbed3d(
+            patch_size=patch_size,
+            in_channels=1,
+            embed_dim=embed_dim,
+            norm_layer=norm_layer,
+        )
+
+    def forward(self, x):
```
langong347, line 69: Make sure to add type hints.

torchmultimodal/modules/encoders/swin_transformer_3d_encoder.py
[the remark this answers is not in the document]
langong347: You can check our newly uploaded wiki about developer guide and let us know if it makes sense (ping you separately in chat). But for convenience here is the relevant block:

>If you have class A which uses instances of class B and C, the init should take in instantiations of B and C (instead of params for B and C and instantiating within init). This will give users flexibility to pass in their own instantiations and now A is independent of any init signature changes of B and C.

torchmultimodal/modules/encoders/swin_transformer_3d_encoder.py
```diff
@@ -0,0 +1,518 @@
...
+        dropout: float = 0.0,
+        attention_dropout: float = 0.0,
+        stochastic_depth_prob: float = 0.0,
+        norm_layer: Optional[Callable[..., nn.Module]] = None,
+        block: Optional[Callable[..., nn.Module]] = None,
+        patch_embed: Optional[Callable[..., nn.Module]] = None,
+    ):
+        super().__init__()
+
+        if block is None:
```
langong347, line 446: No need of those anymore if we pass in directly the instantiated modules (or its partial)

torchmultimodal/modules/encoders/swin_transformer_3d_encoder.py
```diff
@@ -0,0 +1,518 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
```
langong347, line 1: Consider this is a big module: can we add tests for some key components here?

langong347:
Q: how did you make sure that the order of pretrained model keys is the same as that of your model keys?
For example, if I implement the same model in a slightly different way (e.g., additional variables, layers), will it be a problem for me to match the original implementation?

```
key_mapping = {pretrained_keys[i]: model_keys[i] for i in range(len(model_keys))}
```

test/models/test_omnivore.py
YosuaMichael: Thanks for the suggestion! Will do.

torchmultimodal/models/omnivore.py
YosuaMichael: 👍

YosuaMichael:
> Q: how did you make sure that the order of pretrained model keys is the same as that of your model keys?
For example, if I implement the same model in a slightly different way (e.g., additional variables, layers), will it be a problem for me to match the original implementation?

Sometimes this can be a problem, especially if we have branching operation. For instance in omnivore, the head that use `ModuleDict` need to have the same order as the original implementation or otherwise it will not match the correct key using that method (need a more complex way to auto assign like doing a matching with same size).

However from what I understand, if there is no branching (like the swin transformer encoder only, in fact the implementation here is different with the original one: https://github.com/facebookresearch/omnivore/blob/main/models/swin_transformer_3d.py ) usually the layout will follow the same order on how the input will be processed, and in this case it should be okay. Some of the problem that may occur here is that we might store same data in different dimension like 100 x 100 (2D) or 10000 (1D).

torchmultimodal/modules/encoders/swin_transformer_3d_encoder.py
YosuaMichael: ok, will try to do it

torchmultimodal/modules/encoders/swin_transformer_3d_encoder.py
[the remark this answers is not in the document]
YosuaMichael: I think this method of passing instantiated object would be difficult for a dynamically instantiated layers.

For example here SwinTransformer3dEncoder have a stack of SwinTransformer3dBlock where the size of the stack depend on the config and the parameter of each SwinTransformer3dBlock depend on the previous layer. Any suggestions for this?

torchmultimodal/modules/encoders/swin_transformer_3d_encoder.py
YosuaMichael: Hi @langong347 , I have added some more test, could you help me review them?

torchmultimodal/modules/encoders/swin_transformer_3d_encoder.py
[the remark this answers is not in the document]
langong347: >For example here SwinTransformer3dEncoder have a stack of SwinTransformer3dBlock where the size of the stack depend on the config 

Are you referring to `depth`? Since it defines the structure of the SwinTransformer3dEncoder and is not related to the classes it relies on,  I think it's ok to leave it as an argument of the constructor.

> and the parameter of each SwinTransformer3dBlock depend on the previous layer.

What if someone wants to use a variant of SwinTransformer3dBlock? Would it be possible to allow the user to pass the class name instead of the Encoder assigns it? Also, why not making SwinTransformer3dBlock and `partial(nn.LayerNorm, eps=1e-5)` the default values for `block` and `norm_layer`?

I agree the approach described by our contributing guide is not applicable here since class B and C (the building blocks) are dynamically instantiated inside class A (the class that uses them). But ultimately, the idea is to allow customization and reusability. 
If you think SwinTransformer3dEncoder can only use SwinTransformer3dBlock but nothing else then it makes less sense to allow that flexbility.

test/models/test_omnivore.py
```diff
@@ -0,0 +1,36 @@
...
+import torchmultimodal.models.omnivore as omnivore
+from torchmultimodal.utils.common import get_current_device
+
+
+class TestOmnivoreModel(unittest.TestCase):
+    def setUp(self):
+        self.device = get_current_device()
+
+    def test_omnivore_swin_t_forward(self):
+        torch.manual_seed(42)
```
langong347, line 19: Use this:
https://github.com/facebookresearch/multimodal/blob/e8575411d8fdc3f97c51d36789b2f6c530704a50/test/test_utils.py#L60

test/models/test_omnivore.py
langong347: Have you found this?

https://github.com/facebookresearch/multimodal/blob/e8575411d8fdc3f97c51d36789b2f6c530704a50/test/test_utils.py#L74

torchmultimodal/architectures/omnivore.py
```diff
@@ -0,0 +1,39 @@
...
+
+    def __init__(self, encoder: nn.Module, heads: nn.ModuleDict):
+        super().__init__()
+        self.encoder = encoder
+        self.heads = heads
+        self.input_types = set(heads.keys())
+
+    def forward(self, x: torch.Tensor, input_type: str):
+        x = self.encoder(x)
+        assert (
```
langong347, line 35: add a test for this as we are branching out

test/modules/encoders/test_swin_transformer_3d_encoder.py
```diff
@@ -0,0 +1,84 @@
...
+                embed_dim=96,
+                depths=[2, 2, 6, 2],
+                num_heads=[3, 6, 12, 24],
+                window_size=(8, 7, 7),
+                stochastic_depth_prob=0.2,
+                norm_layer=torch.nn.LayerNorm,
+                patch_embed=PatchEmbed3d,
+            )
+            .to(self.device)
+            .eval()
```
langong347, line 37: Why `eval`? Is it to get around the `dropout` layer (but you've already fixed the seed, dropout should also be fixed).

test/modules/encoders/test_swin_transformer_3d_encoder.py
```diff
@@ -0,0 +1,84 @@
...
+        )
+
+    def test_swin_transformer_3d_encoder(self):
+        torch.manual_seed(42)
```
langong347, line 41: redundant? The seed fixed in `setUp` should prevail through the methods right?

torchmultimodal/modules/encoders/swin_transformer_3d_encoder.py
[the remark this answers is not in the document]
langong347: The same consideration applies to `SwinTransformerBlock3d` which also takes in a few lower level classes

test/modules/encoders/test_swin_transformer_3d_encoder.py
```diff
@@ -0,0 +1,84 @@
...
+        self.assertTrue(isinstance(self.encoder, torch.nn.Module))
```
langong347, line 42: Why do we need to assert `nn.Module` here? The class is instantiated statically in `setUp` right? Or is it due to `eval`?

torchmultimodal/modules/encoders/swin_transformer_3d_encoder.py
```diff
@@ -0,0 +1,518 @@
...
+# Modified from 2d Swin Transformers in torchvision:
+# https://github.com/pytorch/vision/blob/main/torchvision/models/swin_transformer.py
+
+from functools import lru_cache, partial
+from typing import Callable, List, Optional, Tuple
+
+import torch
+import torch.nn.functional as F
+from torch import nn, Tensor
+from torchvision.models.vision_transformer import MLPBlock
```
langong347, line 16: Is it possible to bootstrap from our MLP layer?

https://github.com/facebookresearch/multimodal/blob/main/torchmultimodal/modules/layers/mlp.py#L13

torchmultimodal/modules/encoders/swin_transformer_3d_encoder.py
```diff
@@ -0,0 +1,518 @@
...
+    qkv = qkv.reshape(x.size(0), x.size(1), 3, num_heads, C // num_heads).permute(
+        2, 0, 3, 1, 4
+    )
+    q, k, v = qkv[0], qkv[1], qkv[2]
+    q = q * (C // num_heads) ** -0.5
+    attn = q.matmul(k.transpose(-2, -1))
+    # add relative position bias
+    attn = attn + relative_position_bias
+
+    if sum(shift_size) > 0:
```
langong347, line 183: Can we add a test when this is not true?

torchmultimodal/modules/encoders/swin_transformer_3d_encoder.py
YosuaMichael: I think we can do this, but take note that there is a slight different.
The original MLP: https://github.com/facebookresearch/omnivore/blob/main/models/swin_transformer_3d.py#L65
and the one in torchvision have final drop layer after at the end. While the torchmultimodal MLP layer can't have dropout layer at the end.

Not sure how big is the effect of this dropout, but we could try.

torchmultimodal/modules/encoders/swin_transformer_3d_encoder.py
YosuaMichael: sure

torchmultimodal/modules/encoders/swin_transformer_3d_encoder.py
langong347: If the end state of this module is to upstream to torchvision, I think it's fine to stick with vision's components here. Also I notice that `MLPBlock` in torchvision is cusotmized for transformer so I don't think our MLP is the equivalent.

test/modules/encoders/test_swin_transformer_3d_encoder.py
YosuaMichael: Oh, I didn't know that.... I assume it is only called once like __init__ because the value that I have before I add new function dont work anymore on my laptop and I need to change it. Let me double verify this behaviour.

test/models/test_omnivore.py
YosuaMichael: @langong347 after I read the utility `assert_expected` it seems to compare two tensor with float type. In this case, I think `assertEqual` is better for comparing the size since it is integer?

torchmultimodal/architectures/omnivore.py
YosuaMichael: I have added the test on `test/models/test_omnivore.py`

test/modules/encoders/test_swin_transformer_3d_encoder.py
YosuaMichael: @langong347  you are right. I have confirm that setUp indeed called on every method. Hmm... then the problem I encountered before might caused by other thing. Anyway I have added the seed init on setUp instead now.

torchmultimodal/modules/encoders/swin_transformer_3d_encoder.py
YosuaMichael: added the test on `test/modules/encoder/test_swin_transformer_3d_encoder.py:TestSwinTransformer3dComponents.test_shifted_window_attention_3d_zero_shift`

torchmultimodal/architectures/omnivore.py
YosuaMichael: Have talk offline, the OmnivoreArchitecture is indeed quite generic like MTL architecture. However for now we will keep it as Omnivore and if there are other use cases who need MTL we can reuse and make it more generic.

torchmultimodal/architectures/omnivore.py
```diff
@@ -0,0 +1,39 @@
...
+                the correct head.
+    """
+
+    def __init__(self, encoder: nn.Module, heads: nn.ModuleDict):
+        super().__init__()
+        self.encoder = encoder
+        self.heads = heads
+        self.input_types = set(heads.keys())
+
+    def forward(self, x: torch.Tensor, input_type: str):
```
datumbox, line 33: The idiom of passing the `input_type` as string and then choosing to which head to forward the data is not very common. Why can't we instantiate Omnivore modules with the same encoder but different heads and push to the training loop which module you use depending on the input? This will move the complexity from the nn.Module to the loop and it will be beneficial for use-cases where the loss is different for each head. 

Speaking offline with @YosuaMichael there are some concerns on whether the whole approach will work well on a distributed setup.

torchmultimodal/modules/encoders/swin_transformer_3d_encoder.py
```diff
@@ -0,0 +1,512 @@
...
+        pad_size = _compute_pad_size_3d((D, H, W), self.patch_size)
+        x = F.pad(x, (0, pad_size[2], 0, pad_size[1], 0, pad_size[0]))
+        x = self.proj(x)  # B C D Wh Ww
+        x = x.permute(0, 2, 3, 4, 1)  # B D Wh Ww C
+        if self.norm is not None:
+            x = self.norm(x)
+        return x
+
+
+class SwinTransformer3dEncoder(nn.Module):
```
jdsgomes, line 411: I agree with @datumbox  regarding the adaptation to 3d. I think it would be preferable to add it to `torchvision` directly. This main Class can be a good entry point for the unification effort, looking into it, I think we can have a common logic under `SwinTransformer` with extra parameters  in order to cater for the 2d (`SwinTransformer2d`) and 3d (`SwinTransformer3d`) versions. For BC the default values should be for the  2d version. From this main class we could drill down and look into more unification efforts (not possible in all cases probably). For example, we have `ConvNormActivationBlock` and then 2d and 3d vertions (see https://github.com/pytorch/vision/blob/49496c4f6201f05f2351788389a0863b087e78f4/torchvision/ops/misc.py#L68). This was a simpler example, but could be used as inspiration.
Let me know your thoughts @YosuaMichael. I maybe I might have overlooked some of the differences which make this even harder than I think.

YosuaMichael: @jdsgomes I agree by add this to torchvision (planned to do later). But for having generic SwinTransformer for both 2d and 3d, I have a feeling this could be difficult, especially to make it BC. Here are some reasons I can think of (let me know if you have a good solution):
-  some of the params input differs in type. For instance the `window_size` for 2d is `int` (they assume it always a symmetry windows like 16x16), however in 3d we use `Tuple[int, int, int]` since we have windows like `(8, 7, 7)`. 
- In this 3d version, I modularize the `patch_embed` layer because the Omnivore actually use different `patch_embed` method

torchmultimodal/modules/encoders/swin_transformer_3d_encoder.py
```diff
@@ -0,0 +1,512 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+# Modified from 2d Swin Transformers in torchvision:
+# https://github.com/pytorch/vision/blob/main/torchvision/models/swin_transformer.py
```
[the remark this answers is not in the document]
datumbox: @YosuaMichael I recommend taking the time to create a plan for this and similar models. Not only it will deliver better code quality but also sketch out how we will handle supporting Multimodal more effectively on the future. Copy pasting is a shortcut that kicks down the line the problem and only makes it harder to solve.

torchmultimodal/architectures/omnivore.py
YosuaMichael: Hi @datumbox, currently I plan to try on doing the training first and see if there are any particular problem with the architectures (whether the current or multiple model with shared encoder). In particular, I need to check the behaviour of having 2 models `A` and `B` that shared the same `encoder` wrapped individually into `DDP(A)` and `DDP(B)`.

torchmultimodal/modules/encoders/swin_transformer_3d_encoder.py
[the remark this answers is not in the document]
datumbox: @ankitade @langong347 Sharing here what we discussed offline with @YosuaMichael and @kartikayk.

The adaptation of Swin 2d to 3d, highlighted some things we can improve on the original implementation of TorchVision. One of them is the use of single integers instead of tuples/lists for the sizes, another one is the fact that some modules can easily be reused if we make minor adaptations (candidates are `PatchMerging` and `SwinTransformerBlock`). It would be really nice, if we could make these changes on TorchVision's code now to minimize the degree of copy-pasting. The reason for the urgency is the upcoming release and the fact that Swin is a brand new class which can be easily modified without worrying about BC now. After the release, there will be all sorts of BC considerations.

The timing is tight and we definitely don't want to delay you. So we will just give it 1-2 days max to see if that's possible. If we are successful to refactor, we will be able to simplify this implementation and reduce the copy-paste. If not, we will merge this and review improvements on TorchVision on the future.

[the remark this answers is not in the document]
YosuaMichael: Just to update here, the update on torchvision 2d swin transformer is on https://github.com/pytorch/vision/pull/6088

ankitade (approved):

torchmultimodal/architectures/omnivore.py
YosuaMichael: @ankitade yes, the original checkpoint include all the heads

torchmultimodal/architectures/omnivore.py
```diff
@@ -0,0 +1,39 @@
...
+        x = self.encoder(x)
+        assert (
+            input_type in self.input_types
```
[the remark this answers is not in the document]
YosuaMichael: Good suggestion, will do :)

torchmultimodal/models/omnivore.py
```diff
@@ -0,0 +1,101 @@
...
+        patch_size=(2, 4, 4),
+        embed_dim=96,
+        depths=[2, 2, 6, 2],
+        num_heads=[3, 6, 12, 24],
+        window_size=(8, 7, 7),
+        stochastic_depth_prob=0.2,
+        norm_layer=nn.LayerNorm,
+        patch_embed=PatchEmbedOmnivore,
+    )
+    if encoder_only:
```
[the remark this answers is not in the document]
YosuaMichael: Ok, I think we can create a function on the encoder file to return the encoder variant and call it here.

YosuaMichael (commented):
Hi @ankitade, thanks for your comment and sorry for a very late feedback. I miss the notif on your comment...
I will update the code to reflect some of your comment, thanks!

torchmultimodal/modules/encoders/swin_transformer_3d_encoder.py
```diff
@@ -0,0 +1,538 @@
...
+            self.num_heads,
+            shift_size=shift_size,
+            attention_dropout=self.attention_dropout,
+            dropout=self.dropout,
+            qkv_bias=self.qkv.bias,
+            proj_bias=self.proj.bias,
+        )
+
+
+class SwinTransformerBlock(nn.Module):
```
datumbox, line 328: Given you made the necessary changes on TorchVision to support other attention types (https://github.com/pytorch/vision/pull/6088), could you use those components instead of copy-pasting?

YosuaMichael: You are right, will upstream the components from torchvision.

torchmultimodal/architectures/omnivore.py
datumbox: @ankitade @YosuaMichael It's worth keeping in mind that this pattern wont support FX.

FX does tracing which means the flow of the execution of the model should not depend on the input. By adding the string FX won't know what to execute. Even though you might not be interested right now to make it FX traceable, on the future you might want to adopt FX quantization or other FX-based utils.

This is the reason my advice is to split this module in submodules depending on the head. It's a safer idiom that would be forward compatible with future core expansions. It might require some massaging on the original weights to fix but I think that's straightforward to do and worth it.

Up to you. :)

PR #45, opened 2022-05-17
[flava][checkpoints] Config for lightning checkpoints

IvanKobzarev:
Summary:
Adding pytorch_lightning checkopoint functionality usage.

1. config.training.lightning_checkpoint
To be explicit with the clients config section is named lightning_checkpoint, which content will be unpacked to ModelCheckpointCallback ( https://github.com/PyTorchLightning/pytorch-lightning/blob/master/pytorch_lightning/callbacks/model_checkpoint.py ).

2. config.training.lightning -> config.training.lightning_trainer
For the same reason, what is unpacked inside Lightning Trainer renamed to "lightning_trainer".

3. config.training.lightning_load_from_checkpoint
As a checkpoint to recover model state, used for LightningModule.load_from_checkpoint

Test plan:
python train.py config=configs/pretraining/debug.yaml

check that last.ckpt appeared even if training was interrupted

python train.py config=configs/pretraining/debug.yaml config.training.lightning_load_from_checkpoint=last.ckpt

Check that training started from the last epoch

### DIFF
--- a/examples/flava/configs/finetuning/qnli.yaml
+++ b/examples/flava/configs/finetuning/qnli.yaml
@@ -8,6 +8,14 @@ training:
     progress_bar_refresh_rate: 50
     val_check_interval: 1000
     num_sanity_val_steps: 0
+  lightning_checkpoint:
+    dirpath: "."
+    filename: flava-{epoch:02d}-{step}
+    save_last: true
+    every_n_train_steps: 1000
+    save_on_train_epoch_end: true
+    verbose: true
+  lightning_load_from_checkpoint: null
   seed: -1
   batch_size: 32
   num_workers: 4
--- a/examples/flava/definitions.py
+++ b/examples/flava/definitions.py
@@ -66,6 +66,8 @@ class TrainingDatasetsInfo:
 class TrainingArguments:
     # Any lightning args to be pushed here
     lightning: Dict[str, Any] = field(default=dict)
+    lightning_checkpoint: Optional[Dict[str, Any]] = None
+    lightning_load_from_checkpoint: Optional[str] = None
     seed: int = -1
     batch_size: int = 8
     num_workers: int = 4
--- a/examples/flava/train.py
+++ b/examples/flava/train.py
@@ -10,7 +10,7 @@ from definitions import FLAVAArguments
 from model import FLAVAPreTrainingLightningModule
 from omegaconf import OmegaConf
 from pytorch_lightning import seed_everything, Trainer
-from pytorch_lightning.callbacks import LearningRateMonitor
+from pytorch_lightning.callbacks import LearningRateMonitor, ModelCheckpoint
 from utils import build_config, build_datamodule_kwargs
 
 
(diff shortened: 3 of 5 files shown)


examples/flava/configs/finetuning/qnli.yaml
```diff
@@ -2,7 +2,7 @@
 _target_: definitions.FLAVAArguments
 training:
   _target_: definitions.TrainingArguments
-  lightning:
+  lightning_trainer:
```
ankitade, line 5: this renaming isnt needed? lets avoid for now

examples/flava/finetune.py
```diff
@@ -51,7 +51,7 @@ def main():
         adam_weight_decay=config.training.adam_weight_decay,
         adam_betas=config.training.adam_betas,
         warmup_steps=config.training.warmup_steps,
```
ankitade, line 53: also add callback to finetune

examples/flava/definitions.py
```diff
@@ -64,7 +64,9 @@ class TrainingDatasetsInfo:
 @dataclass
 class TrainingArguments:
     # Any lightning args to be pushed here
-    lightning: Dict[str, Any] = field(default=dict)
+    lightning_trainer: Dict[str, Any] = field(default=dict)
```
ankitade, line 67: another way would be to create a data class wrapping around trainer conf and checkpoint conf. but upto u if u want to do it or just leave it this way

examples/flava/train.py
```diff
@@ -41,21 +41,29 @@ def main():
...
+            config.training.lightning_load_from_checkpoint
+        )
 
     trainer = Trainer(
-        **OmegaConf.to_container(config.training.lightning),
+        **OmegaConf.to_container(config.training.lightning_trainer),
         callbacks=[
             LearningRateMonitor(logging_interval="step"),
             MultimodalEvalCallback(imagenet_datamodule=imagenet_datamodule),
+            ModelCheckpoint(
```
ankitade, line 64: how does this behave when checkpoint conf is empty dict? we dont want to start checkpointing on every iteration. can u double check

ankitade (commented):
thanks for the PR. can u also update the test plan with a run without adding any checkpoint configs.

examples/flava/configs/finetuning/qnli.yaml
IvanKobzarev: My reason of renaming it from "lightning" - was to be explicit to clients that all the config arguments will go to the constructor arguments of LightningTrainer. As new config sections will have more details starting with lightning_ - for just. "lightning" can be confusing for clients if not to check the code. I am ok to leave it lightning for now :)

examples/flava/definitions.py
IvanKobzarev: Yes, we can extract all lightning config params to standalone Lightning config dataclass

examples/flava/configs/finetuning/qnli.yaml
ankitade: yeah i understand the intention. just wary of making unrelated changes in the same PR

examples/flava/train.py
IvanKobzarev: I have made it completely no-op for the case when `config.training.lightning_checkpoint:` is missing.
It will be useful to have model checkpointing functionality fully unplugged by config change if some error will be inside that logic on pytorch lightning / pytorch side.

PR #46, opened 2022-05-18
Re-enable flava tests for open source

ankitade:
As title. Added DallE to the dev requirements since FLAVA model depends on it but feels odd to force everyone to install it.
Differential Revision: D36495395

### DIFF
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -1,3 +1,4 @@
 pytest
 mypy
 pre-commit
+DALL-E==0.1
--- a/test/models/test_flava.py
+++ b/test/models/test_flava.py
@@ -20,7 +20,6 @@ class TestFLAVA(unittest.TestCase):
     def setUp(self):
         torch.manual_seed(1234)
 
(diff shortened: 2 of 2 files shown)


langong347 (approved):

langong347 (changes requested):
What's wrong with flava test in CI?

langong347 (commented):
Pending fix on CI passing

dev-requirements.txt
```diff
@@ -1,3 +1,4 @@
 pytest
 mypy
 pre-commit
+DALL-E==0.1
```
langong347, line 4: Non block Q: is this temporary just to enable FLAVA? Feels odd to take a dep on openAI's discrete VAE since we are implementing our own (VQVAE).

langong347 (approved):

dev-requirements.txt
ankitade: this comes due to the ckpt loading from dalle. so i don't have a good solution other than taking this dependency. longer term, yes we can hack the ckpt

PR #47, opened 2022-05-18
Introducing exponential moving average updates to embedding in Quantization layer

RdoubleA:
### Summary
VQVAE implements a codebook of embedding vectors to tokenize an input. It does this by encouraging the encoder output to be close to the embedding vectors during training. The embedding vectors learn this via exponential moving average (EMA) updates that coerce them to become an average of the encoder output. This method of optimization is found to converge faster than MSE codebook loss, as described in the original paper.

This PR introduces EMA updates to the Quantization layer with some refactoring of old unit tests. Additionally, it initializes the embedding weights with random latents from the encoder similar to VideoGPT and Jukebox. In a future PR, codebook restarts will be implemented to alleviate codebook collapse.

### Test plan
Added unit tests for embedding initialization and EMA updates in `test_quantization.py`

### DIFF
--- a/test/modules/layers/test_quantization.py
+++ b/test/modules/layers/test_quantization.py
@@ -8,7 +8,6 @@ import unittest
 
 import torch
 from test.test_utils import assert_expected, set_rng_seed
-from torch import nn
 from torchmultimodal.modules.layers.quantization import Quantization
 
 
@@ -45,7 +44,7 @@ class TestQuantization(unittest.TestCase):
         )
 
     def test_quantized_output(self):
-        self.vq.embedding = nn.Embedding.from_pretrained(self.embedding_weights)
+        self.vq.embedding = self.embedding_weights
         self.vq._is_embedding_init = True
         output = self.vq(self.encoded)
         _, actual_quantized_flat, actual_codebook_indices, actual_quantized = output
@@ -103,14 +102,14 @@ class TestQuantization(unittest.TestCase):
 
         assert_expected(actual_quantized_shape, expected_quantized_shape)
 
-    def test_embed_init(self):
+    def test_init_embedding_and_preprocess(self):
         assert not self.vq._is_embedding_init, "embedding init flag not False initially"
 
         _, _ = self.vq._init_embedding_and_preprocess(self.encoded)
 
         assert self.vq._is_embedding_init, "embedding init flag not True after init"
 
-        actual_weight = self.vq.embedding.weight
+        actual_weight = self.vq.embedding
         expected_weight = torch.Tensor(
             [
                 [2.0, -1.0, 0.0, 2.0, 0.0],
@@ -121,18 +120,18 @@ class TestQuantization(unittest.TestCase):
         )
         assert_expected(actual_weight, expected_weight)
 
-        actual_code_avg = self.vq._code_avg
+        actual_code_avg = self.vq.code_avg
         expected_code_avg = actual_weight
         assert_expected(actual_code_avg, expected_code_avg)
 
-        actual_code_usage = self.vq._code_usage
+        actual_code_usage = self.vq.code_usage
         expected_code_usage = torch.ones(self.num_embeddings)
         assert_expected(actual_code_usage, expected_code_usage)
 
     def test_ema_update(self):
         _ = self.vq(self.encoded)
 
-        actual_weight = self.vq.embedding.weight
+        actual_weight = self.vq.embedding
         expected_weight = torch.Tensor(
             [
                 [1.0000, -1.3333, 0.0000, 1.6667, 0.0000],
@@ -143,7 +142,7 @@ class TestQuantization(unittest.TestCase):
         )
         assert_expected(actual_weight, expected_weight, rtol=0.0, atol=1e-4)
 
-        actual_code_avg = self.vq._code_avg
+        actual_code_avg = self.vq.code_avg
         expected_code_avg = torch.Tensor(
             [
                 [1.5000, -2.0000, 0.0000, 2.5000, 0.0000],
@@ -154,6 +153,6 @@ class TestQuantization(unittest.TestCase):
         )
         assert_expected(actual_code_avg, expected_code_avg, rtol=0.0, atol=1e-4)
 
-        actual_code_usage = self.vq._code_usage
+        actual_code_usage = self.vq.code_usage
         expected_code_usage = torch.Tensor([1.5000, 1.0000, 1.5000, 1.0000])
         assert_expected(actual_code_usage, expected_code_usage, rtol=0.0, atol=1e-4)
--- a/test/modules/layers/test_quantization.py
+++ b/test/modules/layers/test_quantization.py
@@ -28,6 +28,7 @@ class TestQuantization(unittest.TestCase):
                 [[2, 2, -1], [1, -1, -2], [0, 0, 0], [1, 2, 1], [1, 0, 0]],
             ]
         )
+        self.encoded.requires_grad_()
         # This is 4x5
         self.embedding_weights = torch.Tensor(
             [[1, 0, -1, -1, 2], [2, -2, 0, 0, 1], [2, 1, 0, 1, 1], [-1, -2, 0, 2, 0]]
@@ -156,3 +157,30 @@ class TestQuantization(unittest.TestCase):
         actual_code_usage = self.vq.code_usage
         expected_code_usage = torch.Tensor([1.7000, 1.0000, 1.7000, 1.0000])
         assert_expected(actual_code_usage, expected_code_usage, rtol=0.0, atol=1e-4)
+
+    def test_register_buffer_tensors(self):
+        out = self.vq(self.encoded)
+        out.quantized.sum().backward()
+
+        msg_has_grad = "tensor assigned to buffer but accumulated grad"
+        assert not self.vq.code_avg.grad, msg_has_grad
+        assert not self.vq.code_usage.grad, msg_has_grad
+        assert not self.vq.embedding.grad, msg_has_grad
+
+        assert not list(
+            self.vq.parameters()
+        ), "buffer variables incorrectly assigned as params"
+
+    def test_tile(self):
+        encoded_small = self.encoded[:1, :, :2]
+        out = self.vq(encoded_small)
+        encoded_small_flat = out.encoded_flat
+        embed = self.vq.embedding
+        # Check for each embedding vector if there is one equal encoded vector + noise
+        for emb in embed:
+            assert any(
+                [
+                    torch.isclose(emb, enc, rtol=0, atol=0.01).all()
+                    for enc in encoded_small_flat
+                ]
+            ), "embedding initialized from encoder output incorrectly"
--- a/test/modules/layers/test_quantization.py
+++ b/test/modules/layers/test_quantization.py
@@ -40,9 +40,10 @@ class TestQuantization(unittest.TestCase):
         self.vq = Quantization(
             num_embeddings=self.num_embeddings, embedding_dim=self.embedding_dim
         )
-        self.vq.embedding = nn.Embedding.from_pretrained(self.embedding_weights)
 
     def test_quantized_output(self):
+        self.vq.embedding = nn.Embedding.from_pretrained(self.embedding_weights)
+        self.vq._is_embedding_init = True
         output = self.vq(self.encoded)
         _, actual_quantized_flat, actual_codebook_indices, actual_quantized = output
         # This is shape (2,5,3)
--- a/test/modules/layers/test_quantization.py
+++ b/test/modules/layers/test_quantization.py
@@ -7,7 +7,7 @@
 import unittest
 
 import torch
-from test.test_utils import assert_expected
+from test.test_utils import assert_expected, set_rng_seed
 from torch import nn
 from torchmultimodal.modules.layers.quantization import Quantization
 
@@ -18,6 +18,7 @@ class TestQuantization(unittest.TestCase):
     """
 
     def setUp(self):
+        set_rng_seed(4)
         self.num_embeddings = 4
         self.embedding_dim = 5
 
@@ -38,7 +39,9 @@ class TestQuantization(unittest.TestCase):
         )
 
         self.vq = Quantization(
-            num_embeddings=self.num_embeddings, embedding_dim=self.embedding_dim
+            num_embeddings=self.num_embeddings,
+            embedding_dim=self.embedding_dim,
+            decay=0.5,
         )
 
     def test_quantized_output(self):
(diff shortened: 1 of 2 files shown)


torchmultimodal/modules/layers/quantization.py
```diff
@@ -24,24 +24,60 @@ class Quantization(nn.Module):
...
+
+        self._is_embedding_init = True
+
+        # Get random flattened encoder outputs
+        encoded_flat, permuted_shape = self._preprocess(z)
+        idx = torch.randperm(encoded_flat.shape[0])
+        encoded_flat_rand = encoded_flat[idx][: self.num_embeddings]
+
+        # Initialize embedding and intermediate values for EMA updates
+        with torch.no_grad():
```
langong347, line 75: Those are not model parameters but more like "meta data" or "state" of the module. Check out `nn.Module.register_buffer`.

test/modules/layers/test_quantization.py
```diff
@@ -84,23 +88,72 @@ def test_preprocess(self):
...
         quantized = self.vq._postprocess(self.input_tensor_flat, torch.Size([2, 2, 3]))
         actual_quantized_shape = torch.tensor(quantized.shape)
         expected_quantized_shape = torch.tensor([2, 3, 2])
 
-        assert torch.equal(
-            actual_quantized_shape, expected_quantized_shape
-        ), f"actual quantized shape: {actual_quantized_shape}, expected quantized shape: {expected_quantized_shape}"
+        assert_expected(actual_quantized_shape, expected_quantized_shape)
+
+    def test_embed_init(self):
```
langong347, line 106: More explicit this way: `test_{the actual method name}` instead of shortening it

torchmultimodal/modules/layers/quantization.py
```diff
@@ -24,24 +24,60 @@ class Quantization(nn.Module):
...
+    ):
         super().__init__()
         self.embedding_dim = embedding_dim
         self.num_embeddings = num_embeddings
 
         self.embedding = nn.Embedding(self.num_embeddings, self.embedding_dim)
-        self.embedding.weight.data.uniform_(
-            -1 / self.num_embeddings, 1 / self.num_embeddings
-        )
+        # These are used in EMA update of embedding weights as m, N, and lambda, respectively, from Oord et al.
```
langong347, line 51: Just add a line above this class referencing the author's code for tracking purpose?

langong347 (commented):
I like the original author's implementation and the clear notations here. But some features from MUGEN's implementation/OpenAI are not captured. The end state could be a combination of both approaches, taking the best parts of each.

torchmultimodal/modules/layers/quantization.py
```diff
@@ -24,24 +24,60 @@ class Quantization(nn.Module):
...
+        self._code_avg = self.embedding.weight.detach().clone()
+        self._code_usage = torch.zeros(self.num_embeddings)
+        self._decay = decay
+        # Used in Laplace smoothing of code usage
+        self._epsilon = epsilon
+
+        # Flag to track if we need to initialize embedding with encoder output
+        self._is_embedding_init = False
```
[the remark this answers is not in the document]
RdoubleA: I was following the convention of naming boolean flags with "is***", but let me know if aligning with MUGEN's flag name is more important to avoid confusion

torchmultimodal/modules/layers/quantization.py
RdoubleA: if the model needed to be fine tuned, would these intermediate values need to be restored in `state_dict` after pre-training? I also noticed that in the MUGEN implementation they also include the embedding itself in register buffer, but to me it makes more sense as model parameters.

RdoubleA: actually, I see that buffers should be used for parameters that are not trained by the optimizer, but they are still restored in state_dict when the model is loaded (relevant forum [post](https://discuss.pytorch.org/t/what-is-the-difference-between-register-buffer-and-register-parameter-of-nn-module/32723)). In that case it would make sense for the embedding to be in the buffer

RdoubleA:
> I like the original author's implementation and the clear notations here. But some features from MUGEN's implementation are not captured. The end state could be a combination of both versions, taking the best parts of each.

Let me know what features from MUGEN I may have overlooked. Minus codebook restarts, this seems to capture all the steps in [MUGEN's EMA update](https://github.com/mugen-org/MUGEN_baseline/blob/e3507f071080e7c537fe3ac10212482563b7d3db/lib/models/video_vqvae/vqvae.py#L205)

test/modules/layers/test_quantization.py
```diff
@@ -38,11 +38,14 @@ def setUp(self):
...
-            num_embeddings=self.num_embeddings, embedding_dim=self.embedding_dim
+            num_embeddings=self.num_embeddings,
+            embedding_dim=self.embedding_dim,
+            decay=0.5,
         )
-        self.vq.embedding = nn.Embedding.from_pretrained(self.embedding_weights)
 
     def test_quantized_output(self):
+        self.vq.embedding = self.embedding_weights
+        self.vq._is_embedding_init = True
```
ebsmothers, line 48: nit: does _is_embedding_init need to be private? In general I would try to not be setting variables that we claim are private like this. Obviously it's not really private but the intention is implied by the leading underscore (which is why this is only a nit)

torchmultimodal/modules/layers/quantization.py
```diff
@@ -24,24 +25,61 @@ class Quantization(nn.Module):
...
+    def _tile(self, x):
+        # Repeat encoder vectors in cases where the encoder output does not have enough vectors
+        # to initialize the codebook on first forward pass
+        num_encoder_vectors, num_channels = x.shape
+        if num_encoder_vectors < self.embedding_dim:
+            num_repeats = (
+                self.num_embeddings + num_encoder_vectors - 1
+            ) // num_encoder_vectors
+            # Add a small amount of noise to repeated vectors
+            std = 0.01 / torch.sqrt(num_channels)
```
ebsmothers, line 79: Is this value specific to the MUGEN implementation? Wondering if we could either generalize this implementation or explain why we're adding noise in this way

test/modules/layers/test_quantization.py
```diff
@@ -38,11 +38,14 @@ def setUp(self):
         )
 
         self.vq = Quantization(
-            num_embeddings=self.num_embeddings, embedding_dim=self.embedding_dim
+            num_embeddings=self.num_embeddings,
+            embedding_dim=self.embedding_dim,
+            decay=0.5,
```
ebsmothers, line 43: super nit but I would use a different value than 0.5 as it can mask errors with decay and 1-decay mixups

test/modules/layers/test_quantization.py
```diff
@@ -84,23 +87,72 @@ def test_preprocess(self):
...
+
+        actual_code_avg = self.vq.code_avg
+        expected_code_avg = actual_weight
+        assert_expected(actual_code_avg, expected_code_avg)
+
+        actual_code_usage = self.vq.code_usage
+        expected_code_usage = torch.ones(self.num_embeddings)
+        assert_expected(actual_code_usage, expected_code_usage)
+
+    def test_ema_update(self):
```
ebsmothers, line 131: It might also make sense to add a unit test where you call .backward() to confirm that the buffers behave as expected

test/modules/layers/test_quantization.py
RdoubleA: for the purposes of the test I had to set a private variable, but I would imagine for general use I wouldn't want this variable to be modified without knowledgeable intent

test/modules/layers/test_quantization.py
RdoubleA: good point

test/modules/layers/test_quantization.py
RdoubleA: good call

torchmultimodal/modules/layers/quantization.py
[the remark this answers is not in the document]
langong347: I wasn't referring to naming of the variable but the default value.

torchmultimodal/modules/layers/quantization.py
langong347: This is originally from OpenAI, see [here](https://github.com/openai/jukebox/blob/08efbbc1d4ed1a3cef96e08a931944c8b4d63bb3/jukebox/vqvae/bottleneck.py#L28).

There's no mentioning of why this particular way of adding noise in their [paper](https://arxiv.org/pdf/2005.00341.pdf) but the logic looks generic to me:
1) Adding noise because we don't want the codebook vectors (constructed by repeating `x`) to look identical
2) `torch.rand_like(x)`: the noise follows a uniform distribution due to assumptions of VQVAE model
3) `* std`: normalize the noise along each codebook vector to be `U[0,1]`.

torchmultimodal/modules/layers/quantization.py
```diff
@@ -27,6 +28,11 @@ class Quantization(nn.Module):
     The embedding weights are trained with exponential moving average updates as described
     in original paper.
 
+    Code was largely inspired by a PyTorch implementation of the author's original code, found here:
```
langong347, line 31: Relocate this paragraph to be just under the license header? Doesn't seem to be specific to `QuantizationOutput`.

test/modules/layers/test_quantization.py
```diff
@@ -121,18 +120,18 @@ def test_embed_init(self):
...
+        actual_code_avg = self.vq.code_avg
         expected_code_avg = actual_weight
         assert_expected(actual_code_avg, expected_code_avg)
 
-        actual_code_usage = self.vq._code_usage
+        actual_code_usage = self.vq.code_usage
         expected_code_usage = torch.ones(self.num_embeddings)
         assert_expected(actual_code_usage, expected_code_usage)
 
     def test_ema_update(self):
```
langong347, line 131: `test_ema_update_embedding`? Since you already updated the test name for `init_embedding`.

torchmultimodal/modules/layers/quantization.py
```diff
@@ -117,36 +140,34 @@ def _ema_update_embedding(self, encoded_flat: Tensor, codebook_indices: Tensor):
         # Count how often each embedding vector was looked up
```
langong347, line 140: lowercase "count" here?

torchmultimodal/modules/layers/quantization.py
```diff
@@ -117,36 +140,34 @@ def _ema_update_embedding(self, encoded_flat: Tensor, codebook_indices: Tensor):
...
         # Calculate distances from each encoder, E(x), output vector to each embedding vector, e, ||E(x) - e||^2
-        distances = torch.cdist(encoded_flat, self.embedding.weight, p=2.0) ** 2
+        distances = torch.cdist(encoded_flat, self.embedding, p=2.0) ** 2
 
         # Encoding - select closest embedding vectors
         codebook_indices = torch.argmin(distances, dim=1)
 
         # Quantize
-        quantized_flat = self.embedding(codebook_indices)
+        quantized_flat = F.embedding(codebook_indices, self.embedding)
```
langong347, line 170: Is `F.embedding` better than simply indexing `self.embedding`?

torchmultimodal/modules/layers/quantization.py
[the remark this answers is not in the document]
langong347: Ok, I see. MUGEN was using `_need_init` which is the opposite to `_is_embedding_init`.

test/modules/layers/test_quantization.py
```diff
@@ -45,7 +44,7 @@ def setUp(self):
         )
 
     def test_quantized_output(self):
-        self.vq.embedding = nn.Embedding.from_pretrained(self.embedding_weights)
+        self.vq.embedding = self.embedding_weights
```
langong347, line 47: Since you are reusing `self.vq.embedding` for the subsequent tests, should we just move this into `setUp` in case one day we remove this assignment and the other tests will break?

test/modules/layers/test_quantization.py
```diff
@@ -103,14 +102,14 @@ def test_postprocess(self):
...
+    def test_init_embedding_and_preprocess(self):
         assert not self.vq._is_embedding_init, "embedding init flag not False initially"
 
         _, _ = self.vq._init_embedding_and_preprocess(self.encoded)
 
         assert self.vq._is_embedding_init, "embedding init flag not True after init"
 
-        actual_weight = self.vq.embedding.weight
+        actual_weight = self.vq.embedding
         expected_weight = torch.Tensor(
```
langong347, line 113: Can we add a test for the code path when `num_encoder_vectors < self.embedding_dim` so that the random initialization from `_tile` can be tested? I suppose with the current setup `encoded_flat` is `(6, 5)` and `embeddings` is `(4, 5)`?

```
# This is 2x5x3 
self.encode = ...
# This is 4x5
self.embedding = ...
```

test/modules/layers/test_quantization.py
langong347: Agreed

torchmultimodal/modules/layers/quantization.py
```diff
@@ -24,24 +25,61 @@ class Quantization(nn.Module):
...
+        decay: float = 0.99,
+        epsilon: float = 1e-7,
+    ):
         super().__init__()
+        # Embedding weights and parameters for EMA update will be registered to buffer, as they
+        # will not be updated by the optimizer but are still model parameters.
+        # code_usage and code_avg correspond with N and m, respectively, from Oord et al.
+        self.register_buffer("embedding", torch.randn(num_embeddings, embedding_dim))
+        self.register_buffer("code_usage", torch.zeros(num_embeddings))
+        self.register_buffer("code_avg", self.embedding.clone())
```
langong347, line 58: This one is giving `mypy` error. Try asking around in pytorch user group?

torchmultimodal/modules/layers/quantization.py
RdoubleA: self.embedding is no longer an `nn.Embedding`, just a Tensor initialized by `register_buffer`

torchmultimodal/modules/layers/quantization.py
RdoubleA: for what reason? this is the start of a new comment

test/modules/layers/test_quantization.py
RdoubleA: I only need pre-initialized weights for this test, the others depend on init_embedding and would ignore this anyway

langong347 (approved):

torchmultimodal/modules/layers/quantization.py
ebsmothers: Thanks for the explanation. So I guess I am more interested in why the value of 0.01 is being chosen here. It looks to me like this is actually a `N(0, 1e-4 * I)` distribution (modulo me probably screwing up some normalization). But this would be less of a standard (no pun intended) distribution. Lmk if I'm misunderstanding the implementation though

RdoubleA: I think the decision for 0.01 was mainly arbitrary to control the dispersion or magnitude of the noise. If it was true normal noise `N(0,1)` then that could change the encoder output too significantly, better to change it by `+/- 1e-4 / sqrt(num_channels)` most of the time instead

ebsmothers: Should 0.01 be a parameter then?

PR #50, opened 2022-05-21
[Lint] Upgrade usort to 1.0.2; add flake8 plugins

langong347:
Summary:
- Upgrade `usort` to align with Fbcode
- The following plugins are added to `flake8`:
    - [`flake8-bugbear`](https://github.com/PyCQA/flake8-bugbear) to align with fbcode flake8 on detecting bugs/design issues
    - [`pep8-naming`](https://github.com/PyCQA/pep8-naming): naming convention checker
- Consolidate lint/formater installation into `requirements-lint.txt`.

Test Plan:
- Verify that the plugins are installed:
```
pip install -r requirements-lint.txt
flake8 --version
4.0.1 (flake8-bugbear: 22.4.25, mccabe: 0.6.1, naming: 0.12.1, pycodestyle: 2.8.0, pyflakes: 2.4.0)
```

- Create a test file to invoke the corresponding errors:
```
test_flake8.py:4:6: N802 function name 'Incorrect_Function_Name' should be lowercase
test_flake8.py:9:9: B015 Pointless comparison. This comparison does nothing but waste CPU instructions. Either prepend `assert` or remove it.
test_flake8.py:10:5: B001 Do not use bare `except:`, it also catches unexpected events like memory errors, interrupts, system exit, and so on.  Prefer `except Exception:`.  If you're sure what you're doing, be explicit and write `except BaseException:`.
test_flake8.py:10:5: E722 do not use bare 'except'```
```

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -63,6 +63,7 @@ Ideally, flake and ufmt should be run via pre-commit hooks.
 But if for some reason you want to run them separately follow this:
 
 ```
+pip install -r requirements-lint.txt
 flake8 (examples|test|torchmultimodal)
 ufmt format (examples|test|torchmultimodal)
 ```
@@ -70,7 +71,6 @@ ufmt format (examples|test|torchmultimodal)
 Alternatively, you can run on only those files you have modified, e.g.
 
 ```
-pip install flake8==4.0.1 ufmt==1.3.0 black==21.4b2 usort==1.0.2
 flake8 `git diff main --name-only`
 ufmt format `git diff main --name-only`
 ```
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,0 +1,9 @@
+[run]
+omit =
+    *test/*
+    *examples/*
+
+[report]
+omit =
+    *test/*
+    *examples/*
--- a/.flake8
+++ b/.flake8
@@ -1,6 +1,6 @@
 [flake8]
 # Suggested config from pytorch that we can adapat
-select = B,C,E,F,P,T4,W,B9
+select = B,C,E,F,N,P,T4,W,B9
 max-line-length = 120
 # C408 ignored because we like the dict keyword argument syntax
 # E501 is not flexible enough, we're using B950 instead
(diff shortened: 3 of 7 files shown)


CONTRIBUTING.md
```diff
@@ -63,14 +63,14 @@ Ideally, flake and ufmt should be run via pre-commit hooks.
 But if for some reason you want to run them separately follow this:
 
 ```
+pip install -r requirements-lint.txt
```
ebsmothers, line 66: Just wondering, do we really need a separate requirements file just for lint? I don't really see this in other repos (besides pytorch core, which is a much larger repo anyways) and wonder if we're adding unnecessary files at the top-level of the repo by doing this

langong347: I suppose other repo uses just a vanilla flake8 that does not involve plug-ins to be installed. If the latter, the requirement file is just to consolidate all the plug-ins in one place.

langong347: Also, as long as we don't specify it in `setup.py`, the additional top-level requirement files won't go into installation or packaging.

ebsmothers: That's reasonable. Maybe I can instead ask a strictly stronger question: do we really need to allow users to run the lint suite manually 🙂 ?  Basically my concern is that we will give the appearance of extra friction in development by adding this file, which we do not want.

langong347: Manual, you mean running lint on the command line? The developer can also use the correct version of flake8 inside their IDE. 

Where is the extra friction though? If I want to lint on my own, I need to know what the library uses.

ebsmothers: > Manual, you mean running lint on the command line?

Yes, by following the instructions we provide in CONTRIBUTING.

> Where is the extra friction though?

So it is not actually extra friction. But it gives the _appearance_ of extra friction. We will now have three separate files all containing "requirements" in their names, all of which are present on the GitHub landing page. Just from a first-time user's perspective, if I engage in the usual exercise of looking at various files to better understand the repo and I see this many different types of requirements I would feel a bit daunted. And I think the current complexity/maturity of our repo is not high enough to justify a bunch of disjoint sets of requirements like this.

PR #57, opened 2022-05-26
Adding a MUGEN utility for permuting and implemented in codebook

RdoubleA:
Summary:
`shift_dim` is a commonly used utility throughout the MUGEN codebase. Including it in a small separate PR so subsequent PRs for MUGEN can use `shift_dim` quickly.

Test plan:
`pytest test/modules/layers/test_codebook.py -vv`
`pytest test/utils/test_common.py -vv`

### DIFF
--- a/torchmultimodal/utils/common.py
+++ b/torchmultimodal/utils/common.py
@@ -11,6 +11,7 @@ from dataclasses import fields
 from typing import Optional
 
 import torch
+from torch import Tensor
 
 
 def get_current_device():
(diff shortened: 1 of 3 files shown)


torchmultimodal/utils/common.py
```diff
@@ -20,6 +21,50 @@ def get_current_device():
         return torch.device("cpu")
 
 
+def shift_dim(
```
langong347, line 24: Even though we might have already hit the code path in `test_preprocess` for the codebook, it's better to have a dedicated test for this utility.

langong347 (approved):

PR #60, opened 2022-06-01
Modularize TextEmbeddings

joecummings:
Summary:
- Moved `TextEmbeddings` class from `flava.py` to its own module
- Created `embeddings` dir to hold `text_embeddings` and future embeddings modules
- Created test

Differential Revision: D36634188


ebsmothers:
@joecummings looks like the lint and unit test checks are failing. Can you take a look?

PR #61, opened 2022-06-02
Add MDETR image backbone

ebsmothers:
* #110
* #109
* #77
* #62
* __->__ #61

This commit adds support for the ResNet-101 image backbone from MDETR.
This is basically just a wrapper around TorchVision's ResNet class
with support for fetching intermediate layers from the model so that the final pooling and
FC layers are not used. We also add the nested tensor and frozen batch norm classes from
the MDETR repo, ultimately we may factor out the nested tensor class later on.

Also added a test to confirm the results match those obtained by using MDETR's image backbone directly.

This is an initial version, will update docstrings etc if the high-level implementation looks good.

Tested with
```
$ python -m pytest -v test/modules/encoders/test_mdetr_image_encoder.py
test/modules/encoders/test_mdetr_image_encoder.py::TestMDETRImageEncoder::test_resnet_101_forward PASSED

Differential Revision: [D37390044](https://our.internmc.facebook.com/intern/diff/D37390044)

### CODE REVIEW

File: torchmultimodal/modules/encoders/mdetr_image_encoder.py [hunk 1]
```diff
@@ -0,0 +1,46 @@
...
+from torch import nn
+from torchmultimodal.modules.layers.normalizations import FrozenBatchNorm2d
+from torchmultimodal.utils.common import NestedTensor
+from torchvision.models._utils import IntermediateLayerGetter
+from torchvision.models.resnet import ResNet101_Weights
+
+# At this point this class is basically just a wrapper
+# around the backbone to support tensor list and IntermediateLayerGetter.
+# We should refactor to pass tensor and mask separately
+class MDETRBackbone(nn.Module):
```
RdoubleA, line 20: is the plan to change the name of this class? Since this is a general wrapper around IntermediateLayerGetter and NestedTensor

ebsmothers: Good point. To make it truly general we can probably pass return_layers as an argument, otherwise this is too specific to MDETR's ResNet implementation.

torchmultimodal/modules/encoders/mdetr_image_encoder.py
```diff
@@ -0,0 +1,46 @@
...
+        out = OrderedDict()
+        for name, x in xs.items():
+            mask = F.interpolate(
+                tensor_list.mask[None].float(), size=x.shape[-2:]
+            ).bool()[0]
+            out[name] = NestedTensor(x, mask)
+        return out
+
+
+def _mdetr_resnet101_backbone():
```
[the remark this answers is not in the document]
ebsmothers: Fair question. This is not the top-level MDETR model class, so is less likely to be imported directly. But in general I think any builder functions should probably be public, so let me change it.

torchmultimodal/modules/encoders/mdetr_image_encoder.py
```diff
@@ -0,0 +1,46 @@
...
+# around the backbone to support tensor list and IntermediateLayerGetter.
+# We should refactor to pass tensor and mask separately
+class MDETRBackbone(nn.Module):
+    def __init__(self, body: nn.Module):
+        super().__init__()
+        # Note that we need this to skip pooler, flatten, and FC layers in
+        # the standard ResNet implementation.
+        self.body = IntermediateLayerGetter(body, return_layers={"layer4": 0})
+
+    def forward(self, tensor_list):
```
ankitade, line 27: whats the type of tensor_list

torchmultimodal/modules/encoders/mdetr_image_encoder.py
```diff
@@ -0,0 +1,46 @@
...
+        xs = self.body(tensor_list.tensors)
+        out = OrderedDict()
+        for name, x in xs.items():
+            mask = F.interpolate(
+                tensor_list.mask[None].float(), size=x.shape[-2:]
+            ).bool()[0]
+            out[name] = NestedTensor(x, mask)
```
ankitade, line 34: why is nested tensor generic thing?

torchmultimodal/modules/encoders/mdetr_image_encoder.py
```diff
@@ -0,0 +1,46 @@
...
+        return out
+
+
+def _mdetr_resnet101_backbone():
+    # TODO: maybe support passing arg for last dilation operation as in MDETR repo
+    body = getattr(torchvision.models, "resnet101")(
```
ankitade, line 40: why use getattr? cant u just import

torchmultimodal/modules/layers/normalizations.py
```diff
@@ -20,3 +21,56 @@ def forward(self, x):
             self.eps,
         )
         return output.type_as(x)
+
+
+class FrozenBatchNorm2d(nn.Module):
```
ankitade, line 26: is this very mdetr resnet specific? if yes, consider moving to that module

ankitade (commented):
the input / output format are somewhat unintuitive to a lay man. where does the complexity come from?

torchmultimodal/modules/encoders/mdetr_image_encoder.py
ebsmothers: It's really just a wrapper around a list of tensor plus mask. I think we can factor this out later on

torchmultimodal/modules/layers/normalizations.py
ebsmothers: It's also used in the ResNet benchmark from FasterRCNN, for example. But not sure how widespread the usage is in general. I think moving it to the MDETR ResNet module should be reasonable

torchmultimodal/modules/encoders/mdetr_image_encoder.py
ebsmothers: This is just how it's done in MDETR. As far as I can tell it shouldn't make a huge difference

torchmultimodal/modules/encoders/mdetr_image_encoder.py
ebsmothers: It should be a `NestedTensor`. Will add the type info

torchmultimodal/modules/encoders/mdetr_image_encoder.py
```diff
@@ -0,0 +1,146 @@
...
+        mask = F.interpolate(image_masks[None].float(), size=tensor.shape[-2:]).bool()[
+            0
+        ]
+        return tensor, mask
+
+
+def mdetr_resnet101_backbone() -> MaskedIntermediateLayer:
+    body = resnet101(
+        replace_stride_with_dilation=[False, False, False],
+        weights=ResNet101_Weights.IMAGENET1K_V1,
```
ankitade, line 141: can we make this part of function args

torchmultimodal/modules/encoders/mdetr_image_encoder.py
```diff
@@ -0,0 +1,146 @@
...
+        ).flatten(3)
+        pos_y = torch.stack(
+            (pos_y[:, :, :, 0::2].sin(), pos_y[:, :, :, 1::2].cos()), dim=4
+        ).flatten(3)
+        pos = torch.cat((pos_y, pos_x), dim=3).permute(0, 3, 1, 2)
+        return pos
+
+
+class MaskedIntermediateLayer(nn.Module):
+    def __init__(self, body: nn.Module, intermediate_layers: Dict[str, Any]):
```
ankitade, line 118: isnt it Dict[str,int]?

torchmultimodal/modules/encoders/mdetr_image_encoder.py
```diff
@@ -0,0 +1,146 @@
...
+        assert (
+            len(intermediate_layers.keys()) == 1
+        ), "IntermediateLayerGetter for multiple layers is not supported"
+        self.body = IntermediateLayerGetter(body, return_layers=intermediate_layers)
+
+    def forward(
+        self, images: torch.Tensor, image_masks: torch.Tensor
+    ) -> Tuple[Tensor, Tensor]:
+        out = self.body(images)
+        tensor = out[next(reversed(out))]
```
ankitade, line 131: why do we need this?

torchmultimodal/modules/encoders/mdetr_image_encoder.py
```diff
@@ -0,0 +1,146 @@
...
+        mask = F.interpolate(image_masks[None].float(), size=tensor.shape[-2:]).bool()[
```
ankitade, line 132: some comments

torchmultimodal/modules/encoders/mdetr_image_encoder.py
```diff
@@ -0,0 +1,146 @@
...
+        b = self.bias.reshape(1, -1, 1, 1)  # type: ignore
+        rv = self.running_var.reshape(1, -1, 1, 1)  # type: ignore
+        rm = self.running_mean.reshape(1, -1, 1, 1)  # type: ignore
+        eps = 1e-5
+        scale = w * (rv + eps).rsqrt()
+        bias = b - rm * scale
+        return x * scale + bias
+
+
+class PositionEmbeddingSine(nn.Module):
```
ankitade, line 70: neither used nor tested?

torchmultimodal/modules/encoders/mdetr_image_encoder.py
```diff
@@ -0,0 +1,146 @@
...
+            error_msgs,
+        )
+
+    def forward(self, x: torch.Tensor):
+        # move reshapes to the beginning to make it fuser-friendly
+        # ignore mypy errors because fixing them would break checkpoint loading
+        w = self.weight.reshape(1, -1, 1, 1)  # type: ignore
+        b = self.bias.reshape(1, -1, 1, 1)  # type: ignore
+        rv = self.running_var.reshape(1, -1, 1, 1)  # type: ignore
+        rm = self.running_mean.reshape(1, -1, 1, 1)  # type: ignore
```
ankitade, line 63: why not just do these in init?

torchmultimodal/modules/encoders/mdetr_image_encoder.py
```diff
@@ -0,0 +1,146 @@
...
+        prefix,
+        local_metadata,
+        strict,
+        missing_keys,
+        unexpected_keys,
+        error_msgs,
+    ):
+        num_batches_tracked_key = prefix + "num_batches_tracked"
+        if num_batches_tracked_key in state_dict:
+            del state_dict[num_batches_tracked_key]
```
ankitade, line 45: why do we need this? add comments

torchmultimodal/modules/encoders/mdetr_image_encoder.py
```diff
@@ -0,0 +1,146 @@
...
+
+import torch
+import torch.nn.functional as F
+from torch import nn, Tensor
+from torchvision.models._utils import IntermediateLayerGetter
+from torchvision.models.resnet import resnet101, ResNet101_Weights
+
+
+class FrozenBatchNorm2d(nn.Module):
+    """
```
ankitade, line 18: its on your todo but add docstrings

torchmultimodal/modules/encoders/mdetr_image_encoder.py
ebsmothers: Just refactored so this is no longer relevant. Now we only accept a single layer, since only segmentation task requires more than one intermediate layer and removing support for multiple intermediate layers simplifies the code a lot.

torchmultimodal/modules/encoders/mdetr_image_encoder.py
ebsmothers: Because IntermediateLayerGetter returns a dict mapping layer names to layer outputs. So even if we only pass one layer we need to extract the tensor in this way

torchmultimodal/modules/encoders/mdetr_image_encoder.py
ebsmothers: So I added something in the docstring, lmk if it's sufficient

torchmultimodal/modules/encoders/mdetr_image_encoder.py
ebsmothers: Not tested explicitly here, but its used in #77 (and tested there by loading and calling forward on the full model). In the original repo they tie this to the backbone, but I decided to keep it separate. But it is still in here as a holdover. Happy to move it elsewhere if you think that's clearer

torchmultimodal/modules/encoders/mdetr_image_encoder.py
ebsmothers: done

torchmultimodal/modules/encoders/mdetr_image_encoder.py
```diff
@@ -0,0 +1,157 @@
...
+
+        pos_x = x_embed[:, :, :, None] / dim_t
+        pos_y = y_embed[:, :, :, None] / dim_t
+        pos_x = torch.stack(
+            (pos_x[:, :, :, 0::2].sin(), pos_x[:, :, :, 1::2].cos()), dim=4
+        ).flatten(3)
+        pos_y = torch.stack(
+            (pos_y[:, :, :, 0::2].sin(), pos_y[:, :, :, 1::2].cos()), dim=4
+        ).flatten(3)
+        pos = torch.cat((pos_y, pos_x), dim=3).permute(0, 3, 1, 2)
```
ankitade, line 110: shape comment

torchmultimodal/modules/encoders/mdetr_image_encoder.py
```diff
@@ -0,0 +1,157 @@
...
+        num_pos_feats: int = 64,
+        temperature: int = 10000,
+        normalize: bool = False,
+        scale: float = None,
+    ):
+        super().__init__()
+        self.num_pos_feats = num_pos_feats
+        self.temperature = temperature
+        self.normalize = normalize
+        if scale is not None and normalize is False:
```
ankitade, line 82: can u just take in scale as optional float and do away with normalize?

torchmultimodal/modules/encoders/mdetr_image_encoder.py
```diff
@@ -0,0 +1,157 @@
...
+    generalized to work on images.
+
+    Args:   num_pos_feats (int): Number of positional features
+                (should be half the output embedding size). Default = 64
+            temperature (int): Base for generating frequency mesh. Default = 10000
+            normalize (bool): Whether to normalize the image grid to values in [0,1].
+                Default = False
+            scale (float): Scaling factor when performing normalization. Default = None
+
+    Inputs: mask (Tensor): Padding mask (used to determine size of each image in batch).
```
ankitade, line 68: shape info?

torchmultimodal/modules/encoders/mdetr_image_encoder.py
ankitade: maybe call it 2dposition embeddings

torchmultimodal/modules/encoders/mdetr_image_encoder.py
ebsmothers: Added in the docstring

ebsmothers:
@ebsmothers has imported this pull request.  If you are a Facebook employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D37390044).

ankitade (approved):

ebsmothers:
@ebsmothers has imported this pull request.  If you are a Facebook employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D37390044).

ebsmothers:
@ebsmothers has imported this pull request.  If you are a Facebook employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D37390044).

PR #62, opened 2022-06-02
Add MDETR text encoder

ebsmothers:
* (to be filled)

This commit adds support for MDETR's RoBERTa text encoder.
We make some modifications to TorchText's RoBERTa encoder to make
things more modular and ensure the implementation matches the one from
Hugging Face (as used by MDETR). The changes include

1) A class MDETRTextEmbeddings. This is similar to the TextEmbeddings class from
FLAVA, but handles position IDs differently.

2) A WrappedTransformerEncoder class. This class will take MDETR text embeddings
and feed them through the transformer encoder. The main difference between this class
and the TorchText one is that the TorchText class includes embeddings as part of the encoder.

3) A class MDETRTextEncoder to stitch (1) and (2) together

We add a test for each of these three classes and compare to the result obtained when
loading directly from a pretrained MDETR checkpoint.

Similar to the image backbone commit, once the changes look good at a high level I will
add docstrings and the like.

Tested with
```
$ python -m pytest -v test/modules/encoders/test_mdetr_text_encoder.py
    test/modules/encoders/test_mdetr_text_encoder.py::TestMDETRTextEncoder::test_mdetr_text_embeddings PASSED
    test/modules/encoders/test_mdetr_text_encoder.py::TestMDETRTextEncoder::test_mdetr_text_encoder PASSED
    test/modules/encoders/test_mdetr_text_encoder.py::TestMDETRTextEncoder::test_mdetr_wrapped_transformer PASSED
```

### DIFF
--- a/test/modules/encoders/test_mdetr_text_encoder.py
+++ b/test/modules/encoders/test_mdetr_text_encoder.py
@@ -1,0 +1,195 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+import unittest
+from copy import deepcopy
+
+import torch
+from test.test_utils import assert_expected, set_rng_seed
+from torchmultimodal.modules.encoders.mdetr_text_encoder import (
+    MDETRTextEmbeddings,
+    MDETRTextEncoder,
+    WrappedTransformerEncoder,
+)
+from torchmultimodal.utils.common import filter_dict
+from torchtext.models.roberta.bundler import ROBERTA_BASE_ENCODER
+
+
+class TestMDETRTextEncoder(unittest.TestCase):
+    def setUp(self):
+        self.max_position_embeddings = 514
+        self.hidden_size = 768
+        self.embeddings = MDETRTextEmbeddings(
+            hidden_size=self.hidden_size,
+            vocab_size=50265,
+            pad_token_id=1,
+            type_vocab_size=1,
+            max_position_embeddings=self.max_position_embeddings,
+            layer_norm_eps=1e-05,
+            hidden_dropout_prob=0.1,
+        )
+        set_rng_seed(0)
+        self._populate_embedding_weights()
+
+        self.roberta_encoder = ROBERTA_BASE_ENCODER.get_model()
+        # Remove extra args due to TorchText RoBERTa encoder's forward taking in tokens instead of embeddings
+        wrapped_args = filter_dict(
+            lambda x: x not in ["vocab_size", "padding_idx", "max_seq_len", "scaling"],
+            ROBERTA_BASE_ENCODER.encoderConf.__dict__,
+        )
+        self.wrapped_transformer_encoder = WrappedTransformerEncoder(**wrapped_args)
+        self._populate_wrapped_transformer_weights()
+
+        self.text_encoder = MDETRTextEncoder(
+            embeddings=self.embeddings, encoder=self.wrapped_transformer_encoder
+        )
+        self.text_encoder.eval()
+
+        self.input_ids = torch.tensor(
+            [
+                [0, 100, 64, 192, 5, 3778, 2, 1, 1, 1, 1, 1, 1, 1, 1, 1],
+                [
+                    0,
+                    1708,
+                    190,
+                    114,
+                    38,
+                    1395,
+                    192,
+                    5,
+                    3778,
+                    6,
+                    38,
+                    216,
+                    14,
+                    24,
+                    8785,
+                    2,
+                ],
+            ],
+            dtype=torch.int,
+        )
+        self.attention_mask = torch.tensor(
+            [
+                [1, 1, 1, 1, 1, 1, 1, 0, 0, 0, 0, 0, 0, 0, 0, 0],
+                [1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1],
+            ],
+            dtype=torch.int,
+        )
+        self.batch_size, self.input_length = self.input_ids.size()
+
+    def test_mdetr_text_embeddings(self):
+        expected = torch.Tensor(
+            [
+                0.62926,
+                0.71966,
+                0.59531,
+                0.54416,
+                0.61227,
+                0.67181,
+                0.61707,
+                0.47298,
+                0.72058,
+                0.68760,
+                0.75449,
+                0.52463,
+                0.49111,
+                0.57840,
+                0.44969,
+                0.71950,
+            ]
+        )
+        out = self.embeddings(self.input_ids, self.attention_mask)
+        actual = out[1, :, 1]
+        self.assertEqual(
+            out.size(), (self.batch_size, self.input_length, self.hidden_size)
+        )
+        assert_expected(actual, expected, rtol=0.0, atol=1e-4)
+
+    def test_mdetr_wrapped_transformer(self):
+        set_rng_seed(0)
+        inp = torch.rand((2, 16, 768))
+        expected = torch.Tensor(
+            [
+                0.10437,
+                0.10304,
+                0.10998,
+                0.10926,
+                0.10760,
+                0.10326,
+                0.10478,
+                0.10586,
+                0.10500,
+                0.11583,
+                0.10339,
+                0.10585,
+                0.10405,
+                0.10533,
+                0.11012,
+                0.11071,
+            ]
+        )
+        out = self.wrapped_transformer_encoder(inp, self.attention_mask)
+        actual = out[1, :, 1]
+        self.assertEqual(
+            out.size(), (self.batch_size, self.input_length, self.hidden_size)
+        )
+        assert_expected(actual, expected, rtol=0.0, atol=1e-4)
+
+    def test_mdetr_text_encoder(self):
+        expected = torch.Tensor(
+            [
+                0.04690,
+                0.06246,
+                0.05839,
+                0.05727,
+                0.06413,
+                0.05640,
+                0.06599,
+                0.04888,
+                0.07137,
+                0.05381,
+                0.06141,
+                0.06225,
+                0.04788,
+                0.06256,
+                0.05058,
+                0.06433,
+            ]
+        )
+        out = self.text_encoder(self.input_ids, self.attention_mask)
+        actual = out[1, :, 1]
+        self.assertEqual(
+            out.size(), (self.batch_size, self.input_length, self.hidden_size)
+        )
+        assert_expected(actual, expected, rtol=0.0, atol=1e-4)
+
+    # Unlike with the encoder where all weights are already present in the TorchText checkpoint,
+    # for the embedding we need to construct our own weights since word_embeddings are not present
+    # in the TorchText model bundle.
+    def _populate_embedding_weights(self):
+        embeddings_state_dict = {}
+        for k, v in self.embeddings.state_dict().items():
+            # These have already been set on init and shouldn't be changed
+            if k == "position_ids":
+                embeddings_state_dict[k] = v
+            # Set these to match the HF RoBERTa weights
+            elif k == "token_type_embeddings.weight":
+                embeddings_state_dict[k] = torch.zeros(v.size())
+            # Set all other weights randomly
+            else:
+                embeddings_state_dict[k] = torch.rand(v.size())
+        self.embeddings.load_state_dict(embeddings_state_dict)
+
+    def _populate_wrapped_transformer_weights(self):
+        transformer_state_dict = deepcopy(
+            self.roberta_encoder.encoder.transformer.state_dict()
+        )
+        # Remove embedding weights, but keep final layer norm weights
+        wrapped_state_dict = filter_dict(
+            lambda x: "embedding" not in x or "layer_norm" in x, transformer_state_dict
+        )
+        self.wrapped_transformer_encoder.load_state_dict(wrapped_state_dict)
--- a/torchmultimodal/modules/encoders/mdetr_text_encoder.py
+++ b/torchmultimodal/modules/encoders/mdetr_text_encoder.py
@@ -1,0 +1,223 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+from typing import List, Optional, Union
+
+import torch
+from torch import nn, Tensor
+
+
+def create_position_ids_from_input_ids(input_ids, padding_idx):
+    """
+    Replace non-padding symbols with their position numbers. Position numbers begin at padding_idx+1. Padding symbols
+    are ignored. This is modified from fairseq's `utils.make_positions`.
+
+    Args:
+        x: torch.Tensor x:
+
+    Returns: torch.Tensor
+    """
+    # The series of casts and type-conversions here are carefully balanced to both work with ONNX export and XLA.
+    mask = input_ids.ne(padding_idx).int()
+    incremental_indices = torch.cumsum(mask, dim=1).type_as(mask) * mask
+    return incremental_indices.long() + padding_idx
+
+
+# TODO (later): Possibly refactor into a common class with FLAVA text embeddings class
+class MDETRTextEmbeddings(nn.Module):
+    """This class is identical to FLAVA's TextEmbeddings class except for its handling of position IDs."""
+
+    def __init__(
+        self,
+        hidden_size: int = 768,
+        vocab_size: int = 30522,
+        pad_token_id: int = 0,
+        type_vocab_size: int = 2,
+        max_position_embeddings: int = 512,
+        layer_norm_eps: float = 1e-12,
+        hidden_dropout_prob: float = 0.1,
+    ):
+        super().__init__()
+        self.word_embeddings = nn.Embedding(
+            vocab_size, hidden_size, padding_idx=pad_token_id
+        )
+        self.padding_idx = pad_token_id
+        self.position_embeddings = nn.Embedding(
+            max_position_embeddings, hidden_size, padding_idx=self.padding_idx
+        )
+        self.token_type_embeddings = nn.Embedding(type_vocab_size, hidden_size)
+
+        # self.LayerNorm is not snake-cased to stick with TensorFlow model variable name and be able to load
+        # any TensorFlow checkpoint file
+        self.LayerNorm = nn.LayerNorm(hidden_size, eps=layer_norm_eps)
+        self.dropout = nn.Dropout(hidden_dropout_prob)
+        # position_ids (1, len position emb) is contiguous in memory and exported when serialized
+        position_ids_range = torch.arange(max_position_embeddings).expand((1, -1))
+        self.register_buffer("position_ids", position_ids_range.clone())
+        self.register_buffer(
+            "token_type_ids",
+            torch.zeros(position_ids_range.size(), dtype=torch.long),
+            persistent=False,
+        )
+
+    def forward(
+        self,
+        input_ids: Tensor,
+        attention_mask: Optional[Tensor] = None,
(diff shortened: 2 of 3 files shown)


torchmultimodal/modules/encoders/mdetr_text_encoder.py
```diff
@@ -0,0 +1,223 @@
...
+            position_ids = create_position_ids_from_input_ids(
+                input_ids, self.padding_idx
+            )
+
+        if attention_mask is None:
+            attention_mask = torch.ones(((batch_size, seq_length)), device=device)
+
+        # Setting the token_type_ids to the registered buffer in constructor where it is all zeros, which usually occurs
+        # when its auto-generated, registered buffer helps users when tracing the model without passing token_type_ids, solves
+        # issue #5664
```
ankitade, line 88: i know its copied from flava but please remove the random comments

torchmultimodal/modules/encoders/mdetr_text_encoder.py
```diff
@@ -0,0 +1,223 @@
...
+        position_embeddings = self.position_embeddings(position_ids)
+        embeddings += position_embeddings
+
+        embeddings = self.LayerNorm(embeddings)
+        embeddings = self.dropout(embeddings)
+        return embeddings
+
+
+# Wrap TorchText's TransformerEncoder to take in embeddings instead of tokens
+class WrappedTransformerEncoder(nn.Module):
```
ankitade, line 118: this is more of a copy of torch text roberta instead of wrapper around it?

torchmultimodal/utils/common.py
```diff
@@ -113,3 +113,9 @@ def from_tensor_list(cls, tensor_list, do_round=False):
 
     def __repr__(self):
         return repr(self.tensors)
+
+
+def filter_dict(
```
ankitade, line 118: where is this used?

torchmultimodal/modules/encoders/mdetr_text_encoder.py
```diff
@@ -0,0 +1,223 @@
...
+
+
+class MDETRTextEncoder(nn.Module):
+    def __init__(self, embeddings: nn.Module, encoder: nn.Module):
+        super().__init__()
+        self.embeddings = embeddings
+        self.encoder = encoder
+
+    # Note: MDETR's RoBERTa encoder also returns pooler outputs in forward, but
+    # these are only used for contrastive loss. Since contrastive loss is not used anywhere,
```
ankitade, line 198: what do u mean by contrastive loss not being used anywhere?

torchmultimodal/modules/encoders/mdetr_text_encoder.py
```diff
@@ -0,0 +1,223 @@
...
+    """
+    Replace non-padding symbols with their position numbers. Position numbers begin at padding_idx+1. Padding symbols
+    are ignored. This is modified from fairseq's `utils.make_positions`.
+
+    Args:
+        x: torch.Tensor x:
+
+    Returns: torch.Tensor
+    """
+    # The series of casts and type-conversions here are carefully balanced to both work with ONNX export and XLA.
```
ankitade, line 23: why do we care about onnx and xla?

ankitade (commented):
generally worried about the amount of code we have to copy paste (i know why u are forced to do it). 2 things to resolve soon:
1. having common bert embeddings. I can work with you on this.
2. copy pasting from Torch text roberta iiuc. Let's raise this issue as a post in domains.

torchmultimodal/modules/encoders/mdetr_text_encoder.py
ebsmothers: So there are two contrastive losses: contrastive alignment loss and contrastive loss. Contrastive alignment loss is used but as far as I can tell contrastive loss is not. So I am referring to the latter here

torchmultimodal/utils/common.py
ebsmothers: Used in the test right now, but in general will be used for state_dict mapping stuff

torchmultimodal/modules/encoders/mdetr_text_encoder.py
ebsmothers: That's fair. The main difference is to take in the embeddings instead of tokens. Either way I can probably come up with a better name here

torchmultimodal/modules/encoders/mdetr_text_encoder.py
ebsmothers: This is copy-paste, let me remove it

torchmultimodal/modules/encoders/mdetr_text_encoder.py
```diff
@@ -0,0 +1,223 @@
...
+    mask = input_ids.ne(padding_idx).int()
+    incremental_indices = torch.cumsum(mask, dim=1).type_as(mask) * mask
+    return incremental_indices.long() + padding_idx
+
+
+# TODO (later): Possibly refactor into a common class with FLAVA text embeddings class
```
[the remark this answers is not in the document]
ebsmothers: So I think there is a slight difference. For BERT embeddings the position IDs are just a slice of an `arange`, but for RoBERTa there is an offset. It's a really small difference so we can definitely generalize to use a common implementation for both.

torchmultimodal/modules/encoders/mdetr_text_encoder.py
```diff
@@ -0,0 +1,240 @@
...
+
+    def _forward_return_all_layers(
+        self,
+        embeddings: torch.Tensor,
+        attn_mask: Optional[torch.Tensor] = None,
+    ) -> List[torch.Tensor]:
+        encoded = embeddings
+        states = [encoded]
+        for layer in self.layers.layers:
+            encoded = layer(encoded, src_key_padding_mask=attn_mask)
```
ebsmothers, line 152: @abhinavarora @parmeet this may be the source of the discrepancy

parmeet: @ebsmothers I am not sure what attn_mask is referring to here? could you provide pointer to the use site to understand what is being passed in the forward function. (Ideally this should be the mask which indicates which tokens are padding tokens i.e this mask is true where tokens match Padding Index, otherwise false)

Could you please help understand why this is a source of discrepancy? 

BTW I just compared the torchtext's output with fairseq's output at padded tokens, and it matches.

ebsmothers: @parmeet you can see the usage [here](https://github.com/ashkamath/mdetr/blob/main/models/transformer.py#L120-L121). Here `tokenized` contains `attention_mask`, which has shape `(batch_size, seq_len)` and tells us which tokens are padded. The problem is when we use TorchText's RoBERTaEncoder, we cannot modify this field at all. 

Starting from the bottom of the stack in `nn.MultiheadAttention`, you can see [here](https://github.com/pytorch/pytorch/blob/38e717dc876e0f9af66b32e9cc59b44b1f9edf27/torch/nn/modules/activation.py#L1024-L1032) that `attn_mask` there has shape `(target_seq_len, source_seq_len)`, which we do not want. Instead, we want `key_padding_mask`, which has shape `(batch_size, seq_len)`.

These two arguments are passed from `src_mask` and `src_key_padding_mask`, respectively, in `nn.TransformerEncoderLayer` ([ref](https://github.com/pytorch/pytorch/blob/master/torch/nn/modules/transformer.py#L460-L474)). But you can see that in TorchText's `TransformerEncoder` the value passed to `src_key_padding_mask` is constructed in the `forward()` ([ref](https://github.com/pytorch/text/blob/main/torchtext/models/roberta/modules.py#L139-L160)). In other words, we cannot actually pass the attention mask(padding mask? sorry, it's very hard to be consistent with naming here...) we get back from the tokenizer!

This is why I pointed out this line btw, as you can see in the [equivalent line](https://github.com/pytorch/text/blob/main/torchtext/models/roberta/modules.py#L160) in TorchText the value passed to `src_key_padding_mask` is not exposed in the API.

Hope this helps clarify things! Please let me know if anything is still unclear.

parmeet: Yes @ebsmothers, what you said is right. But why do we need to expose the padding mask? It is automatically constructed internally and is being passed to torch.nn.TransformerEncoderLayer. Sorry for noob question, but I am still not sure why this is source of discrepancy?  Assuming we expose the paddding mask in the API, what value would you be passing here to match the results from HF?

ebsmothers: @parmeet the padding mask seems to be used differently in the HF implementation than in PyTorch. I am not sure which is the "correct" way per se, but the implementation here (messy though it may be) is the only way I have been able to reproduce the results from the MDETR paper. There are two main differences between this version and the one in TorchText:

(1) inversion of the attention mask before passing it to TransformerEncoder
(2) my version doesn't apply the mask at the top level of the encoder 

Without both of these changes, I cannot match the results from MDETR. I think (1) is pretty standard, I know that HF and PyTorch have different mask definitions. But (2) is a bit more puzzling to me. Looking at the HF RoBERTa implementation, I do not see any use of their `attention_mask` outside of the self-attention layers. The `TransformerEncoder` layer in PyTorch core is similar. But in the TT `TransformerEncoder` for RoBERTa I see [this line](https://github.com/pytorch/text/blob/ed44dc88539b51785df03a276410cf71adfe6bfd/torchtext/models/roberta/modules.py#L309) applying the mask before passing to the encoder layers. Is this part of the standard RoBERTa implementation? If so, I am wondering why it's missing from the HF version.

parmeet: Thanks @ebsmothers for details. Regarding 2, It's a good question, I am not sure why HF version do not account for padding before passing to the encoder layers. I guess this explains the difference why the results mismatch (though only at padded tokens). The Fairseq version seems to be taking this into account [here](https://github.com/facebookresearch/fairseq/blob/08fe88479f8dd1f53705e203474a031d14cc2d75/fairseq/models/transformer/transformer_encoder.py#L211).

ebsmothers:
@ebsmothers has imported this pull request.  If you are a Facebook employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D37390041).

ebsmothers:
@ebsmothers has imported this pull request.  If you are a Facebook employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D37390041).

ebsmothers:
@ebsmothers has imported this pull request.  If you are a Facebook employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D37390041).

torchmultimodal/modules/encoders/mdetr_text_encoder.py
```diff
@@ -0,0 +1,284 @@
...
+        hidden_size=768,
+        vocab_size=50265,
+        pad_token_id=1,
+        type_vocab_size=1,
+        max_position_embeddings=514,
+        layer_norm_eps=1e-05,
+        hidden_dropout_prob=0.1,
+    )
+
+    wrapped_args = {
```
ankitade, line 273: why do u need this logic?cant u just write them yourself and expose them as function args

torchmultimodal/modules/encoders/mdetr_text_encoder.py
```diff
@@ -0,0 +1,284 @@
...
+        return encoded
+
+    def forward(
+        self,
+        embeddings: torch.Tensor,
+        attention_mask: Optional[torch.Tensor] = None,
+    ) -> Union[torch.Tensor, List[torch.Tensor]]:
+        out: Union[torch.Tensor, List[torch.Tensor]]
+        if self.return_all_layers:
+            out = self._forward_return_all_layers(embeddings, attention_mask)
```
ankitade, line 210: do u even need this? doesnt mdetr use last one only?

torchmultimodal/modules/encoders/mdetr_text_encoder.py
```diff
@@ -0,0 +1,284 @@
...
+
+    def _forward_return_last_layer(
+        self,
+        embeddings: torch.Tensor,
+        attention_mask: Optional[torch.Tensor] = None,
+    ) -> torch.Tensor:
+        encoded = embeddings
+        for layer in self.layers.layers:
+            encoded = layer(encoded, src_key_padding_mask=attention_mask)
+        if self.normalize_before:
```
ankitade, line 199: isnt this normalize after?

torchmultimodal/modules/encoders/mdetr_text_encoder.py
```diff
@@ -0,0 +1,284 @@
...
+    """ ""
+
+    def __init__(self, embeddings: nn.Module, encoder: nn.Module):
+        super().__init__()
+        self.embeddings = embeddings
+        self.encoder = encoder
+
+    def forward(
+        self,
+        input_ids: Optional[torch.Tensor] = None,
```
ankitade, line 243: not optional right?

torchmultimodal/modules/encoders/mdetr_text_encoder.py
```diff
@@ -0,0 +1,284 @@
...
+                (determines starting point of position IDs).
+    """
+    mask = input_ids.ne(padding_idx).int()
+    incremental_indices = torch.cumsum(mask, dim=1).type_as(mask) * mask
+    return incremental_indices.long() + padding_idx
+
+
+class MDETRTextEmbeddings(nn.Module):
+    """
+    Class for RoBERTa embeddings as used in MDETR. This is very similar to the FLAVA
```
ankitade, line 31: yeah idk this should be same as flava in theory (both roberta?). are you alluding to the handling of mask in position embeddings?

test/modules/encoders/test_mdetr_text_encoder.py
```diff
@@ -0,0 +1,198 @@
...
+        )
+        out = self.embeddings(self.input_ids)
+        actual = out[1, :, 1]
+        self.assertEqual(
+            out.size(), (self.batch_size, self.input_length, self.hidden_size)
+        )
+        assert_expected(actual, expected, rtol=0.0, atol=1e-4)
+
+    def test_mdetr_wrapped_transformer(self):
+        set_rng_seed(0)
```
ankitade, line 114: not needed

ankitade (commented):
in general, little confused by the whole wrapped args thing.

torchmultimodal/modules/encoders/mdetr_text_encoder.py
ebsmothers: Yeah it is more for consistency with TorchText. But lemme remove

torchmultimodal/modules/encoders/mdetr_text_encoder.py
ebsmothers: `normalize_before` return? 😄 . This is also how TorchText does it. Lmk if you want me to change it

torchmultimodal/modules/encoders/mdetr_text_encoder.py
ebsmothers: I thought FLAVA was using BERT embeddings? At least based on [this comment](https://github.com/facebookresearch/multimodal/blob/82480d8cf4a9359ecc28fbb4de257785bf25a1c3/torchmultimodal/models/flava/flava_text_encoder.py#L23)

torchmultimodal/modules/encoders/mdetr_text_encoder.py
ebsmothers: Not necessary, mainly just to make the dependence on TorchText RoBERTa more explicit. But can also remove these

torchmultimodal/modules/encoders/mdetr_text_encoder.py
```diff
@@ -0,0 +1,267 @@
...
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+from typing import List, Optional, Union
+
+import torch
+from torch import nn, Tensor
+
+
+def create_position_ids_from_input_ids(input_ids: Tensor, padding_idx: int):
```
ankitade, line 13: why global?

test/modules/encoders/test_mdetr_text_encoder.py
```diff
@@ -0,0 +1,198 @@
...
+                embeddings_state_dict[k] = v
+            # Set these to match the HF RoBERTa weights
+            elif k == "token_type_embeddings.weight":
+                embeddings_state_dict[k] = torch.zeros(v.size())
+            # Set all other weights randomly
+            else:
+                embeddings_state_dict[k] = torch.rand(v.size())
+        self.embeddings.load_state_dict(embeddings_state_dict)
+
+    def _populate_wrapped_transformer_weights(self):
```
ankitade, line 188: this test seems more complex that needed . why use the wrapped terminology and loading actual roberta weights?

torchmultimodal/modules/encoders/mdetr_text_encoder.py
```diff
@@ -0,0 +1,267 @@
...
+            activation="gelu",
+            batch_first=True,
+            norm_first=normalize_before,
+        )
+        self.layers = torch.nn.TransformerEncoder(
+            encoder_layer=layer, num_layers=num_encoder_layers
+        )
+        self.normalize_before = normalize_before
+        self.embedding_layer_norm = nn.LayerNorm(embedding_dim)
+        self.return_all_layers = return_all_layers
```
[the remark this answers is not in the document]
ebsmothers: No longer used after latest edits, lemme remove

torchmultimodal/modules/encoders/mdetr_text_encoder.py
```diff
@@ -0,0 +1,267 @@
...
+            torch.zeros(position_ids_range.size(), dtype=torch.long),
+            persistent=False,
+        )
+
+    def forward(
+        self,
+        input_ids: Tensor,
+        token_type_ids: Optional[Tensor] = None,
+        position_ids: Optional[Tensor] = None,
+        inputs_embeds: Optional[Tensor] = None,
```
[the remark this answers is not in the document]
ebsmothers: Yeah used in the HF version so kept it for flexibility. But for our implementation it's not needed

test/modules/encoders/test_mdetr_text_encoder.py
ebsmothers: Yeah the idea here was to explicitly check that the two pretrained checkpoints line up, since that is what had caused so much trouble previously. Agree it can be less complex though. Let me address this in a follow-up if that's good with you

torchmultimodal/modules/encoders/mdetr_text_encoder.py
ebsmothers: Thanks, lemme remove the reference to BERT. I still think the handling of position IDs is different though (at least when they are not passed, which is the case for us). In FLAVA they are just taking a [slice](https://github.com/facebookresearch/multimodal/blob/main/torchmultimodal/models/flava/flava_text_encoder.py#L73-L75), whereas here we explicitly use the `padding_idx`

torchmultimodal/modules/encoders/mdetr_text_encoder.py
```diff
@@ -0,0 +1,267 @@
...
+        self.embedding_dim = embedding_dim
+
+    def forward(
+        self,
+        embeddings: Tensor,
+        attention_mask: Optional[Tensor] = None,
+    ) -> Union[Tensor, List[Tensor]]:
+        encoded = embeddings
+        for layer in self.layers.layers:
+            encoded = layer(encoded, src_key_padding_mask=attention_mask)
```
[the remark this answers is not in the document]
ebsmothers: Update: this is causing memory problems when I remove the loop. Gonna dig into it more but for now just leaving the comment to explain things

ankitade (approved):
PTAL at the comment on the unit test

ebsmothers:
@ebsmothers has imported this pull request.  If you are a Facebook employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D37390041).

PR #64, opened 2022-06-03
[FLAVA] Tests for flava model

ankitade:
Summary:
- Adding some unit test for flava model
- Fixing small bug in the forward logic

Test plan:
pytest

### DIFF
--- a/test/models/test_flava.py
+++ b/test/models/test_flava.py
@@ -7,11 +7,16 @@
 import unittest
 
 import torch
+from test.test_utils import assert_expected
+from torch import nn
 from torchmultimodal.models.flava import (
+    flava_image_encoder,
     flava_model_for_classification,
     flava_model_for_pretraining,
+    flava_text_encoder,
+    FLAVAModel,
 )
-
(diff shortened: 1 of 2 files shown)


test/models/test_flava.py
```diff
@@ -126,3 +131,124 @@ def test_forward_pretraining(self):
...
+        text_to_mm_projection = nn.Identity()
+        self.flava = FLAVAModel(
+            image_encoder=self.image_encoder,
+            text_encoder=self.text_encoder,
+            mm_encoder=mm_encoder,
+            image_to_mm_projection=image_to_mm_projection,
+            text_to_mm_projection=text_to_mm_projection,
+        )
+
+    def _assert_empty(self, field):
```
RdoubleA, line 168: would it be useful for future testing to make this general for any of our model outputs? something like so:
'''
self.assertTrue(all(map(lambda x: x is None, field)))
'''
and then we could move to `test_utils.py`

ankitade: lets do it if / when we need it :)

PR #66, opened 2022-06-03
read training strategy from config

katrina433:
Summary:

* Read training strategy from config file
* Add `strategy: ddp` to all existing config files
* Can also pass in strategy from the terminal by `python train.py config=configs/pretraining/debug.yaml training.lightning.strategy=ddp`
* Priority: terminal argument > config

Test plan:

logged `self.trainer.strategy` in `FLAVAPreTrainingLightningModule`

test 1
* command: `python train.py config=configs/pretraining/debug.yaml`
* config: `strategy: ddp`
* result: `DDPStrategy`

test 2
* command: `python train.py config=configs/pretraining/debug.yaml training.lightning.strategy=deepspeed`
* config: `strategy: ddp`
* result: `DeepSpeedStrategy`

### DIFF
--- a/examples/flava/definitions.py
+++ b/examples/flava/definitions.py
@@ -66,6 +66,8 @@ class TrainingDatasetsInfo:
 class TrainingArguments:
     # Any lightning args to be pushed here
     lightning: Dict[str, Any] = field(default=dict)
+    lightning_checkpoint: Optional[Dict[str, Any]] = None
+    lightning_load_from_checkpoint: Optional[str] = None
     seed: int = -1
     batch_size: int = 8
     num_workers: int = 4
--- a/examples/flava/definitions.py
+++ b/examples/flava/definitions.py
@@ -74,6 +74,7 @@ class TrainingArguments:
     adam_weight_decay: float = 0.01
     adam_betas: Tuple[float, float] = field(default_factory=lambda: (0.9, 0.999))
     warmup_steps: int = 2000
+    strategy: str = "ddp"
 
 
 @dataclass
--- a/examples/flava/definitions.py
+++ b/examples/flava/definitions.py
@@ -74,7 +74,6 @@ class TrainingArguments:
     adam_weight_decay: float = 0.01
     adam_betas: Tuple[float, float] = field(default_factory=lambda: (0.9, 0.999))
     warmup_steps: int = 2000
-    strategy: str = "ddp"
 
 
 @dataclass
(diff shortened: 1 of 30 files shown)


ebsmothers (commented):
The changes look good to me. Since it's hard to add a unit test to test these changes, it would be good to provide more detail in the test plan though. Can you share commands you ran and the resulting outputs for a few different cases? (e.g. changing the default value or passing a command-line arg and demonstrating it results in the expected value in the trainer)

katrina433:
> The changes look good to me. Since it's hard to add a unit test to test these changes, it would be good to provide more detail in the test plan though. Can you share commands you ran and the resulting outputs for a few different cases? (e.g. changing the default value or passing a command-line arg and demonstrating it results in the expected value in the trainer)

updated description

examples/flava/train.py
```diff
@@ -59,7 +59,7 @@ def main():
             LearningRateMonitor(logging_interval="step"),
             MultimodalEvalCallback(imagenet_datamodule=imagenet_datamodule),
         ],
-        strategy="ddp",
+        strategy=config.training.strategy,
```
ankitade, line 62: i dont think u need this. line 57 should take care of this

examples/flava/definitions.py
```diff
@@ -74,6 +74,7 @@ class TrainingArguments:
     adam_weight_decay: float = 0.01
     adam_betas: Tuple[float, float] = field(default_factory=lambda: (0.9, 0.999))
     warmup_steps: int = 2000
+    strategy: str = "ddp"
```
ankitade, line 77: same as other comment. this should be specifiable through lightning config in line 68

ankitade (changes requested):
thanks for the changes. after u try out my suggestions, can u paste the logs from lightning in the test plan instead of relying on prints.

katrina433:
> thanks for the changes. after u try out my suggestions, can u paste the logs from lightning in the test plan instead of relying on prints.

Updated

examples/flava/utils.py
```diff
@@ -38,4 +38,8 @@ def build_config():
         "max_steps" in config.training.lightning
     ), "lightning config must specify 'max_steps'"
 
+    assert (
```
ankitade, line 41: do we need this? lightning works without any strategy override i thought

katrina433: removed

ankitade (approved):

PR #70, opened 2022-06-06
[MUGEN] Add multihead full/axial attention

RdoubleA:
Summary:
Introduces the `MultiHeadAttention` module which is used in MUGEN in the encoder and decoder for Video VQVAE and the GPT decoder. This module differs from PyTorch Core in that it allows for a flexible attention function by passing in the attention module you want to use to calculate attention instead of defaulting to full attention. For Video VQVAE, axial attention is needed to reduce cost of computing attention across three dimensions of the input. Axial Attention and Full Attention are introduced as separate modules to be used in `MultiHeadAttention` for this reason.

Test plan:
On a local notebook, compared the outputs from our implementation to the outputs from the MUGEN implementation and ensured they were equivalent. Then, used these as expected outputs in unit tests for each module and its methods in `test_attention.py`

`pytest -vv`
image

### DIFF
--- a/torchmultimodal/modules/layers/attention.py
+++ b/torchmultimodal/modules/layers/attention.py
@@ -26,7 +26,7 @@ class MultiHeadAttention(nn.Module):
         dim_q (int): dimensionality of query
         dim_kv (int): dimensionality of key/value
         n_head (int): number of attention heads
-        n_layer (int): ?
+        n_layer (int): number of attention layers being used in higher level stack
         causal (bool): use causal attention or not
         attn_module (nn.Module): module of attention mechanism to use
 
@@ -35,6 +35,8 @@ class MultiHeadAttention(nn.Module):
                           a [b, 1, ..., 1, c] tensor if decode_step is not None
 
     """
+
+    # TODO: remove dependency on n_layer, higher level detail should not be a parameter
 
     def __init__(
         self,
--- a/torchmultimodal/modules/layers/attention.py
+++ b/torchmultimodal/modules/layers/attention.py
@@ -1,0 +1,182 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+from typing import Dict, Tuple
+
+import torch
+from torch import nn, Tensor
+from torch.nn import functional as F
+from torchmultimodal.utils.common import shift_dim, view_range
+
+
+class MultiHeadAttention(nn.Module):
+    """Compute multihead attention with flexible attention function
+
+    Args:
+        shape (Tuple): ?
+        dim_q (int): dimensionality of input into query weights
+        dim_kv (int): dimensionality of input into key and value weights
+        n_head (int): number of attention heads
+        n_layer (int): ?
+        causal (bool): use causal attention or not
+        attn_module (nn.Module): module of attention function to use
+
+    Inputs:
+        q, k, v (Tensor): a [b, d1, ..., dn, c] tensor or
+                          a [b, 1, ..., 1, c] tensor if decode_step is not None
+
+    """
+
+    def __init__(
+        self,
+        shape: Tuple,
+        dim_q: int,
+        dim_kv: int,
+        n_head: int,
+        n_layer: int,
+        causal: bool,
+        attn_module: nn.Module,
+    ):
+        super().__init__()
+        self.causal = causal
+        self.shape = shape
+
+        self.d_k = dim_q // n_head
+        self.d_v = dim_kv // n_head
+        self.n_head = n_head
+        self.w_qs = nn.Linear(dim_q, n_head * self.d_k, bias=False)  # q
+        self.w_qs.weight.data.normal_(std=1.0 / torch.sqrt(torch.tensor(dim_q)))
+
+        self.w_ks = nn.Linear(dim_kv, n_head * self.d_k, bias=False)  # k
+        self.w_ks.weight.data.normal_(std=1.0 / torch.sqrt(torch.tensor(dim_kv)))
+
+        self.w_vs = nn.Linear(dim_kv, n_head * self.d_v, bias=False)  # v
+        self.w_vs.weight.data.normal_(std=1.0 / torch.sqrt(torch.tensor(dim_kv)))
+
+        self.fc = nn.Linear(n_head * self.d_v, dim_q, bias=True)  # c
+        self.fc.weight.data.normal_(std=1.0 / torch.sqrt(torch.tensor(dim_q * n_layer)))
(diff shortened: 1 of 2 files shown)


torchmultimodal/modules/layers/attention.py
```diff
@@ -0,0 +1,240 @@
...
+import torch
+from torch import nn, Tensor
+from torch.nn import functional as F
+from torchmultimodal.utils.common import shift_dim
+
+
+class MultiHeadAttention(nn.Module):
+    """Compute multihead attention with flexible attention function.
+
+    Multihead attention linearly projects queries, keys, and values into an embedding
```
langong347, line 18: `embedding space` sounds very generic. Would instead focus on the purpose of the linear projection which is to run the scaled product attention multiple times in parallel.

torchmultimodal/modules/layers/attention.py
```diff
@@ -0,0 +1,240 @@
...
+    space, which is divided into multiple 'heads'. Attention is computed for each head
+    individually instead of across the entire embedding space once. This enables more
+    varied representations and allows the model to jointly attend to information from
+    different representation subspaces at different positions, as described in
+    Attention Is All You Need (Vaswani et al. 2017).
+
+    Args:
+        shape (Tuple): shape of input data (d1 x ... x dn)
+        dim_q (int): dimensionality of input into query weights
```
langong347, line 27: "dimension of the query" is enough? The same for key and value below.

torchmultimodal/modules/layers/attention.py
```diff
@@ -0,0 +1,240 @@
...
+        dim_kv (int): dimensionality of input into key and value weights
+        n_head (int): number of attention heads
+        n_layer (int): ?
```
langong347, line 30: `n_layer` is the number of attention blocks specified by the GPT transformer. I'm not sure why it's used by the feed forward layer for a single attention block. Could you post a question to Songyang?

torchmultimodal/modules/layers/attention.py
```diff
@@ -0,0 +1,240 @@
...
+        causal (bool): use causal attention or not
+        attn_module (nn.Module): module of attention function to use
```
langong347, line 32: "attention function" --> "attention mechanism"?

langong347 (commented):
Just to confirm: should I add the tests for decoding?

torchmultimodal/modules/layers/attention.py
```diff
@@ -0,0 +1,240 @@
...
+
+    Inputs:
+        q, k, v (Tensor): a [b, h, d1, ..., dn, c] tensor or
+                          a [b, h, 1, ..., 1, c] tensor if decode_step is not None
+
+    """
+
+    def __init__(self, axial_dim: int) -> None:
+        super().__init__()
+        self.axial_dim = axial_dim + 2  # account for batch, head
```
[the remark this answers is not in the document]
RdoubleA: imo it's confusing because first case ignores batch, head, hidden dim but the negative case does not, at least in MUGEN code. I didn't think it would be necessary to support this. Supporting negative indexing where it ignores the other dims might be nice, I could add that.

torchmultimodal/modules/layers/attention.py
```diff
@@ -0,0 +1,240 @@
...
+            k, v = self.cache["k"], self.cache["v"]
+
+        a = self.attn(q, k, v, decode_step, decode_idx)
+        a = self._combine_multihead(a)
+        a = self.fc(a)
+
+        return a
+
+
+class FullAttention(nn.Module):
```
ankitade, line 130: is the above MHA + Full Attention the same as the MHA from core? if yes, why not use that for gpt and have an implementation for axial attention?

RdoubleA: that would still require our own MHA class. was also thinking we could punt consolidating with core later

RdoubleA: resolved offline

langong347 (approved):
Awesome, thanks! Now the ball is in my court.

torchmultimodal/modules/layers/attention.py
```diff
@@ -0,0 +1,241 @@
...
+        # fast decoding
+        if decode_step is not None:
+            if decode_step == 0:
+                if self.causal:
+                    k_shape = (
+                        q.shape[0],
+                        n_head,
+                        *self.shape,
+                        self.d_k,
+                    )  # B x n_head x 3 x 32 x 32 x d
```
ebsmothers, line 105: Is this comment correct?

ebsmothers (commented):
Sorry, meant to get these comments in earlier today but seems they never went through. Since it's already accepted, please don't consider them blocking comments 🙂

torchmultimodal/modules/layers/attention.py
RdoubleA: need to remove this

torchmultimodal/modules/layers/attention.py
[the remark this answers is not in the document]
RdoubleA: if we want to specify the second axis of a 3d input, we should be able to use 1 or -2. If the input tensor is `b,h,d1,d2,d3,c`, then 1 += 2 maps to 3 which is correct. 2 + (3 + 1 + -2) maps to 4 which is not correct... the user would have to use -3 to account for channel dimension, but in the positive case 1 does not have to account for any extra dimension.

torchmultimodal/modules/layers/attention.py
```diff
@@ -0,0 +1,241 @@
...
+        old_shape = list(v.shape)
+        v = v.flatten(end_dim=-3)
+
+        out = scaled_dot_product_attention(q, k, v, training=self.training)
+        out = out.view(*old_shape)
+        out = shift_dim(out, -2, self.axial_dim)
+        return out
+
+
+def scaled_dot_product_attention(
```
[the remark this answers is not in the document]
RdoubleA: that's an interesting idea, actually pytorch core handles setting dropout to 0 if not training one level higher so I'll follow that

torchmultimodal/modules/layers/attention.py
```diff
@@ -0,0 +1,241 @@
...
+        causal (bool): use causal attention or not
+        attn_module (nn.Module): module of attention mechanism to use
+
+    Inputs:
+        q, k, v (Tensor): a [b, d1, ..., dn, c] tensor or
+                          a [b, 1, ..., 1, c] tensor if decode_step is not None
+
+    """
+
+    # TODO: remove dependency on n_layer, higher level detail should not be a parameter
```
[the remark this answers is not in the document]
RdoubleA: yeah I could remove n_layer and have a higher level module specify the initialization of the FC layer. PyTorch core seems to initialize this differently

PR #74, opened 2022-06-07
Added tests for CLIPTextEncoder

sophiazhi:
Summary:
Added pytest unit tests for CLIPTextEncoder in `test/modules/encoders/test_clip_text_encoder.py` to reflect the new modular design for CLIP. Edited/strengthened typing in CLIPTextEncoder to satisfy the new scripting test.

Test plan:
Tested by running the command `pytest --cov=torchmultimodal/modules/encoders/ test/modules/encoders/test_clip_text_encoder.py::TestCLIPTextEncoder -vv` from the project root. All test cases passed and coverage report stated 97% coverage of `clip_text_encoder.py`. (One type-checking line is not covered because it's impossible to hit with the current code.)
Screen Shot 2022-06-09 at 4 35 36 PM

### DIFF
--- a/test/modules/encoders/test_clip_text_encoder.py
+++ b/test/modules/encoders/test_clip_text_encoder.py
@@ -1,0 +1,161 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+import pytest
+import torch
+from test.test_utils import assert_expected, set_rng_seed
+from torchmultimodal.modules.encoders.clip_text_encoder import CLIPTextEncoder
+
+
+class TestCLIPTextEncoder:
+    @pytest.fixture
+    def start(self):
+        set_rng_seed(1234)
+        context_length = 77
+        batch_size, embedding_dim = 2, 4
+
+        def build_text(text_length):
+            return torch.randint(1, 10, (batch_size, text_length), dtype=torch.long)
+
+        def build_encoder(
+            embedding_dim=embedding_dim,
+            use_clip_init=True,
+            context_length=context_length,
+        ):
+            return CLIPTextEncoder(
+                embedding_dim=embedding_dim,
+                use_clip_init=use_clip_init,
+                context_length=context_length,
+            )
+
+        return build_encoder, build_text
+
+    def test_initialization(self, start):
+        """Test whether the ``use_clip_init`` parameter changes the initialized parameters."""
+        build_encoder, _ = start
+
+        # First, initialize without CLIP parameters and get the standard deviation of token embedding weights
+        # Use larger embedding size for stability in std
+        text_encoder = build_encoder(use_clip_init=False, embedding_dim=50)
+        assert isinstance(text_encoder, torch.nn.Module)
+        actual_std_no_init = torch.std(
+            text_encoder.encoder.token_embedding.weight
+        ).item()
+
+        # Get the standard deviation with CLIP params
+        text_encoder = build_encoder(embedding_dim=50)
+        actual_std_clip_init = torch.std(
+            text_encoder.encoder.token_embedding.weight
+        ).item()
+
+        assert abs(actual_std_clip_init - actual_std_no_init) > 1e-1
+
+    def test_clip_parameters(self, start):
+        build_encoder, _ = start
+        # Use larger embedding size for stability in std
+        text_encoder = build_encoder(embedding_dim=50)
+
+        assert_expected(
+            actual=torch.std(text_encoder.encoder.token_embedding.weight).item(),
+            expected=0.02,
+            rtol=0,
+            atol=1e-4,
+        )
+        assert_expected(
+            actual=torch.std(
+                text_encoder.encoder.positional_embedding.embedding.weight
+            ).item(),
+            expected=0.01,
+            rtol=0,
+            atol=1e-4,
+        )
+
+        proj_std = 0.0090
+        attn_std = 0.0442
+        fc_std = 0.0313
+        for layer in text_encoder.encoder.layers.layers:
+            assert_expected(
+                actual=torch.std(layer.self_attn.in_proj_weight).item(),
+                expected=attn_std,
+                rtol=0,
+                atol=1e-4,
+            )
+            assert_expected(
+                actual=torch.std(layer.self_attn.out_proj.weight).item(),
+                expected=proj_std,
+                rtol=0,
+                atol=1e-4,
+            )
+            assert_expected(
+                actual=torch.std(layer.linear1.weight).item(),
+                expected=fc_std,
+                rtol=0,
+                atol=1e-4,
+            )
+            assert_expected(
+                actual=torch.std(layer.linear2.weight).item(),
+                expected=proj_std,
+                rtol=0,
+                atol=1e-4,
+            )
+
+        assert_expected(
+            actual=torch.std(text_encoder.projection.weight).item(),
+            expected=0.0442,
+            rtol=0,
+            atol=1e-4,
+        )
+
+    def test_attention_mask(self, start):
+        build_encoder, _ = start
+        text_encoder = build_encoder(context_length=4)
+        assert isinstance(text_encoder, torch.nn.Module)
+
+        actual = text_encoder.build_attention_mask()
+        expected = torch.Tensor(
+            [[0, 1, 1, 1], [0, 0, 1, 1], [0, 0, 0, 1], [0, 0, 0, 0]]
+        ).to(dtype=bool)
+        assert_expected(actual=actual, expected=expected, rtol=0, atol=0)
+
+    def test_forward(self, start):
+        build_encoder, build_text = start
+        text = build_text(text_length=3)
+
+        text_encoder = build_encoder()
+        assert isinstance(text_encoder, torch.nn.Module)
+
+        actual_clip_init = text_encoder(text)
+        expected_clip_init = torch.Tensor(
+            [[-1.9412, -0.8044, -0.7616, -0.3185], [0.2408, -0.3241, -0.2544, -0.3608]]
+        )
+        assert_expected(
+            actual=actual_clip_init, expected=expected_clip_init, rtol=0, atol=1e-4
+        )
+
+    def test_forward_over_context_length(self, start):
+        build_encoder, build_text = start
+
+        text_encoder = build_encoder()
+        assert isinstance(text_encoder, torch.nn.Module)
+
+        text = build_text(text_encoder.context_length + 1)
+
+        with pytest.raises(IndexError):
+            text_encoder(text)
+
+    def test_scripting(self, start):
+        build_encoder, build_text = start
+        text = build_text(text_length=3)
+
+        text_encoder = build_encoder()
+        assert isinstance(text_encoder, torch.nn.Module)
+        scripted_encoder = torch.jit.script(text_encoder)
+
+        actual = scripted_encoder(text)
+        expected = torch.Tensor(
+            [[-1.9412, -0.8044, -0.7616, -0.3185], [0.2408, -0.3241, -0.2544, -0.3608]]
+        )
+        assert_expected(actual=actual, expected=expected, rtol=0, atol=1e-4)
--- a/torchmultimodal/modules/encoders/clip_text_encoder.py
+++ b/torchmultimodal/modules/encoders/clip_text_encoder.py
@@ -86,11 +86,15 @@ class CLIPTextEncoder(nn.Module):
         nn.init.normal_(self.projection.weight, std=self.width**-0.5)
 
     def build_attention_mask(self) -> torch.Tensor:
-        mask = torch.full((self.context_length, self.context_length), True).triu(1)
+        # To support torchscripting, we have to pass an int as fill_value
+        mask = torch.full((self.context_length, self.context_length), int(True)).triu(1)
         return mask.to(device=None, dtype=torch.bool)
 
     def forward(self, text: torch.Tensor) -> torch.Tensor:
         embeddings = self.encoder(text, attn_mask=self.build_attention_mask())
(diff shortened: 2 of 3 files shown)


test/modules/encoders/test_clip_text_encoder.py
```diff
@@ -0,0 +1,93 @@
...
+from test.test_utils import assert_expected
+from torchmultimodal.modules.encoders.clip_text_encoder import CLIPTextEncoder
+from torchmultimodal.utils.common import get_current_device
+
+torch.manual_seed(1234)
+context_length = 77
+device = get_current_device()
+
+
+class TestCLIPTextEncoder:
```
ebsmothers, line 14: Can you inherit from `unittest.TestCase`?

test/modules/encoders/test_clip_text_encoder.py
```diff
@@ -0,0 +1,93 @@
...
+        embedding_dim = 256
+        batch_size = 16
+        text = torch.randint(1, 79, (batch_size, context_length + 5), dtype=torch.long)
+
+        text_encoder = CLIPTextEncoder(embedding_dim=embedding_dim, use_clip_init=True)
+        text_encoder = text_encoder.to(device)
+        assert isinstance(text_encoder, torch.nn.Module)
+
+        with pytest.raises(IndexError):
+            out = text_encoder(text)
```
ebsmothers, line 101: You can also add a test that the module can be torchscripted (I think it should be able to be scripted, if not let me know). [Example](https://github.com/facebookresearch/multimodal/blob/main/test/modules/encoders/test_mil_encoder.py#L76)

test/modules/encoders/test_clip_text_encoder.py
```diff
@@ -0,0 +1,93 @@
...
+    def test_over_context_length(self):
+        embedding_dim = 256
+        batch_size = 16
+        text = torch.randint(1, 79, (batch_size, context_length + 5), dtype=torch.long)
+
+        text_encoder = CLIPTextEncoder(embedding_dim=embedding_dim, use_clip_init=True)
+        text_encoder = text_encoder.to(device)
+        assert isinstance(text_encoder, torch.nn.Module)
+
+        with pytest.raises(IndexError):
```
ebsmothers, line 100: With `unittest` you can do `with self.assertRaises` here

test/modules/encoders/test_clip_text_encoder.py
```diff
@@ -0,0 +1,93 @@
...
+from torchmultimodal.modules.encoders.clip_text_encoder import CLIPTextEncoder
+from torchmultimodal.utils.common import get_current_device
+
+torch.manual_seed(1234)
+context_length = 77
+device = get_current_device()
+
+
+class TestCLIPTextEncoder:
+    def test_clip_init(self):
```
ebsmothers: I'm not sure that this test is really testing the CLIP initialization explicitly. Maybe you can add checks on some of the parameters that get set in `initialize_parameters`.

test/modules/encoders/test_clip_text_encoder.py
```diff
@@ -0,0 +1,93 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+import pytest
```
ebsmothers, line 7: You can use `unittest` instead as in our other tests

test/modules/encoders/test_clip_text_encoder.py
```diff
@@ -0,0 +1,93 @@
...
+        batch_size = 16
+        text = torch.randint(1, 79, (batch_size, context_length - 5), dtype=torch.long)
+
+        text_encoder = CLIPTextEncoder(embedding_dim=embedding_dim, use_clip_init=True)
+        text_encoder = text_encoder.to(device)
+        assert isinstance(text_encoder, torch.nn.Module)
+
+        out = text_encoder(text)
+        assert out.shape == (batch_size, embedding_dim)
+        assert_expected(out.sum().item(), -142.1651, rtol=0, atol=1e-3)
```
ebsmothers: Similar comment to `test_clip_init`, this isn't really testing the fact that the input text is shorter than the context length. You could check that padding is occurring if you want to explicitly test this.

test/modules/encoders/test_clip_text_encoder.py
sophiazhi: Added a scripting test but also modified CLIPTextEncoder to pass the test -- see recent changes

test/modules/encoders/test_clip_text_encoder.py
sophiazhi: added a check for some layers that the mean of the weights is 0

torchmultimodal/modules/encoders/clip_text_encoder.py
```diff
@@ -86,12 +86,13 @@ def initialize_parameters(self) -> None:
         nn.init.normal_(self.projection.weight, std=self.width**-0.5)
 
     def build_attention_mask(self) -> torch.Tensor:
-        mask = torch.full((self.context_length, self.context_length), True).triu(1)
+        mask = torch.full((self.context_length, self.context_length), int(True)).triu(1)
```
ebsmothers, line 89: Thanks for making the changes to support making CLIPTextEncoder scriptable!

test/modules/encoders/test_clip_text_encoder.py
```diff
@@ -0,0 +1,121 @@
...
+        context_length = 77
+        return context_length
+
+    def test_clip_init(self, setUp):
+        context_length = setUp
+        batch_size, embedding_dim = 4, 8
+        text_length = min(context_length, 3)
+        text = torch.randint(1, 10, (batch_size, text_length), dtype=torch.long)
+
+        text_encoder = CLIPTextEncoder(embedding_dim=embedding_dim, use_clip_init=True)
```
ebsmothers, line 26: A nice thing to do here to confirm that `use_clip_init` is actually doing something (as opposed to e.g. parameters being initialized as all zeros) would be to set `use_clip_init=False` and assert that the mean is nonzero, then call `initialize_parameters` directly and run the zero-mean assertions

ebsmothers (changes requested):
Overall it looks good! Left a couple more small comments but that should be it. And thanks for making the changes in the text encoder to support scripting!

test/modules/encoders/test_clip_text_encoder.py
```diff
@@ -0,0 +1,121 @@
...
+from test.test_utils import assert_expected, set_rng_seed
+from torchmultimodal.modules.encoders.clip_text_encoder import CLIPTextEncoder
+
+
+class TestCLIPTextEncoder:
+    @pytest.fixture
+    def setUp(self):
+        set_rng_seed(1234)
+        context_length = 77
+        return context_length
```
langong347, line 18: How about using `pytest.fixture` to initialize `CLIPTextEncoder` as it's needed by every test? Although you'll need to find a way to pipe in runtime arguments such as `use_clip_init`. Return a function is just one way to do so:

https://github.com/facebookresearch/multimodal/blob/5cbf74edfba952005439cef3ac25eb10882c9302/test/modules/layers/test_attention.py#L13-L22

torchmultimodal/modules/encoders/clip_text_encoder.py
```diff
@@ -86,12 +86,13 @@ def initialize_parameters(self) -> None:
...
         return mask.to(device=None, dtype=torch.bool)
 
     def forward(self, text: torch.Tensor) -> torch.Tensor:
         embeddings = self.encoder(text, attn_mask=self.build_attention_mask())
         # [n_ctx, bs, transformer.width] -> [bs, n_ctx, transformer.width]
+        assert isinstance(embeddings, torch.Tensor)
```
langong347, line 95: Why do we need the assertion here? Also, it's better to `if/raise` a specific error instead of a general assertion error.

test/modules/encoders/test_clip_text_encoder.py
```diff
@@ -0,0 +1,121 @@
...
+        text = torch.randint(1, 10, (batch_size, text_length), dtype=torch.long)
+
+        text_encoder = CLIPTextEncoder(embedding_dim=embedding_dim, use_clip_init=False)
+        assert isinstance(text_encoder, torch.nn.Module)
+
+        out = text_encoder(text)
+        assert out.shape == (batch_size, embedding_dim)
+        assert_expected(out.sum().item(), -2.3848, rtol=0, atol=1e-3)
+
+    def test_small_embedding(self, setUp):
```
langong347, line 60: Does the module behave differently for small embedding dim?

torchmultimodal/modules/encoders/clip_text_encoder.py
sophiazhi: The assertion is needed by torchscript because aten::permute (on the next line) only takes tensors as input, but the encoder output is annotated as either a tensor or a list of tensors. I'll change the type check to the if/raise format.

test/modules/encoders/test_clip_text_encoder.py
```diff
@@ -0,0 +1,111 @@
...
+        def build_encoder(embedding_dim=embedding_dim, use_clip_init=True):
+            return CLIPTextEncoder(
+                embedding_dim=embedding_dim, use_clip_init=use_clip_init
+            )
+
+        return build_encoder, build_text, context_length
+
+    def test_initialization(self, start):
+        build_encoder, build_text, context_length = start
+        text_length = min(context_length, 3)
```
ebsmothers, line 32: The min is redundant here, I think it's clearer to just write `text_length=3`

ebsmothers (commented):
Other than the two minor comments, this looks good to go.

test/modules/encoders/test_clip_text_encoder.py
```diff
@@ -12,110 +12,100 @@
...
+            expected=expected_std_clip_init,
+            rtol=0,
+            atol=1e-4,
+        )
 
-    def test_at_context_length(self, setUp):
-        context_length = setUp
-        batch_size, embedding_dim = 4, 8
-        text = torch.randint(1, 10, (batch_size, context_length), dtype=torch.long)
+    def test_at_context_length(self, start):
```
langong347, line 75: Is there any significant difference due to `text_length` and `context_length`? If the purpose is just to make sure the code runs when `text_length <= context_length`, shall we just have one test (e.g., `test_forward`) since you already have a test for the edge case `over_context_length`?

torchmultimodal/modules/encoders/clip_text_encoder.py
```diff
@@ -91,8 +91,9 @@ def build_attention_mask(self) -> torch.Tensor:
 
     def forward(self, text: torch.Tensor) -> torch.Tensor:
         embeddings = self.encoder(text, attn_mask=self.build_attention_mask())
+        if not isinstance(embeddings, torch.Tensor):
```
langong347, line 94: Why is this needed?

sophiazhi: aten::permute (used by torchscript on the next line) only takes tensors as input, but the encoder output is annotated as either a tensor or a list of tensors

test/modules/encoders/test_clip_text_encoder.py
```diff
@@ -12,110 +12,100 @@
...
+            [[-1.9412, -0.8044, -0.7616, -0.3185], [0.2408, -0.3241, -0.2544, -0.3608]]
+        )
+        assert_expected(
+            actual=actual_clip_init, expected=expected_clip_init, rtol=0, atol=1e-4
+        )
 
-        out = text_encoder(text)
-        assert out.shape == (1, embedding_dim)
-        assert_expected(out.sum().item(), -1.4762, rtol=0, atol=1e-3)
+        actual_std_clip_init = torch.std(
```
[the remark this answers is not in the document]
sophiazhi: Looked at the source code for nn.init.normal_, and it doesn't look like there's a direct way to get the distribution type. Manual methods that I saw also look handwavy :(

test/modules/encoders/test_clip_text_encoder.py
```diff
@@ -20,85 +20,136 @@ def start(self):
...
+        # Get the standard deviation with CLIP params
         text_encoder.initialize_parameters()
+        actual_std_clip_init = torch.std(
+            text_encoder.encoder.token_embedding.weight
+        ).item()
 
-        actual_clip_init = text_encoder(text)
-        expected_clip_init = torch.Tensor(
-            [[-1.9412, -0.8044, -0.7616, -0.3185], [0.2408, -0.3241, -0.2544, -0.3608]]
+        assert abs(actual_std_clip_init - actual_std_no_init) > 1e-4
```
langong347, line 53: - Since we are optimizing `actual_std_clip_init` being different from `actual_std_no_init`,  could we try to use a larger embedding matrix (so you have more samples) for `TransformerEncoder` such that the actual `std`s become more stable (also see [law of large numbers](https://en.wikipedia.org/wiki/Law_of_large_numbers))?

- What is the maximum delta we can go up here?  Since the preset values are in the order of [`1e-2`](https://github.com/facebookresearch/multimodal/blob/main/torchmultimodal/modules/encoders/clip_text_encoder.py#L30-L31), I would expect a reasonable delta to be at least `1e-2`. Anything below that can be prone to statistical noise.

test/modules/encoders/test_clip_text_encoder.py
```diff
@@ -20,85 +20,136 @@ def start(self):
...
+            ).item(),
+            expected=0.01,
+            rtol=0,
+            atol=1e-4,
         )
 
-        actual_std_clip_init = torch.std(
-            text_encoder.encoder.token_embedding.weight
-        ).item()
+        proj_std = (text_encoder.width**-0.5) * (
```
langong347, line 75: Compute ground truths independently to avoid copying the logic from the code you are testing.

langong347 (commented):
Please address the comments but overall big improvements from previous versions.

langong347 (approved):

PR #79, opened 2022-06-10
[examples] Extract MultiData Module/Loader to common

IvanKobzarev:
* #92
* __->__ #79


ankitade (commented):
- can u add some unit tests to this PR
- why are the imports changing everywhere? are those related changes
- multitask.py might be a better name than multidata

IvanKobzarev:
> * can u add some unit tests to this PR
Yep, working on it, its in progress - added test/data/test_iteration_strategies.py with one test and will be covering it more

> * why are the imports changing everywhere? are those related changes
That was at the stage of placing to `examples/common/` to have `examples` as a parent package for `flava` and `common` to be able to use `common` from `flava` (it can be done only if to be in the same package) - in that case it was usable  as `cd examples && python -m flava.train config=...`

But that does not work to add test.
So placing in `torchmultimodal/data` is better place and tests can reuse it and it is more separated from examples functionality.

> * multitask.py might be a better name than multidata
Hm. As it contains MultiDataLoader and MultiDataModule - for me multidata was closer to the common part. Multitask in my understanding when we have multiple different heads for different tasks, while here it is more controlling data loading that's why I choosed `multidata`

ankitade:
this got merged into another PR right. can we close this one @IvanKobzarev

PR #80, opened 2022-06-10
[ALBEF] First Draft of the ALBEF Model

katrina433:
Summary:

This is a draft of the ALBEF model, based on the [pre-training](https://github.com/salesforce/ALBEF/blob/6224e78e5292757ca9c7d5add3e415010955a58b/models/model_pretrain.py) logic in the ALBEF repo. 

Here's the [paper](https://arxiv.org/pdf/2107.07651.pdf). 

In addition to the 3 embeddings (image, text, multimodal), `forward` also returns similarity (for `loss_ita`), momentum image embeds  and text attention mask (for `loss_mlm`). 

The original plan was to create an ALBEF architecture, but there are specific computations that need to happen in order to compute the multimodal embedding (momentum updates and similarity calculation), so we thought this serves more as a model rather than an architecture. 

TODO: inside `_update_queue` there is an assertion for simplicity purpose (to match the original repo and and make sure that the outputs match). But eventually want to remove the assertion (which might introduce some randomness when computing similarity with momentum embeddings)

Test plan:

I am working on writing tests, but just wanted to put out a draft to get some initial feedback.

### DIFF
--- a/torchmultimodal/models/albef.py
+++ b/torchmultimodal/models/albef.py
@@ -11,6 +11,10 @@ import torch
 import torch.nn.functional as F
 from torch import nn
 
+from torchmultimodal.modules.losses.contrastive_loss_with_temperature import (
+    _gather_embeddings_and_labels,
+)
+
 
 ALBEFOutput = namedtuple(
     "ALBEFOutput",
@@ -32,7 +36,7 @@ ALBEFSimilarity = namedtuple(
 )
 
 
-class ALBEFArchitecture(nn.Module):
+class ALBEFModel(nn.Module):
     def __init__(
         self,
         vision_encoder: nn.Module,
@@ -41,7 +45,7 @@ class ALBEFArchitecture(nn.Module):
         vision_proj: nn.Module,
         text_proj: nn.Module,
         embed_dim: int = 256,
-        queue_size: int = 65534,
+        queue_size: int = 65536,
         temp: float = 0.07,
         momentum: float = 0.995,
     ):
@@ -149,22 +153,10 @@ class ALBEFArchitecture(nn.Module):
                 )
 
     @torch.no_grad()
-    def _concat_all_gather(self, tensor):
-        if torch.is_distributed(tensor):
-            tensors_gather = [
-                torch.ones_like(tensor)
-                for _ in range(torch.distributed.get_world_size())
-            ]
-            torch.disstributed.all_gather(tensors_gather, tensor, async_op=False)
-            output = torch.cat(tensors_gather, dim=0)
-            return output
-        else:
-            return tensor
-
-    @torch.no_grad()
     def _update_queue(self, image_feat_m, text_feat_m):
-        image_feats = self._concat_all_gather(image_feat_m)
-        text_feats = self._concat_all_gather(text_feat_m)
+        image_feats, text_feats, _ = _gather_embeddings_and_labels(
+            image_feat_m, text_feat_m
+        )
         batch_size = image_feats.shape[0]
         ptr = int(self.queue_ptr)
         assert self.queue_size % batch_size == 0
--- a/torchmultimodal/models/albef.py
+++ b/torchmultimodal/models/albef.py
@@ -61,7 +61,22 @@ class ALBEFModel(nn.Module):
         self.vision_proj_m = copy.deepcopy(vision_proj)
         self.text_proj_m = copy.deepcopy(text_proj)
 
-        self._copy_params()
+        self.models = [
+            self.vision_encoder,
+            self.text_encoder,
+            self.multimodal_encoder,
+            self.vision_proj,
+            self.text_proj,
+        ]
+        self.models_m = [
+            self.vision_encoder_m,
+            self.text_encoder_m,
+            self.multimodal_encoder_m,
+            self.vision_proj_m,
+            self.text_proj_m,
+        ]
+
+        self._copy_params_momentum_models()
 
         self.queue_size = queue_size
         self.temp = temp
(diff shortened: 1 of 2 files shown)


torchmultimodal/models/albef.py
```diff
@@ -0,0 +1,244 @@
...
+    ],
+    defaults=(None, None, None, None, None, None),
+)
+
+ALBEFSimilarity = namedtuple(
+    "ALBEFSimilarity",
+    ["sim_i2t", "sim_t2i", "sim_i2t_targets", "sim_t2i_targets"],
+    defaults=(None, None, None, None),
+)
+
```
RdoubleA, line 37: some comments explaining what each of these fields represent would be great

torchmultimodal/models/albef.py
```diff
@@ -0,0 +1,244 @@
...
+        self.register_buffer("text_queue", torch.randn(embed_dim, queue_size))
+        self.register_buffer("queue_ptr", torch.zeros(1, dtype=torch.long))
+        self.image_queue = nn.functional.normalize(self.image_queue, dim=0)
+        self.text_queue = nn.functional.normalize(self.text_queue, dim=0)
+
+    def forward(
+        self,
+        image: torch.Tensor,
+        text: torch.Tensor,
+        text_atts: torch.Tensor = None,
```
RdoubleA, line 80: Use `Optional[Tensor]`

torchmultimodal/models/albef.py
```diff
@@ -0,0 +1,244 @@
...
+    def _update_queue(self, image_feat_m, text_feat_m):
+        image_feats, text_feats, _ = _gather_embeddings_and_labels(
+            image_feat_m, text_feat_m
+        )
+        batch_size = image_feats.shape[0]
+        ptr = int(self.queue_ptr)
+        assert self.queue_size % batch_size == 0
+
+        self.image_queue[:, ptr : ptr + batch_size] = image_feats.T
+        self.text_queue[:, ptr : ptr + batch_size] = text_feats.T
```
RdoubleA, line 165: I liked the comment in the original repo that this is a dequeue and enqueue process, it makes it more clear what these lines of code are doing. perhaps you could add your own comments to explain these lines, or at the top of the function to explain what this is doing

torchmultimodal/models/albef.py
```diff
@@ -0,0 +1,244 @@
...
+        self._copy_params()
+
+        self.queue_size = queue_size
+        self.temp = temp
+        self.momentum = momentum
+
+        self.register_buffer("image_queue", torch.randn(embed_dim, queue_size))
+        self.register_buffer("text_queue", torch.randn(embed_dim, queue_size))
+        self.register_buffer("queue_ptr", torch.zeros(1, dtype=torch.long))
+        self.image_queue = nn.functional.normalize(self.image_queue, dim=0)
```
RdoubleA, line 73: I know the paper uses queues "inspired by MoCo", but could you add some comments explaining why they are needed? also some comments explaining the other class variables above would be great

torchmultimodal/models/albef.py
```diff
@@ -0,0 +1,244 @@
...
+            sim_i2t_targets = (
+                alpha * F.softmax(sim_i2t_m, dim=1) + (1 - alpha) * sim_targets
+            )
+            sim_t2i_targets = (
+                alpha * F.softmax(sim_t2i_m, dim=1) + (1 - alpha) * sim_targets
+            )
+
+        sim_i2t = image_feat @ text_feat_all / self.temp
+        sim_t2i = text_feat @ image_feat_all / self.temp
+        self._update_queue(image_feat_m, text_feat_m)
```
RdoubleA, line 192: question just for my understanding: why are the similarities not symmetric?

torchmultimodal/models/albef.py
```diff
@@ -0,0 +1,244 @@
...
+            [self.vision_proj, self.vision_proj_m],
+            [self.text_proj, self.text_proj_m],
+        ]:
+            for param, param_m in zip(model.parameters(), model_m.parameters()):
+                param_m.data = param_m.data * self.momentum + param.data * (
+                    1 - self.momentum
+                )
+
+    @torch.no_grad()
+    def _update_queue(self, image_feat_m, text_feat_m):
```
RdoubleA, line 156: need type annotations on all functions - unfortunately our mypy config does not catch this right now

RdoubleA (commented):
overall looks awesome! I like that code is pretty straightforward to read and organized well. Perhaps some more comments would be helpful to explain some of the deeper functions (neg_embeddings, update queue). Also docstrings for the ALBEFModel class.

torchmultimodal/models/albef.py
ebsmothers: It might be better to move the call to `update_queue` into `_momentum_update` instead for clarity (and I don't think either `image_feat_m` or `text_feat_m` gets modified between there and here).

Re @RdoubleA 's question, the similarity is computed by comparing the current batch's embeddings to those from the queue, so we want to calculate both `current_batch_image_embeds * full_queue_text_embeds^T` for the current batch's image similarity with all text embeddings and `current_batch_text_embeds * full_queue_image_embeds^T` for the current batch's text similarity with all image embeddings. Btw this is similar to what CLIP does in the distributed case ([ref](https://github.com/facebookresearch/multimodal/blob/main/torchmultimodal/modules/losses/contrastive_loss_with_temperature.py#L110-L117))

ebsmothers (commented):
Overall this looks great! I would also recommend adding a unit test as we discussed (just some simple modules as encoders should be sufficient for now).

ebsmothers (commented):
Overall this looks great! I would also recommend adding a unit test as we discussed (just some simple modules as encoders should be sufficient for now).

RdoubleA (approved):
Tests look well put together to me! Assuming the model tests will come in another PR after the encoders are implemented, this looks good to me.

ebsmothers (approved):

PR #69, opened 2022-06-06
[MUGEN]Spatialtemporal position embedding;tensor slicing utility

langong347:
Summary:
- Spatial-temporal position embedding is the first layer of the multimodal GPT attention stack
- Refactor [`AddBroadcastEmbedding`](https://github.com/mugen-org/MUGEN_baseline/blob/main/lib/models/gpt/attention.py#L256) and the associated utility [`tesor_slice`](https://github.com/mugen-org/MUGEN_baseline/blob/main/lib/models/gpt/utils.py#L75)
- Use `pytest` to perform parametrized tests, test setup and etc. as the starting point to migrate our testing framework away from python `unittest`.

Test Plan:
```
(torchmm) langong-mbp:gpt_attention langong$ python -m pytest --cov=torchmultimodal/modules/layers test/modules/layers/test_position_embedding.py
================================================================================ test session starts =================================================================================
platform darwin -- Python 3.8.13, pytest-7.1.2, pluggy-1.0.0
rootdir: /Users/langong/gpt_attention
plugins: cov-3.0.0
collected 7 items

test/modules/layers/test_position_embedding.py .......                                                                                                                         [100%]

---------- coverage: platform darwin, python 3.8.13-final-0 ----------
Name                                                   Stmts   Miss  Cover
--------------------------------------------------------------------------
torchmultimodal/modules/layers/attention.py               97     97     0%
torchmultimodal/modules/layers/codebook.py                81     81     0%
torchmultimodal/modules/layers/conv.py                    74     74     0%
torchmultimodal/modules/layers/mlp.py                     22     22     0%
torchmultimodal/modules/layers/normalizations.py           7      7     0%
torchmultimodal/modules/layers/position_embedding.py      38      0   100%
torchmultimodal/modules/layers/transformer.py            133    133     0%
--------------------------------------------------------------------------
TOTAL                                                    452    414     8%


====================== 7 passed in 2.38s ==================
```

```
(torchmm) langong-mbp:gpt_attention langong$ python -m pytest --cov=torchmultimodal/utils/ test/utils/test_common.py -vv
================================================================================ test session starts =================================================================================
platform darwin -- Python 3.8.13, pytest-7.1.2, pluggy-1.0.0 -- /Users/langong/local/miniconda3/envs/torchmm/bin/python
cachedir: .pytest_cache
rootdir: /Users/langong/gpt_attention
plugins: cov-3.0.0
collected 6 items

[...]

PR #88, opened 2022-06-14
[Draft] [MUGEN] Add Transformer Decoder

langong347:
* #122
* __->__ #88
* #69


torchmultimodal/modules/layers/transformer_decoder.py
```diff
@@ -0,0 +1,199 @@
...
+            attention_module=FullAttention(shape, True, attention_dropout),
+        )
+        self.pre_mlp_norm = nn.LayerNorm(embedding_dim)
+        self.post_mlp_dropout = nn.Dropout(dropout)
+        self.mlp_block = MLPBlock(embedding_dim, embedding_dim, [embedding_dim * 4])
+
+    def forward(self, x, decode_step, decode_idx):
+        h = self.pre_attention_norm(x)
+        if self.training:
+            h = checkpoint(self.attention, h, h, h, decode_step, decode_idx)
```
ebsmothers, line 95: What's the purpose of checkpointing here?

torchmultimodal/modules/layers/transformer_decoder.py
```diff
@@ -0,0 +1,199 @@
...
+            * self.sos
+        )  # (batch, 1, embedding_dim)
+        sos = sos.type_as(x)
+        # Shift one unit to the right along dim ``seq_len``
+        x = torch.cat([sos, x[:, :-1, :]], axis=1)  # (batch, seq_len, embedding_dim)
+        x = x.view(*x_shape)
+        return x
+
+
+class MLPBlock(MLP):
```
ebsmothers, line 57: Is this based on [this layer](https://github.com/mugen-org/Baseline/blob/7961b537ba49fa5bd1ada099e0f6067894254cc0/lib/models/gpt/attention.py#L101)? If so, can you just use our MLP directly?

torchmultimodal/modules/layers/transformer_decoder.py
```diff
@@ -0,0 +1,199 @@
...
+
+import torch
+from torch import nn
+from torch.utils.checkpoint import checkpoint
+
+from torchmultimodal.modules.layers.attention import FullAttention, MultiHeadAttention
+from torchmultimodal.modules.layers.mlp import MLP
+
+
+class SiLU(nn.Module):
```
ebsmothers, line 15: Is this actually SiLU though? Pytorch already has SiLU as x * sigma(x). I see why you need this one but would consider naming it differently

torchmultimodal/modules/layers/transformer_decoder.py
```diff
@@ -0,0 +1,199 @@
...
+
+    def forward(self, x):
+        return torch.sigmoid(1.702 * x) * x
+
+
+class RightShift(nn.Module):
+    def __init__(self, embedding_dim):
+        super().__init__()
+        self.embedding_dim = embedding_dim
+        self.sos = nn.Parameter(
```
ebsmothers, line 34: sos = start of sentence? Is this implicitly for text only then? Also any particular reason for the choice of initialization?

torchmultimodal/modules/layers/transformer_decoder.py
```diff
@@ -0,0 +1,199 @@
...
+
+    def _get_training_embeddings(self, x, position_embedding):
+        # Use the full sequence for training
+        return x + position_embedding(x)
+
+    def forward(self, x_input, x_output, decode_step=None, decode_idx=None):
+        # Trigger inference mode if either of the input/output-modality sequence is ``None`` as generation
+        # is based the previous data point (autoregressive) where only one modality is needed at any point
+        # along the sequence.
+        if x_output is None:
```
ebsmothers, line 166: Lots of if/elses in this and other forwards, would be nice to clean this up. One suggestion is to decouple some of the embedding logic from the core transformer decoder logic. E.g. create a separate module (or two modules, one for inference and one for training) to take in some combination of x_input, x_output, and decode_step and output x. Then this module will just be right shift + net. You could also use `if self.train` for some of these checks to make the conditions clearer (assuming that makes sense here).

torchmultimodal/modules/layers/transformer_decoder.py
```diff
@@ -0,0 +1,199 @@
...
+                f"must be equal to that of the output modality {position_embedding_output.embedding_dim}"
+            )
+
+        self.embedding_dim = position_embedding_input.embedding_dim
+        self.right_shift = RightShift(self.embedding_dim)
+        self.position_embedding_input = position_embedding_input
+        self.position_embedding_output = position_embedding_output
+        self.input_seq_len = self.position_embedding_input.seq_len
+        self.output_seq_len = self.position_embedding_output.seq_len
+        total_len = self.input_seq_len + self.output_seq_len
```
ebsmothers, line 138: Lots of assumptions that various fields in the embeddings exist, would be sure to check these

torchmultimodal/modules/layers/transformer_decoder.py
```diff
@@ -0,0 +1,199 @@
...
+    def __init__(self, embedding_dim):
+        super().__init__()
+        self.embedding_dim = embedding_dim
+        self.sos = nn.Parameter(
+            torch.FloatTensor(embedding_dim).normal_(std=0.02), requires_grad=True
+        )
+
+    def forward(self, x, decode_step=None):
+        if decode_step is not None and decode_step > 0:
+            return x
```
ebsmothers, line 40: Maybe this can just go in transformer decoder? Imo it would be clearer that way

torchmultimodal/modules/layers/transformer_decoder.py
```diff
@@ -0,0 +1,199 @@
...
+        # is based the previous data point (autoregressive) where only one modality is needed at any point
+        # along the sequence.
+        if x_output is None:
+            # Skip inferencing if ``decode_step = 0`` as decoding requires previous data point.
+            # ``decode_step`` ranges between ``(0, input_seq_len]`` controlled by the GPT model.
+            # When ``decode_step = input_seq_len``, the first output modality is generated from the last input
+            # modality.
+            if decode_step > 0:
+                x_input = self._get_inference_embeddings(
+                    x_input, self.position_embedding_input, decode_step
```
ankitade, line 173: you dont need to pass position embedding

torchmultimodal/modules/layers/transformer_decoder.py
```diff
@@ -0,0 +1,199 @@
...
+
+        self.embedding_dim = position_embedding_input.embedding_dim
+        self.right_shift = RightShift(self.embedding_dim)
+        self.position_embedding_input = position_embedding_input
+        self.position_embedding_output = position_embedding_output
+        self.input_seq_len = self.position_embedding_input.seq_len
+        self.output_seq_len = self.position_embedding_output.seq_len
+        total_len = self.input_seq_len + self.output_seq_len
+
+        self.attention_nets = nn.ModuleList(
```
ankitade, line 140:
```suggestion
        self.transformer_layers = nn.ModuleList(
```

torchmultimodal/modules/layers/transformer_decoder.py
```diff
@@ -0,0 +1,199 @@
...
+            raise ValueError(
+                f"The position embedding dimension of the input modality {position_embedding_input.embedding_dim}"
+                f"must be equal to that of the output modality {position_embedding_output.embedding_dim}"
+            )
+
+        self.embedding_dim = position_embedding_input.embedding_dim
+        self.right_shift = RightShift(self.embedding_dim)
+        self.position_embedding_input = position_embedding_input
+        self.position_embedding_output = position_embedding_output
+        self.input_seq_len = self.position_embedding_input.seq_len
```
ankitade, line 136: hm how does this work, arent seq lengths variable?

torchmultimodal/modules/layers/transformer_decoder.py
```diff
@@ -0,0 +1,199 @@
...
+    def forward(self, x):
+        return torch.sigmoid(1.702 * x) * x
+
+
+class RightShift(nn.Module):
+    def __init__(self, embedding_dim):
+        super().__init__()
+        self.embedding_dim = embedding_dim
+        self.sos = nn.Parameter(
+            torch.FloatTensor(embedding_dim).normal_(std=0.02), requires_grad=True
```
ankitade, line 35: actually why / how are we learning this
consider moving this to somewhere is the model forward or as a util

torchmultimodal/modules/layers/transformer_decoder.py
```diff
@@ -0,0 +1,199 @@
...
+                x_input = self._get_inference_embeddings(
+                    x_input, self.position_embedding_input, decode_step
+                )
+            x = x_input
+        elif x_input is None:
+            # Continue to generate output-modality sequence from ``input_seq_len + 1``(inclusive) to
+            # ``input_seq_len + output_seq_len - 1``(inclusive).
+            x = self._get_inference_embeddings(
+                x_output,
+                self.position_embedding_output,
```
ankitade, line 181: if we want this to become a generic decoder, we need to push this preprocessing to the caller.

torchmultimodal/modules/layers/transformer_decoder.py
```diff
@@ -0,0 +1,199 @@
...
+        # Use the data point at ``decode_step - 1`` to generate the next.
+        return position_embedding(x, decode_step)
+
+    def _get_training_embeddings(self, x, position_embedding):
+        # Use the full sequence for training
+        return x + position_embedding(x)
+
+    def forward(self, x_input, x_output, decode_step=None, decode_idx=None):
+        # Trigger inference mode if either of the input/output-modality sequence is ``None`` as generation
+        # is based the previous data point (autoregressive) where only one modality is needed at any point
```
ankitade, line 164: do we envision single/different decoder for autoregressive and NAR? good to think for unification cc: @RdoubleA

torchmultimodal/modules/layers/transformer_decoder.py
```diff
@@ -0,0 +1,199 @@
...
+        x = x + h
+
+        return x
+
+
+class TransformerDecoder(nn.Module):
+    def __init__(
+        self,
+        position_embedding_input,
+        position_embedding_output,
```
RdoubleA, line 116: I'm assuming these are supposed to be of type `BroadcastedPositionEmbedding`? Are there use cases where a user will use some other position embedding? If so, there are assumptions that whatever the user feeds in will have `seq_len` and `embedding_dim` fields. If only `BroadcastedPositionEmbedding` is used, why not just instantiate it here so the user doesn't have to worry about it?

torchmultimodal/modules/layers/transformer_decoder.py
RdoubleA: also would suggest using `nn.Sequential` instead of `nn.ModuleList` since you will forward pass through these in order anyway. then you can just call `self.transformer_layers(x)` below.

torchmultimodal/modules/layers/transformer_decoder.py
RdoubleA: I think having the positional embedding in this class makes sense since it's closely tied with the transformer decoder (correct me if I'm wrong, or can different types of positional embeddings be used?), but maybe it could be more relegated into preprocessing functions

torchmultimodal/modules/layers/transformer_decoder.py
langong347: This is activation used by the transformer decoder layer

langong347: It's different from pytorch SiLU by a multiplier of 1.702 to the input tensor. Amended docstring to highlight.

torchmultimodal/modules/layers/transformer_decoder.py
```diff
@@ -0,0 +1,199 @@
...
+        embedding_dim,
+        num_attention_heads,
+        num_hidden_layers,
+        dropout,
+        attention_dropout,
+    ):
+        super().__init__()
+        self.pre_attention_norm = nn.LayerNorm(embedding_dim)
+        self.post_attention_dropout = nn.Dropout(dropout)
+        self.attention = MultiHeadAttention(
```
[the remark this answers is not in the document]
langong347: Pytorch MHA does not cache key/value for generative models which I am planning to refactor in a different PR.
I'm monitoring Rafi's PR about MHA and will coordinate with him on that front.

langong347:
Updated PR: #134

PR #89, opened 2022-06-14
Generalize CLIPArchitecture 

sophiazhi:
Summary:
Generalize CLIPArchitecture to allow two encoders of any modalities and added a test suite for CLIPArchitecture. Ultimately, the goal is to support multimodal models beyond image/text, like MUGEN which uses audio/text/video.

Test plan:
Run command `pytest --cov=torchmultimodal/architectures/ test/architectures/test_clip.py::TestCLIPArchitecture -vv` to run the unit test included in this PR.
Screen Shot 2022-06-16 at 8 27 42 PM

### DIFF
--- a/test/architectures/test_clip.py
+++ b/test/architectures/test_clip.py
@@ -18,10 +18,7 @@ class TestCLIPArchitecture:
 
         encoder_a = torch.nn.Linear(5, 3)
         encoder_b = torch.nn.Linear(4, 3)
-        encoders = torch.nn.ModuleDict(
-            {"modality_a": encoder_a, "modality_b": encoder_b}
-        )
-        clip = CLIPArchitecture(encoders=encoders)
+        clip = CLIPArchitecture(encoder_a, encoder_b)
 
         input_a = torch.randint(1, 8, (2, 5), dtype=torch.float)
         input_b = torch.randint(1, 8, (2, 4), dtype=torch.float)
@@ -32,13 +29,15 @@ class TestCLIPArchitecture:
         clip, input_a, input_b = start
         assert isinstance(clip, torch.nn.Module)
 
-        out = clip(modalities={"modality_a": input_a, "modality_b": input_b})
+        out = clip(input_a, input_b)
         assert (
-            hasattr(out, "modality_a") and hasattr(out, "modality_b") and len(out) == 2
+            hasattr(out, "embeddings_a")
+            and hasattr(out, "embeddings_b")
+            and len(out) == 2
         )
 
-        actual_a_embedding = out.modality_a
-        actual_b_embedding = out.modality_b
+        actual_a_embedding = out.embeddings_a
+        actual_b_embedding = out.embeddings_b
         expected_a_embedding = torch.Tensor(
             [[-0.8066, -0.1749, 0.5647], [-0.7709, -0.1118, 0.6271]]
         )
@@ -50,26 +49,4 @@ class TestCLIPArchitecture:
         )
         assert_expected(
             actual=actual_b_embedding, expected=expected_b_embedding, rtol=0, atol=1e-4
-        )
-
-    def test_forward_missing_input(self, start):
-        clip, input_a, _ = start
-
-        with pytest.raises(ValueError):
-            clip(modalities={"modality_a": input_a})
-
-    def test_forward_extra_input(self, start):
-        clip, input_a, input_b = start
-
-        with pytest.warns(UserWarning):
-            out = clip(
-                modalities={
-                    "modality_a": input_a,
-                    "modality_b": input_b,
-                    "extra": torch.Tensor([1]).to(dtype=float),
-                }
-            )
-
-        assert (
-            hasattr(out, "modality_a") and hasattr(out, "modality_b") and len(out) == 2
         )
--- a/test/architectures/test_clip.py
+++ b/test/architectures/test_clip.py
@@ -1,0 +1,81 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+import pytest
+
+import torch
+from test.test_utils import assert_expected, set_rng_seed
+from torchmultimodal.architectures.clip import CLIPArchitecture
+
+
+class TestCLIPArchitecture:
+    @pytest.fixture
+    def start(self):
+        set_rng_seed(1234)
+
+        query_encoder = torch.nn.Linear(5, 3)
+        retrieval_encoder = torch.nn.Linear(4, 3)
+        encoders = torch.nn.ModuleDict(
+            {"query": query_encoder, "retrieval": retrieval_encoder}
+        )
+        clip = CLIPArchitecture(encoders=encoders)
+
+        input_query = torch.randint(1, 8, (2, 5), dtype=torch.float)
+        input_retrieval = torch.randint(1, 8, (2, 4), dtype=torch.float)
+
+        return clip, input_query, input_retrieval
+
+    def test_forward(self, start):
+        clip, input_query, input_retrieval = start
+        assert isinstance(clip, torch.nn.Module)
+
+        out = clip(modalities={"query": input_query, "retrieval": input_retrieval})
+        assert (
+            hasattr(out, "query_embeddings")
+            and hasattr(out, "retrieval_embeddings")
+            and len(out.__dict__) == 2
+        )
+
+        actual_q_embedding = out.query_embeddings
+        actual_r_embedding = out.retrieval_embeddings
+        expected_q_embedding = torch.Tensor(
+            [[-0.8066, -0.1749, 0.5647], [-0.7709, -0.1118, 0.6271]]
+        )
+        expected_r_embedding = torch.Tensor(
+            [[-0.1719, 0.7932, 0.5842], [-0.2805, 0.8761, -0.3921]]
+        )
+        assert_expected(
+            actual=actual_q_embedding, expected=expected_q_embedding, rtol=0, atol=1e-4
+        )
+        assert_expected(
+            actual=actual_r_embedding, expected=expected_r_embedding, rtol=0, atol=1e-4
+        )
+
+    def test_forward_missing_input(self, start):
+        clip, input_query, _ = start
+        assert isinstance(clip, torch.nn.Module)
+
+        with pytest.raises(AssertionError):
+            clip(modalities={"query": input_query})
+
+    def test_forward_extra_input(self, start):
+        clip, input_query, input_retrieval = start
+        assert isinstance(clip, torch.nn.Module)
+
+        with pytest.warns(UserWarning):
+            out = clip(
+                modalities={
+                    "query": input_query,
+                    "retrieval": input_retrieval,
+                    "extra": torch.Tensor([1]).to(dtype=float),
+                }
+            )
+
+        assert (
+            hasattr(out, "query_embeddings")
+            and hasattr(out, "retrieval_embeddings")
+            and len(out.__dict__) == 2
+        )
--- a/test/architectures/test_clip.py
+++ b/test/architectures/test_clip.py
@@ -16,66 +16,60 @@ class TestCLIPArchitecture:
     def start(self):
         set_rng_seed(1234)
 
-        query_encoder = torch.nn.Linear(5, 3)
-        retrieval_encoder = torch.nn.Linear(4, 3)
+        encoder_a = torch.nn.Linear(5, 3)
+        encoder_b = torch.nn.Linear(4, 3)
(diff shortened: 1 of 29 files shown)


torchmultimodal/architectures/clip.py
```diff
@@ -20,32 +22,38 @@ class CLIPArchitecture(nn.Module):
...
+        embeddings = {}
+        for key, encoder in self.encoders.items():
+            assert key in modalities, f"{key} missing in input"
+            embeddings[f"{key}_embeddings"] = F.normalize(encoder(modalities[key]))
+        for key in modalities.keys():
+            if key not in self.encoders:
+                warnings.warn(f"Missing encoder for extra input {key}")
+
+        # Return a dataclass object instead of a dictionary
+        clip_output = make_dataclass(
```
ebsmothers, line 55: Is there a specific reason we want to return a dataclass here? Imo one of the main advantages of dataclasses is that they follow a fixed schema, so returning one dynamically feels a bit unnatural.

torchmultimodal/architectures/clip.py
```diff
@@ -20,32 +22,38 @@ class CLIPArchitecture(nn.Module):
...
+            "CLIPOutput",
+            [(f"{k}_embeddings", torch.Tensor) for k in self.encoders.keys()],
```
ebsmothers, line 57: Why return a different key than what the user passes in? This could be unintuitive and they will inevitably have to come here and look at the code when they get a key error

torchmultimodal/architectures/clip.py
```diff
@@ -20,32 +22,38 @@ class CLIPArchitecture(nn.Module):
...
-
-        img_embeddings = self.vision_encoder(image)
-        text_embeddings = self.text_encoder(text)
-        img_embeddings = F.normalize(img_embeddings)
-        text_embeddings = F.normalize(text_embeddings)
-        return {"image": img_embeddings, "text": text_embeddings}
+        embeddings = {}
+        for key, encoder in self.encoders.items():
+            assert key in modalities, f"{key} missing in input"
+            embeddings[f"{key}_embeddings"] = F.normalize(encoder(modalities[key]))
```
ebsmothers, line 49: Random question: do we definitely want normalize in the architecture? (I know it's already here now so not necessarily relevant to your PR, but this is another piece we could consider making the encoder's responsibility, similar to projections as we discussed)

torchmultimodal/architectures/clip.py
```diff
@@ -20,32 +22,38 @@ class CLIPArchitecture(nn.Module):
...
+        for key in modalities.keys():
+            if key not in self.encoders:
+                warnings.warn(f"Missing encoder for extra input {key}")
```
ebsmothers, line 52: I think your choice to raise a warning here makes sense. We might also want to do the same in [late_fusion](https://github.com/facebookresearch/multimodal/blob/main/torchmultimodal/architectures/late_fusion.py#L46-L48) for the sake of consistency (doesn't have to be done in this PR though)

test/architectures/test_clip.py
```diff
@@ -0,0 +1,81 @@
...
+from torchmultimodal.architectures.clip import CLIPArchitecture
+
+
+class TestCLIPArchitecture:
+    @pytest.fixture
+    def start(self):
+        set_rng_seed(1234)
+
+        query_encoder = torch.nn.Linear(5, 3)
+        retrieval_encoder = torch.nn.Linear(4, 3)
```
ebsmothers, line 20: nit: since you are showing off how you made the architecture more general, I would not anchor the names here to a retrieval use case. Instead you could just name them `encoder_a` and `encoder_b` or something, with corresponding keys (just an example, I'm sure you can come up with something better than that 🙂 )

test/architectures/test_clip.py
```diff
@@ -0,0 +1,81 @@
...
+        )
+
+    def test_forward_missing_input(self, start):
+        clip, input_query, _ = start
+        assert isinstance(clip, torch.nn.Module)
+
+        with pytest.raises(AssertionError):
+            clip(modalities={"query": input_query})
+
+    def test_forward_extra_input(self, start):
```
ebsmothers, line 64: I like this test case!

ebsmothers (changes requested):
Overall this looks pretty good! Left a few comments, but other than the stuff about the forward outputs, they're all relatively minor

ankitade (changes requested):
Thanks for the changes
I dont think we should change CLIP which is a "standard" model to make it play nice with mugen. Other options are to either have a different model if we want to finally get to another "Standard" model like video clip or have a version in examples/mugen

torchmultimodal/architectures/clip.py
sophiazhi: I agree it feels unnatural (it took me a while to figure out how to make a dataclass dynamically). I used a dataclass to match the pattern set by other modules, but now I realize a lot of modules don't have it, so unless anyone is a strong proponent of output classes then I can return a dictionary instead

test/architectures/test_clip.py
langong347: Make sense. The relationship "query" and "retrieval" is meaningful for the retrieval task per se. For CLIP model it's just a matter of computing the similarities between the two modalities (or more).

torchmultimodal/architectures/clip.py
langong347: The L_p normalization puts the encodings for different modalities within the same scale `(0, 1)`? I suppose that's an essential step before computing the cosine similarity. The CLIP paper also has this step: https://arxiv.org/pdf/2103.00020.pdf.

ebsmothers:
@langong347 I do see @ankitade's point here. At the very least this is no longer really `CLIPArchitecture` since the LI in CLIP stand for language and image 🙂. A separate question is whether we want to keep `CLIPArchitecture` because it is a foundational model. Seems like the two options would be to either

(1) keep `CLIPArchitecture` as is and implement a separate generalized architecture to be used by MUGEN, or 
(2) rename this to e.g. `ContrastiveArchitecture` and let both CLIP and MUGEN use it. 

To me, the argument for (1) is that CLIP is a very important model and should be a first-class citizen with its own architecture. While the argument for (2) is better generality (I think we have said we should not have an architecture unless it is used by multiple models anyways). Personally, I would lean slightly towards (2) but would like to hear others' thoughts as well.

torchmultimodal/architectures/clip.py
```diff
@@ -20,32 +22,38 @@ class CLIPArchitecture(nn.Module):
...
     ) -> Dict[str, torch.Tensor]:
-
-        img_embeddings = self.vision_encoder(image)
-        text_embeddings = self.text_encoder(text)
-        img_embeddings = F.normalize(img_embeddings)
-        text_embeddings = F.normalize(text_embeddings)
-        return {"image": img_embeddings, "text": text_embeddings}
+        embeddings = {}
+        for key, encoder in self.encoders.items():
+            assert key in modalities, f"{key} missing in input"
```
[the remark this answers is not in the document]
sophiazhi: i think a ValueError is more fitting since the input is missing a key, rather than the input containing a bad key, so I've changed the assertionerror to a valueerror

RdoubleA:
@ebsmothers I'm leaning towards option 1. `CLIPArchitecture` is just a convenient wrapper around encoder -> projection -> l2 norm -> output (on a separate note, where is the projection layer?). Since it has been used many times in different papers, I think that warrants it's own architecture even though it would be an image-text instance of a general `ContrastiveArchitecture` (much like Video VQVAE gets its own file even though it's an instance of VQVAE). 

As for MUGEN, the linear projection layer after the encoder is slightly different than the CLIP paper (which only uses one linear layer I believe?): https://github.com/mugen-org/MUGEN_baseline/blob/02c7058cd221f4b651d4ace2276b085cac1c5efd/lib/models/videoclip/modules.py#L15. So that leads me to believe MUGEN should have its own `ContrastiveArchitecture`.

As for supporting more than two encoders, I'm not convinced of the benefit for that over multiple CLIPs, other than convenience for getting all three embeddings at once for training or inference. That seems MUGEN specific, warranting the separate contrastive architecture for MUGEN anyway.

test/architectures/test_clip.py
```diff
@@ -0,0 +1,81 @@
...
+        )
+        assert_expected(
+            actual=actual_r_embedding, expected=expected_r_embedding, rtol=0, atol=1e-4
+        )
+
+    def test_forward_missing_input(self, start):
+        clip, input_query, _ = start
+        assert isinstance(clip, torch.nn.Module)
+
+        with pytest.raises(AssertionError):
```
RdoubleA, line 61: you've changed the assertion to ValueError, so should change it here as well

torchmultimodal/architectures/clip.py
```diff
@@ -20,32 +22,38 @@ class CLIPArchitecture(nn.Module):
...
     def __init__(
         self,
-        vision_encoder: nn.Module,
-        text_encoder: nn.Module,
+        encoders: nn.ModuleDict,
     ):
         super().__init__()
-        self.vision_encoder = vision_encoder
-        self.text_encoder = text_encoder
+        self.encoders = nn.ModuleDict({k: encoders[k] for k in sorted(encoders.keys())})
```
RdoubleA, line 40: why do you need to recreate the ModuleDict here?

torchmultimodal/architectures/clip.py
RdoubleA: I would prefer `NamedTuple` for consistency with all our other model outputs, unless there's a clear advantage of using `dataclass` over `NamedTuple`

RdoubleA (changes requested):

torchmultimodal/architectures/clip.py
sophiazhi: creating NamedTuple dynamically causes issues with mypy such that i have to include `# type: ignore` on the namedtuple creation line, but besides that i don't see other relative advantages of dataclass

ankitade:
1. In the final state, we should just have models/clip.py with the CLIPModel (some version of current CLIPArch) + different standard instantiations clip_vit16 etc (we already have this i think)

2. For video clip (standard one, not mugen specific, if they dont line up), there can be models/video_clip.py if we want to add it and it does something different like Rafi's point about projections being different 

3. If we start making clip take in a dict of modalities, its confusing for people for whom contrastive loss with batch negatives work for 2 dimensions (aka modalities for us).  Actually I don't understand how the loss and 0 shot will eventually work for more than 2 entries in the dict

sophiazhi:
> where is the projection layer?

Both the original CLIPArchitecture and this generalized CLIPArchitecture avoided explicitly including the projection layer(s) because users may want different types of projections and projection logic can be folded into the encoder that's passed in. We also can't guarantee that any projections passed in as arguments by the user have the same output size, so I don't see an advantage to including a projection argument. (Though this choice does assume that we want one general clip architecture and not two versions)

langong347:
>For video clip (standard one, not mugen specific, if they dont line up), there can be models/video_clip.py if we want to add it and it does something different like Rafi's point about projections being different

The projection can be absorbed into the encoders (see Sophia's post), we can reuse the same CLIPArchitecture for arbitrary pair of modalities --- that's how CLIP has been extended for other research.

ebsmothers:
Agree with both @langong347 and @sophiazhi's points about keeping the projection layer out of the architecture. Even in CLIP the projection layer is not guaranteed to be present (I think they have one in the ViT version but not the ResNet one). A corollary to this would be that MUGEN does not need its own architecture just because it has a different projection layer.

I wouldn't recommend returning the similarity as part of the architecture though. Then we are starting to integrate our loss into the architecture, which we don't want in general. This had to be done for ALBEF because of how the similarities get used in the multimodal encoder, but this is also part of the reason that class was implemented as a model and not an architecture (since it then becomes much more specific to that particular model). Also, even simple old cosine similarity can be implemented in different ways, with both FLAVA and CLIP handling propagation of gradients differently. So I would keep this out and leave it up to the user how to use the embeddings.

For @ankitade's 3rd point, agree that returning more than two modalities doesn't really make sense for zero shot. Though hopefully if the user is running zero shot (or contrastive with batch negatives), they wouldn't pass more than two modalities anyways. However, these assumptions plus excessive generality could potentially cause confusion for the users on the "flagship" instantiation of CLIPArchitecture (CLIP 😉 ).

So ultimately I agree with @RdoubleA: leaving CLIPArchitecture as is feels like the right path. But I do think we should generalize the MUGEN architecture to return a dict (as opposed to an arbitrary pair). Otherwise in a case like [this](https://github.com/mugen-org/MUGEN_baseline/blob/b45da493f3ddbcced52d44eee02d443f90842474/lib/models/videoclip/clip.py#L75-L79), we would have to call each of the encoders multiple times. 

(As an aside, this whole convo is yet another interesting test of our "don't generalize until you need to" principle...)

langong347 (approved):

PR #77, opened 2022-06-08
Add MDETR transformer and model class

ebsmothers:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* #110
* #109
* __->__ #77

This PR adds the multimodal transformer and main model class for MDETR. Similar to the previous PRs, this is still an initial version. The transformer closely follows the original
implementation, but without the intermediate caching of encoder outputs. The model class has been decoupled from the losses and takes in all encoders, transformers, and various embedding or
projection modules and returns classification logits and their corresponding bounding boxes in its forward.

Rather than writing a unit test for the class, I've added a notebook that demonstrates how to load weights from the pretrained model, call forward, and check that the results match.

Differential Revision: [D37390042](https://our.internmc.facebook.com/intern/diff/D37390042)

PR #90, opened 2022-06-15
Add Flickr postprocessing transform for phrase grounding.

ebsmothers:
* #103
* __->__ #90
* #77
* #62
* #61

Test plan:
Added a unit test under examples/mdetr
python -m pytest examples/mdetr/test/test_transforms.py
======================================= test session starts ========================================
platform linux -- Python 3.8.13, pytest-7.1.2, pluggy-1.0.0
rootdir: /data/home/ebs/torchmultimodal
collected 4 items

examples/mdetr/test/test_transforms.py ....                                                  [100%]

======================================== 4 passed in 3.56s =========================================

### DIFF
--- a/examples/mdetr/test/test_transforms.py
+++ b/examples/mdetr/test/test_transforms.py
@@ -1,0 +1,196 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+from copy import deepcopy
+
+import pytest
+import torch
+from examples.mdetr.transforms import box_cxcywh_to_xyxy, FlickrPostProcessTransform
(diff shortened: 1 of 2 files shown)


ankitade (commented):
High level noob q, this utility is for processing model outputs right and not a traditional transform which is generally applied to batches fed to the model. whats the rationale for making this a "transform" vs just a function?

ebsmothers:
@ankitade fair question. Your understanding is correct; I wrote it this way mostly based on personal preference. The original implementation does have it as a function, so if you feel it's clearer that way I can change it back

ankitade:
> @ankitade fair question. Your understanding is correct; I wrote it this way mostly based on personal preference. The original implementation does have it as a function, so if you feel it's clearer that way I can change it back

i dont think it changes much implementation wise. maybe just avoid transforms in the class name and module name to prevent confusion among people familiar with torch* concept of transforms

PR #91, opened 2022-06-15
[ALBEF] vision encoder

katrina433:
Summary:

Implemented `ALBEFVisionEncoder`, which is a wrapper of `torchvision`'s  `VisionTransformer` implementation. The only change is removed the `heads` from the `VisionTransformer`. 

Test plan:

* Started by copying the entire [`vit.py`](https://github.com/salesforce/ALBEF/blob/6224e78e5292757ca9c7d5add3e415010955a58b/models/vit.py) from ALBEF repo, and its import from timm ([PatchEmbed](https://github.com/rwightman/pytorch-image-models/blob/master/timm/models/layers/patch_embed.py))
* Wrote tests for ALBEF's `VisionTransformer`, as well as its building blocks (with assigned `state_dict` so can regenerate the models later)
* Wrote the wrapper class `ALBEFVisionEncoder`, and replaced the testing components with the new implementation (loaded with the same `state_dict` from before)
* Checked that the outputs remained the same

### DIFF
--- a/test/modules/encoders/test_albef_vision_encoder.py
+++ b/test/modules/encoders/test_albef_vision_encoder.py
@@ -1,0 +1,31 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+from typing import OrderedDict
+
+import torch
+from test.test_utils import assert_expected, set_rng_seed
+from torch import Tensor
+from torchmultimodal.modules.encoders.albef_vision_encoder import PatchEmbed
+
+
+class TestALBEFVisionEncoder:
+    def test_conv_proj(self):
+        set_rng_seed(0)
+        state_dict = OrderedDict(
+            [("proj.weight", torch.randn(3, 3, 4, 4)), ("proj.bias", torch.randn(3))]
+        )
+        conv_proj = PatchEmbed(
+            img_size=4,
+            patch_size=4,
+            in_chans=3,
+            embed_dim=3,
+        )
+        conv_proj.load_state_dict(state_dict)
+        input = torch.randn(1, 3, 4, 4)
+        output = conv_proj(input)
+        expected = Tensor([14.094812, 13.545728, 14.223189]).reshape(1, 1, 3)
+        assert_expected(output, expected)
--- a/test/modules/encoders/test_albef_vision_encoder.py
+++ b/test/modules/encoders/test_albef_vision_encoder.py
@@ -6,6 +6,7 @@
 
 from functools import partial
 
+import pytest
 import torch
 from test.test_utils import assert_expected, set_rng_seed
 from torch import nn, Tensor
@@ -68,3 +69,23 @@ class TestALBEFVisionEncoder:
             ]
         ).unsqueeze(0)
         assert_expected(output, expected, rtol=0, atol=1e-4)
+
+    def test_invalid_input_length(self):
+        input = torch.randn(3, 4, 4)
+        with pytest.raises(ValueError, match="not enough values to unpack"):
+            self.vision_encoder(input)
+
+    def test_invalid_image_channel_dim(self):
+        input = torch.rand(1, 1, 4, 4)
+        with pytest.raises(RuntimeError, match="channels"):
+            self.vision_encoder(input)
+
+    def test_invalid_image_height(self):
+        input = torch.rand(1, 3, 5, 4)
+        with pytest.raises(AssertionError, match="Wrong image height!"):
+            self.vision_encoder(input)
+
+    def test_invalid_image_width(self):
+        input = torch.rand(1, 3, 4, 3)
+        with pytest.raises(AssertionError, match="Wrong image width!"):
+            self.vision_encoder(input)
--- a/test/modules/encoders/test_albef_vision_encoder.py
+++ b/test/modules/encoders/test_albef_vision_encoder.py
@@ -9,12 +9,13 @@ from typing import OrderedDict
 import torch
 from test.test_utils import assert_expected, set_rng_seed
 from torch import Tensor
-from torchmultimodal.modules.encoders.albef_vision_encoder import PatchEmbed
+from torchmultimodal.modules.encoders.albef_vision_encoder import Attention, PatchEmbed
 
 
 class TestALBEFVisionEncoder:
     def test_conv_proj(self):
         set_rng_seed(0)
+        input = torch.randn(1, 3, 4, 4)
         state_dict = OrderedDict(
             [("proj.weight", torch.randn(3, 3, 4, 4)), ("proj.bias", torch.randn(3))]
         )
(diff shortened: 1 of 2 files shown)


torchmultimodal/modules/encoders/albef_vision_encoder.py
```diff
@@ -0,0 +1,49 @@
...
+            image_size=image_size,
+            patch_size=patch_size,
+            num_layers=num_layers,
+            num_heads=num_heads,
+            hidden_dim=hidden_dim,
+            mlp_dim=mlp_dim,
+            num_classes=num_classes,
+            norm_layer=norm_layer,
+        )
+        self.heads = None
```
ebsmothers, line 37: What is this used for?

torchmultimodal/modules/encoders/albef_vision_encoder.py
```diff
@@ -0,0 +1,49 @@
...
+
+from functools import partial
+from typing import Callable
+
+import torch
+from torch import nn
+from torchvision.models.vision_transformer import VisionTransformer
+
+
+class ALBEFVisionEncoder(VisionTransformer):
```
ebsmothers, line 15: Inheritance?! 😱 (Kidding, mostly.) But we should have a good reason to do this

torchmultimodal/modules/encoders/albef_vision_encoder.py
```diff
@@ -0,0 +1,49 @@
...
+    def forward(self, x: torch.Tensor):
+        # Reshape and permute the input tensor
+        x = self._process_input(x)
+        n = x.shape[0]
+
+        # Expand the class token to the full batch
+        batch_class_token = self.class_token.expand(n, -1, -1)
+        x = torch.cat([batch_class_token, x], dim=1)
+
+        x = self.encoder(x)
```
ebsmothers, line 48: Why not just import the encoder from torchvision and set that as self.encoder in the init? Then stitch together with the conv layer used in _process_input, and you should be able to remove the use of inheritance altogether

test/modules/encoders/test_albef_vision_encoder.py
```diff
@@ -0,0 +1,117 @@
...
+        + [("conv_proj." + key, val) for key, val in proj_state_dict.items()]
+        + [
+            ("encoder.layers.encoder_layer_0." + key, val)
+            for key, val in encoder_block_state_dict.items()
+        ]
+        + [
+            ("encoder.layers.encoder_layer_1." + key, val)
+            for key, val in encoder_block_state_dict.items()
+        ]
+    )
```
ebsmothers, line 65: In this case it might be clearer to just set the values directly rather than using load_state_dict a bunch of times

test/modules/encoders/test_albef_vision_encoder.py
```diff
@@ -0,0 +1,117 @@
...
+from torchmultimodal.modules.encoders.albef_vision_encoder import ALBEFVisionEncoder
+
+
+class TestALBEFVisionEncoder:
+    set_rng_seed(0)
+    input = torch.randn(1, 3, 4, 4)
+    proj_input = torch.randn(1, 1, 3)
+    proj_state_dict = OrderedDict(
+        [("weight", torch.randn(3, 3, 4, 4)), ("bias", torch.randn(3))]
+    )
```
ebsmothers, line 22: Should all this stuff go in the setUp method?

test/modules/encoders/test_albef_vision_encoder.py
```diff
@@ -0,0 +1,117 @@
...
+        patch_size=4,
+        num_layers=2,
+        num_heads=1,
+        hidden_dim=3,
+        mlp_dim=6,
+        norm_layer=partial(nn.LayerNorm, eps=1e-6),
+        num_classes=3,
+    )
+
+    def test_conv_proj(self):
```
ebsmothers, line 77: I like that you're thoroughly testing each one of the components here. I think some of the cleanup I mention above, plus some comments on the test, will help elucidate this to the code reader

test/modules/encoders/test_albef_vision_encoder.py
```diff
@@ -0,0 +1,117 @@
...
+        vit = self.vision_encoder
+        vit.load_state_dict(self.vit_state_dict)
+        output = vit(self.input)
+        expected = Tensor(
+            [
+                [-2.992489e-02, -3.361803e-01, 2.282364e00],
+                [-1.769625e-03, -3.862467e-01, 2.380805e00],
+            ]
+        ).unsqueeze(0)
+        assert_expected(output, expected, rtol=0, atol=1e-4)
```
ebsmothers, line 117: If possible, also add some kind of test for invalid inputs and torchscript

torchmultimodal/modules/encoders/albef_vision_encoder.py
```diff
@@ -0,0 +1,49 @@
...
+            num_layers=num_layers,
+            num_heads=num_heads,
+            hidden_dim=hidden_dim,
+            mlp_dim=mlp_dim,
+            num_classes=num_classes,
+            norm_layer=norm_layer,
+        )
+        self.heads = None
+
+    def forward(self, x: torch.Tensor):
```
ebsmothers, line 39: add return type

torchmultimodal/modules/encoders/albef_vision_encoder.py
```diff
@@ -0,0 +1,49 @@
...
+from functools import partial
+from typing import Callable
+
+import torch
+from torch import nn
+from torchvision.models.vision_transformer import VisionTransformer
+
+
+class ALBEFVisionEncoder(VisionTransformer):
+    def __init__(
```
ebsmothers, line 16: Make sure to add the docstring when you get a chance

ebsmothers (changes requested):

torchmultimodal/modules/encoders/albef_vision_encoder.py
```diff
@@ -0,0 +1,128 @@
...
+        self,
+        image_size: int = 256,
+        patch_size: int = 16,
+        num_layers: int = 12,
+        num_heads: int = 12,
+        hidden_dim: int = 768,
+        mlp_dim: int = 3072,
+        dropout: float = 0.0,
+        attention_dropout: float = 0.0,
+        norm_layer: Callable[..., torch.nn.Module] = partial(nn.LayerNorm, eps=1e-6),
```
RdoubleA, line 48: nit: remove torch here?
```suggestion
        norm_layer: Callable[..., nn.Module] = partial(nn.LayerNorm, eps=1e-6),
```

RdoubleA (approved):

torchmultimodal/modules/encoders/albef_vision_encoder.py
```diff
@@ -0,0 +1,128 @@
...
+    ):
```
RdoubleA, line 49: add return type None here
```suggestion
    ) -> None:
```

ebsmothers (commented):
Thanks for the changes, I think this looks a lot better. One more suggestion: I notice that you have to copy a lot of the stuff from TorchVision's ViT (like `_process_input`) in order to avoid using inheritance. I understand why and think ultimately this is reasonable. But want to present another option as well. TorchVision has a util to return intermediate layers (see [here](https://github.com/pytorch/vision/blob/e722e9c7d128b0d8f12cdf956469ea7de18f3821/torchvision/models/_utils.py#L13)). It has some assumptions, but it may work for this case, and if so would remove some of the copied code. Basically something like

```
from torchvision.models._utils import IntermediateLayerGetter
from torchvision.models.vision_transformer import VisionTransformer
vit = VisionTransformer(args)
vit_encoder =  IntermediateLayerGetter(vit, return_layers={"encoder": "call_this_whatever"})
```

You'd need to check it will line up with the ALBEF version (I'm not positive offhand). If you can't make it work, I think the current solution is good too.

torchmultimodal/modules/encoders/albef_vision_encoder.py
YosuaMichael: Agree, having a docstring will be good

ebsmothers (approved):
Looks great! One final nit is to leave a comment explaining the use of `encoder.ln` (and maybe just define this string as a class variable since it's used in both init and forward). Otherwise this should be good to go!

PR #97, opened 2022-06-16
datapipe based datamodule to replace multidataloader

ninginthecloud:
Summary:
accessing multiple data sources is required for multimodal. Current solution is to create an multidataloader and self defined iterator to going through different data sources. This solution is not very easy to scale. 
To solve this issue, we are prototyping to adopt datapipes from torchdata (0.3.0) to replace multidataloader and achieve native pytorch dataloader experience. 

Test plan:

Fixes #{issue number}


examples/flava/data/datapipes.py
```diff
@@ -0,0 +1,428 @@
...
+    def train_datapipe(self):
+        return self._build_datapipe(datapipe=self.train_datapipe_source,collate_fn=self._build_collator)
+
+    def val_datapipe(self):
+        return self._build_datapipe(datapipe=self.val_datapipe_source, collate_fn=self._build_collator, shuffle=False)
+
+    def _build_datapipe(self, datapipe: IterDataPipe,  collate_fn : Any, drop_last: bool = False, shuffle: bool = True):
+        output_datapipe = datapipe.batch(batch_size=self.batch_size, drop_last=drop_last).collate(collate_fn)
+        if shuffle:
+            return output_datapipe.shuffle()
```
ejguan, line 117: `sharding_filter` should be placed after `shuffle` here. And, you can always add `shuffle` in the pipeline. If you want to disable shuffling explicitly, you can either:
- `shuffle_dp = shuffle_dp.set_shuffle(False)`
- `DataLoader(datapipe, shuffle=False)`

PR #87, opened 2022-06-14
[MUGEN] Add Video VQVAE encoder and decoder

RdoubleA:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* #101
* #98
* __->__ #87
* #86

Add Video VQVAE specific encoder and decoder from [MUGEN](https://github.com/mugen-org/MUGEN_baseline/blob/main/lib/models/video_vqvae/vqvae.py).

Differential Revision: [D37288372](https://our.internmc.facebook.com/intern/diff/D37288372)

PR #86, opened 2022-06-14
[MUGEN] Add axial attention blocks for encoder and decoder

RdoubleA:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* #101
* #98
* #87
* __->__ #86

## Summary
Refactor `AxialBlock` and `AttentionResidualBlock` from [MUGEN](https://github.com/mugen-org/MUGEN_baseline/blob/main/lib/models/video_vqvae/vqvae.py). Included some minor edits to `MultiHeadAttention` after testing attention blocks. `AxialBlock` was placed in `attention.py` since it can be generally used to apply axial attention to n-dimensional inputs. `AttentionResidualBlock` was placed in `video_vqvae.py` since it is specifically used by that model and is not yet generalized.

## Test Plan
Compare output to MUGEN's code directly and run added unit tests.

`AxialBlock`
image

`AttentionResidualBlock`
image

```
test/modules/layers/test_attention.py::TestAttention::test_axial_attention PASSED                                             [ 57%]
test/modules/layers/test_attention.py::TestAttention::test_axial_block_channel_dim PASSED                                     [ 58%]
test/modules/layers/test_attention.py::TestAttention::test_axial_block_forward PASSED                                         [ 59%]
test/models/test_video_vqvae.py::TestAttentionResidualBlock::test_hidden_dim_assertion PASSED                                 [ 14%]
test/models/test_video_vqvae.py::TestAttentionResidualBlock::test_forward PASSED                                              [ 15%]
```

Differential Revision: [D37288326](https://our.internmc.facebook.com/intern/diff/D37288326)

PR #101, opened 2022-06-19
Make shape optional in MultiHeadAttention

RdoubleA:
`shape` is only used when `causal=True`, so this PR makes shape optional and checks that it isn't None when causal is used. This simplifies the API for non-causal MHA.

* __->__ #101
* #98
* #87
* #86

Differential Revision: [D37745470](https://our.internmc.facebook.com/intern/diff/D37745470)


torchmultimodal/modules/layers/attention.py
```diff
@@ -102,15 +100,18 @@ class MultiHeadAttention(nn.Module):
...
         dim_q: int,
         dim_kv: int,
         n_head: int,
         n_layer: int,
         causal: bool,
         attn_module: nn.Module,
+        shape: Optional[Tuple[int]] = None,
     ) -> None:
         super().__init__()
+        if causal and shape is None:
```
ankitade, line 112: and other way round also right is causal is false, shape should be none

ankitade (approved):

torchmultimodal/modules/layers/attention.py
RdoubleA: shape is ignored if causal is false, so it should be ok as is

RdoubleA:
@RdoubleA has imported this pull request.  If you are a Facebook employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D37745470).

PR #104, opened 2022-06-20
Support S3D-backed video encoder

sophiazhi:
Summary:
For the S3D-backed video encoder used by MUGEN video-text retrieval, added:

- `examples/mugen/retrieval/s3d.py`: S3D model
- `examples/mugen/test/retrieval/test_s3d.py`: S3D model unit tests
- `examples/mugen/retrieval/video_clip.py`: video encoder
- `examples/mugen/test/retrieval/test_video_clip.py`: video encoder tests

Test plan:
For a coverage report, temporarily remove the "*examples/*" lines from `.coveragerc`, then run `python -m pytest --cov=examples/mugen/retrieval examples/mugen/test/retrieval/test_s3d.py -vv`. Otherwise, run `python -m pytest examples/mugen/test/retrieval/test_s3d.py::TestS3D -rP`.
Similarly,
`python -m pytest examples/mugen/test/retrieval/test_video_clip.py::TestVideoCLIP -rP`
Screen Shot 2022-06-27 at 11 27 53 AM
Screen Shot 2022-06-27 at 11 28 23 AM


examples/s3d/test_s3d.py
```diff
@@ -0,0 +1,88 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
```
langong347, line 1: Let's get some feedback about how we should structure test folder for examples. Tentatively, this could be:
```
examples/my_project/
    tests/
        test_module_1.py
        test_module_2.py
    module_1.py
    module_2.py
```

examples/s3d/test_s3d.py
```diff
@@ -0,0 +1,88 @@
...
+
+import importlib
+
+import pytest
+import torch
+from examples.s3d.s3d import BasicConv3d, S3D, S3DEncoder, SepConv3d
+from test.test_utils import assert_expected, set_rng_seed
+
+
+class TestS3D:
```
langong347, line 15: Separate the tests for the encoder and the backbone (`S3D`).

examples/s3d/test_s3d.py
```diff
@@ -0,0 +1,88 @@
...
+    def test_s3d_encoder(self, start):
+        make_input_video, _ = start
+        input_video = make_input_video(c_dim=-1)
+        s3d_e = S3DEncoder(trainable=True)
+        out = s3d_e(input_video)
+        expected_sum = 809.0687
+        assert_expected(
+            actual=out.shape, expected=torch.Size([2, 1024])
+        )  # batch x embedding
+        if abs(out.sum() - expected_sum) > 5e-4:
```
langong347, line 39: Why do we need the comparison here? Is it equivalent to `assert_expected(out.sum(), expected, atol=1e-4)`?

sophiazhi: it's equivalent except that `assert_expected` only takes tensors. i'll use `torch.as_tensor(expected_sum)`

examples/mugen/retrieval/s3d.py
```diff
@@ -0,0 +1,372 @@
...
+        x = self.bn_s(x)
+        x = self.relu_s(x)
+
+        x = self.conv_t(x)
+        x = self.bn_t(x)
+        x = self.relu_t(x)
+        return x
+
+
+class Mixed3b(nn.Module):
```
RdoubleA, line 114: do you know the reason for the Mixed names? although I can't think of a clearer name for these...

examples/mugen/test/retrieval/test_s3d.py
```diff
@@ -0,0 +1,76 @@
...
+import importlib
+
+import pytest
+import torch
+from examples.mugen.retrieval.s3d import BasicConv3d, S3D, SepConv3d
+from test.test_utils import assert_expected, set_rng_seed
+
+
+class TestS3D:
+    @pytest.fixture
```
RdoubleA, line 16: add autouse?

examples/mugen/retrieval/s3d.py
sophiazhi: because they use a mix of 3D and 2D convolutions! [paper](https://arxiv.org/pdf/1712.04851.pdf) page 3

sophiazhi: I'm not sure about the 3b part though

examples/mugen/test/retrieval/test_s3d.py
sophiazhi: i still have to request the fixture to get the returned functions but i added autouse to a new fixture that only calls set_rng_seed

examples/mugen/retrieval/s3d.py
```diff
@@ -0,0 +1,372 @@
...
+            Mixed5b(),
+            Mixed5c(),
+        )
+        self.fc = nn.Sequential(
+            nn.Conv3d(1024, num_class, kernel_size=1, stride=1, bias=True),
+        )
+
+    def forward(self, x):
+        y = self.base(x)
+        y = F.avg_pool3d(y, (2, y.size(3), y.size(4)), stride=1)
```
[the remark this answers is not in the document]
sophiazhi: I think because the kernel uses y.size(3) and y.size(4), which are dependent on the size of the input x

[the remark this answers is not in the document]
RdoubleA: good point

examples/mugen/retrieval/s3d.py
RdoubleA: maybe you could rename to MixedConvs3b? your call though

examples/mugen/test/retrieval/test_video_clip.py
```diff
@@ -0,0 +1,37 @@
...
+        make_input_video = start
+        input_video = make_input_video()
+        encoder = VideoEncoder()
+        out = encoder(input_video)
+        expected_sum = 846.3781
+        assert_expected(
+            actual=out.shape, expected=torch.Size([2, 1024])
+        )  # batch x embedding
+        assert_expected(
+            actual=out.sum(), expected=torch.as_tensor(expected_sum), rtol=0, atol=5e-4
```
langong347, line 36: nit: why specifically `5e-4`? How about `1e-4` or `1e-3`?

examples/mugen/retrieval/video_clip.py
```diff
@@ -0,0 +1,31 @@
...
+    Args:
+        trainable (bool): true if model weights should be trained
+        preprocess_mean (list): sequence of means to normalize each channel to
+        preprocess_std (list): sequence of standard deviations to normalize each channel to
+
+    Inputs:
+        x (Tensor): batch of videos with dimensions (batch, channel, time, height, width)
+    """
+
+    def __init__(self):
```
langong347, line 24: you will still need to add transform later on?

examples/mugen/retrieval/s3d.py
ebsmothers: Possibly a noob suggestion here, but all these `Mixed*` classes have very similar structure. Why not create a class per branch (except maybe `branch0` since it's just a single op) that you can pass the primitive params (in/out dims, kernel_size, stride) to, then write a single class to handle the `torch.cat` portion of things? Then you don't have to write the same 4 sequentials a ton of times

examples/mugen/retrieval/video_clip.py
```diff
@@ -0,0 +1,31 @@
...
+        super().__init__()
+        self.model = S3D(400)
+        self.embedding_dim = self.model.fc.in_channels
```
ebsmothers, line 27: Where does this get used?

examples/mugen/retrieval/video_clip.py
sophiazhi: Resolving due to our offline discussion + [this thread](https://github.com/facebookresearch/multimodal/pull/112#discussion_r906439176) deciding to keep the transform separate

examples/mugen/retrieval/s3d.py
sophiazhi: Lan and I discussed not generalizing the S3D architecture yet, since that process requires more time to understand the existing torchvision VideoResnet code and generalize S3D in a consistent way (reusing torchvision classes, generalizing torchvision classes, following similar submodule structure). I like your suggestions but for now we didn’t want to generalize this S3D code if it has to be changed later anyway to a possibly different structure.

langong347 (approved):

examples/mugen/retrieval/s3d.py
ebsmothers: I am not necessarily thinking about how to generalize the S3D architecture as a whole, more about how to reduce the copypasta in all the `Mixed*` classes. Seems they could all come from a common class with a different set of params passed, no?

sophiazhi: Sure I'll try that out in a new commit

PR #111, opened 2022-06-23
Video transform and test

sophiazhi:
Summary:
Add transform class for videos, as part of onboarding S3D video encoder.
- `multimodal/transforms/video_transform.py`: video preprocessing 
- `test/transforms/test_video_transform.py`: video preprocessing tests

Test plan:
Run `python -m pytest test/transforms/test_video_transform.py::TestVideoTransform -rP`


torchmultimodal/transforms/video_transform.py
```diff
@@ -0,0 +1,94 @@
...
+DEFAULT_NORMALIZE_STD = [0.22803, 0.22145, 0.216989]
+DEFAULT_RESIZE_SHAPE = (224, 224)
+
+
+class VideoTransform:
+    """Transform videos for encoding
+
+    Args:
+        time_samples (int): number of frames to sample in the time dimension
+        mean (list): sequence of means of each channel
```
langong347, line 24: nit: `List[float]`

torchmultimodal/transforms/video_transform.py
```diff
@@ -0,0 +1,94 @@
...
+        std (list): sequence of standard deviations of each channel
+        resize_shape (tuple): shape to resize each frame to
```
langong347, line 26: `Tuple[int]`

torchmultimodal/transforms/video_transform.py
```diff
@@ -0,0 +1,94 @@
...
+
+    Inputs:
+        video (Tensor): batch of videos with dimensions (batch, time, height, width, channel)
+
+    Outputs:
```
langong347, line 31: `Returns`
https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html

langong347 (approved):
A few more `nit` comments. Otherwise I think it's good to go.

torchmultimodal/transforms/video_transform.py
```diff
@@ -0,0 +1,96 @@
...
+        return video
+
+    def normalize(self, video: torch.Tensor) -> torch.Tensor:
+        """Normalizes video of dims (b, t, h, w, c) to mean 0, std 1"""
+        b, t, _, _, _ = video.shape
+        video = video.flatten(start_dim=0, end_dim=1)  # "b t h w c -> (b t) h w c"
+        video = shift_dim(video, -1, 1)  # "(b t) h w c -> (b t) c h w"
+
+        video = video.float() / 255.0
+        video = normalize(video, mean=self.mean, std=self.std)
```
ebsmothers, line 92: nit: either import this as a different name or rename the method, otherwise it can be confusing

torchmultimodal/transforms/video_transform.py
```diff
@@ -0,0 +1,96 @@
...
+                shift_dim(video, -1, 1), size=[self.time_samples, h, w]
+            )  # "b t h w c -> b c t h w"
+            video = shift_dim(video, 1, -1)  # "b c t h w -> b t h w c"
+        return video
+
+    def resize_hw(self, video: torch.Tensor) -> torch.Tensor:
+        """Resizes height and width of video of dims (b, t, h, w, c)"""
+        b, t, h, w, _ = video.shape
+        video = video.flatten(start_dim=0, end_dim=1)  # "b t h w c -> (b t) h w c"
+        video = shift_dim(video, -1, 1)  # "(b t) h w c -> (b t) c h w"
```
ebsmothers, line 75: Maybe not specific to this PR, but what is the purpose of shift_dim in general? Why not just use permute?

torchmultimodal/transforms/video_transform.py
```diff
@@ -0,0 +1,96 @@
...
+            video = F.interpolate(
+                shift_dim(video, -1, 1), size=[self.time_samples, h, w]
+            )  # "b t h w c -> b c t h w"
+            video = shift_dim(video, 1, -1)  # "b c t h w -> b t h w c"
+        return video
+
+    def resize_hw(self, video: torch.Tensor) -> torch.Tensor:
+        """Resizes height and width of video of dims (b, t, h, w, c)"""
+        b, t, h, w, _ = video.shape
+        video = video.flatten(start_dim=0, end_dim=1)  # "b t h w c -> (b t) h w c"
```
ebsmothers, line 74: Given that you are already getting the shape in this method, might be best to just reshape explicitly using those values. Then you don't need the comment. Plus I think flatten will make a copy, whereas reshape will at least try to do a view (unless you want a copy?)

torchmultimodal/transforms/video_transform.py
sophiazhi: `shift_dim` is more concise, but I'm not sure if there's another reason why it exists

PR #112, opened 2022-06-23
Support DistilBERT-Backed Text Encoder

sophiazhi:
Summary:
Add DistilBERT-backed encoder as the text encoder for MUGEN video-text retrieval.

- `mugen/retrieval/video_clip.py`: TextEncoder
- `mugen/test/retrieval/test_video_clip.py`: TestTextEncoder
- `torchmultimodal/transforms/bert_text_transform.py`: BertTextTransform
- `test/transforms/test_bert_text_transform.py`: TestBertTextTransform
- `mugen/requirements.txt`: add huggingface dependency

Test plan:
For a coverage report of videoclip, temporarily remove the "examples/" lines from .coveragerc, then run `python -m pytest --cov=examples/mugen/retrieval examples/mugen/test/retrieval/test_video_clip.py -vv`
For a coverage report of bert text transform, run `python -m pytest --cov=torchmultimodal/transforms test/transforms/test_bert_text_transform.py -vv`
Screen Shot 2022-06-27 at 11 33 45 AM
Screen Shot 2022-06-27 at 11 30 58 AM


examples/mugen/test/retrieval/test_video_clip.py
```diff
@@ -0,0 +1,29 @@
...
+# LICENSE file in the root directory of this source tree.
+
+import pytest
+import torch
+from examples.mugen.retrieval.video_clip import TextEncoder
+
+from test.test_utils import assert_expected, set_rng_seed
+
+
+class TestTextEncoder:
```
langong347, line 14: For simplicity we can write this as a function since it's just one test.

examples/mugen/retrieval/video_clip.py
```diff
@@ -0,0 +1,46 @@
...
+        self,
+        model_name="distilbert-base-uncased",
+        pretrained=True,
+        max_length=200,
+    ):
+        super().__init__()
+        self.max_length = max_length
+        self.tokenizer = DistilBertTokenizer.from_pretrained(model_name)
+        if pretrained:
+            self.model = DistilBertModel.from_pretrained(model_name)
```
langong347, line 30: non blocking comment: We should align with torchtext on the timeline of landing this model in parallel.

examples/mugen/retrieval/video_clip.py
```diff
@@ -0,0 +1,46 @@
...
+
+    def __init__(
+        self,
+        model_name="distilbert-base-uncased",
+        pretrained=True,
+        max_length=200,
+    ):
+        super().__init__()
+        self.max_length = max_length
+        self.tokenizer = DistilBertTokenizer.from_pretrained(model_name)
```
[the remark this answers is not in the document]
ebsmothers: Unrelated comment on this line: do we need to have the tokenizer inside of the encoder? Personally I would bias towards keeping it separate to make things more modular

examples/mugen/retrieval/video_clip.py
```diff
@@ -0,0 +1,46 @@
...
+        self.target_token_idx = 0
+
+    def forward(self, raw_text):
+        device = get_current_device()
+        batch_encoding = self.tokenizer(
+            raw_text, padding=True, truncation=True, max_length=self.max_length
+        )
+        input_ids = torch.tensor(batch_encoding["input_ids"]).to(device)
+        attention_mask = torch.tensor(batch_encoding["attention_mask"]).to(device)
+        output = self.model(input_ids=input_ids, attention_mask=attention_mask)
```
ebsmothers, line 44: Not sure if the tokenizer is returning other keys besides `input_ids` and `attention_mask`, but if not a more condensed way to do this may be 
```suggestion
        output = self.model(**batch_encoding)
```

examples/mugen/retrieval/video_clip.py
```diff
@@ -0,0 +1,46 @@
...
+        pretrained=True,
+        max_length=200,
+    ):
+        super().__init__()
+        self.max_length = max_length
+        self.tokenizer = DistilBertTokenizer.from_pretrained(model_name)
+        if pretrained:
+            self.model = DistilBertModel.from_pretrained(model_name)
+        else:
+            self.model = DistilBertModel(config=DistilBertConfig())
```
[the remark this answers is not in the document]
sophiazhi: I can have the constructor accept a `model_config` dict and then pass in `DistilBertConfig(**model_config)` here

examples/mugen/retrieval/video_clip.py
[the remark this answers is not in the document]
sophiazhi: Looks like the convention among [openAI CLIP](https://github.com/openai/CLIP/blob/c0065a27ad170a67d9572f693292b594982ab9b2/README.md#L36), [torchvision video classifiers](https://github.com/pytorch/vision/blob/main/references/video_classification/train.py), and our own CLIPTextEncoder is to put tokenizing/preprocessing outside of the model, so I'll separate it from both the TextEncoder and VideoEncoder for mugen. thanks for the suggestion!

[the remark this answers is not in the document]
parmeet: > TorchText has just landed [BERTTokenizer](https://github.com/pytorch/text/blob/main/torchtext/transforms.py#L539) which [should be the same as DistilBERTTokenizer](https://huggingface.co/docs/transformers/model_doc/distilbert#transformers.DistilBertTokenizer). Can we take a look?


Yupp, that's a great suggestion @langong347 . Let me know if I can help with any questions here :)

examples/mugen/retrieval/video_clip.py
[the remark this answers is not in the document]
ebsmothers: I would bias more towards giving access to specific parameters over configs since blindly passing configs around can make things less interpretable. But I understand that the config approach is better aligned with this HF model. So ultimately I think it's up to you which you prefer 🙂

examples/mugen/retrieval/video_clip.py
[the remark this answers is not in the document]
parmeet: Note that if we are to you torchtext's BERT tokenizer, it returns token ids as list[list[str]], so we need to do few steps in order to match the output from HF tokenizer:
1) Convert str IDs into int IDs (you may use [StrToIntTransform)](https://github.com/pytorch/text/blob/31eabf8afe6449842fc88e934c55523230f02fb0/torchtext/transforms.py#L260)
2) Pad tokens and create tensor object. You may use [ToTensor](https://github.com/pytorch/text/blob/31eabf8afe6449842fc88e934c55523230f02fb0/torchtext/transforms.py#L115) transform
3) Create attention mask by comparing against the padding index in token ids.

Hope this helps!

examples/mugen/retrieval/video_clip.py
sophiazhi: resolving since the tokenizer has been moved outside of the encoder

examples/mugen/test/retrieval/test_video_clip.py
sophiazhi: I added a test for the attention mask, but if there was still only one test -- I prefer to group the text encoder test in its own class because this file will later also contain tests for VideoEncoder and the model builder, and it's a nice way to organize the text-specific fixtures and potential future text encoder tests. It also follows the pattern set by other test files in our repo (although having test classes might just be a unittest requirement).

examples/mugen/retrieval/video_clip.py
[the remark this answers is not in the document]
sophiazhi: I would prefer the config approach since none of the parameters strike me as more important than the others and there are so many, but maybe having the user pass in a DistilBertConfig object (rather than a dict) adds some transparency to what params are configurable?

examples/mugen/test/retrieval/test_video_clip.py
```diff
@@ -12,18 +12,47 @@
...
+        )
 
-    def test_forward(self, start):
-        encoder, text_input = start
-        out = encoder(text_input)
-        expected_sum = -5.8399
-        assert_expected(actual=out.shape, expected=torch.Size([1, 768]), rtol=0, atol=0)
+    def test_forward_untrained(self, utils):
+        input_ids = utils
+        encoder = TextEncoder(pretrained=False, model_config={})
```
langong347, line 41: Should we just set `model_config` as `None` unless you meant to override certain default configs?

examples/mugen/retrieval/video_clip.py
```diff
@@ -6,41 +6,54 @@
...
+
+    Inputs:
+        input_ids (Tensor): tensor of (batch, text_length) tokenized text
+
+    Returns:
+        Tensor: encoded text
+
     """
 
     def __init__(
```
langong347, line 32: Since you are adding type hints for the methods, should we also add them to the constructor?

examples/mugen/retrieval/video_clip.py
```diff
@@ -6,41 +6,54 @@
...
+        pretrained (bool): whether to use a pretrained model or not
+        model_name (str): name of pretrained model, used when pretrained is True
+        model_config (dict): model config for DistilBERT, used when pretrained is False
+        padding_value (int): value that was used to pad the input text
+
+    Inputs:
+        input_ids (Tensor): tensor of (batch, text_length) tokenized text
+
+    Returns:
+        Tensor: encoded text
```
langong347, line 28: Add dimension here too? Would also elaborate on the purpose of indexing the last hidden states with the `CLS token` even though it's probably a well known practice for BERT model.

examples/mugen/retrieval/video_clip.py
```diff
@@ -6,41 +6,54 @@
...
 import torch
 from torch import nn
-from torchmultimodal.utils.common import get_current_device
-from transformers import DistilBertConfig, DistilBertModel, DistilBertTokenizer
+from transformers import DistilBertConfig, DistilBertModel
 
 
 class TextEncoder(nn.Module):
-    """Encode text to a fixed-length vector
+    """Encode tokenized text to a fixed-length vector
```
langong347, line 13: What are you referring to by "fixed length vector"? 

- For seq2seq model, "fixed length vector" refers to a fixed intermediate vector ("context vector") to encoder a sentence. 
- Transformer-based models like BERT learns the pair-wise correlation between the source and the target tokens and then uses the learnt correlation for prediction given a new sequence of inputs. I don't think there's a corresponding "fixed vector" in this case as the correlation is calculated amongst all pairs of source/target tokens.

torchmultimodal/transforms/bert_text_transform.py
```diff
@@ -0,0 +1,40 @@
...
+from torchtext.transforms import (
+    AddToken,
+    BERTTokenizer,
+    Sequential,
+    StrToIntTransform,
+    ToTensor,
+)
+
+
+class BertTextTransform:
```
langong347, line 18: Add docstring to the class.

torchmultimodal/transforms/bert_text_transform.py
```diff
@@ -0,0 +1,40 @@
...
+    def __init__(
+        self,
+        vocab_file: str = "https://huggingface.co/distilbert-base-uncased/resolve/main/vocab.txt",
+        do_lower_case: bool = True,
```
langong347, line 22: Where do those defaults come from? And what do they mean? Explain in the docstring?

examples/mugen/retrieval/video_clip.py
sophiazhi: edited the docstring summary to explain the CLS token and that the encoding is the last hidden state of the CLS token

examples/mugen/retrieval/video_clip.py
```diff
@@ -5,36 +5,42 @@
...
 
     def __init__(
         self,
-        pretrained=True,
-        model_name="distilbert-base-uncased",
-        model_config=None,
-        padding_value=0,
+        pretrained: bool = True,
+        model_name: str = "distilbert-base-uncased",
+        model_config: dict = None,
```
langong347, line 42: When the default is `None`, wrap it with `Optional`, e.g., `Optional[Dict[str, Any]]` or if the values are of a specific type, use that.

examples/mugen/retrieval/video_clip.py
```diff
@@ -5,36 +5,42 @@
...
+        padding_value (int): value that was used to pad the input text.
+            Defaults to 0, Hugging Face's BERT pad token.
 
     Inputs:
         input_ids (Tensor): tensor of (batch, text_length) tokenized text
 
     Returns:
-        Tensor: encoded text
-
+        Tensor: encoded text with dimensions (batch, 768)
```
langong347, line 35: `768` is just the [default of huggingface config](https://huggingface.co/docs/transformers/model_doc/distilbert#transformers.DistilBertConfig.dim) which could be different from a customized `model_config` or when `pretrained=False`. Maybe `model_config.dim` or a name that is more generic?

PR #113, opened 2022-06-23
[Lint] Fix mypy missing annotations

RdoubleA:
## Summary
Some files have been excluded from the `mypy.ini` config, so I've added these back as well as the line `disallow_untyped_defs = True` to find any outstanding mypy issues in the codebase. 

Some functions I had to make some modifications to fit mypy so take a look to see if it's still the same. I've also moved flava builders to the bottom because it fixed some mypy issues...

## Test plan
`pytest -vv` to ensure nothing broke


ebsmothers (commented):
So outside of FLAVA everything looks good. For flava.py you should also coordinate with @ankitade since she is moving a lot of stuff around in #115. The diff rendering here seems messed up, makes it look like you're changing a bunch of the DALL-E code. As far as I can tell you're just moving the builders to the end, anything else I'm missing here?

torchmultimodal/utils/common.py
```diff
@@ -127,20 +127,20 @@ def load_model(
 
 
 class ModelOutput(OrderedDict):
-    def keys(self):
+    def keys(self) -> Any:
```
langong347, line 130: Should we use `Generator` as the type hint for `yield`? 

https://docs.python.org/3/library/typing.html#typing.Generator

ankitade (commented):
where are the mypy.ini changes?

torchmultimodal/utils/common.py
ebsmothers: Ultimately did not adopt this suggestion as ModelOutput inherits from OrderedDict which has its own type for this. [Ref](https://gist.github.com/joequery/12332f410a05e6c7c949#file-ordered_dict-py-L175)

langong347: The generator is higher than the keys right? I.e., we are returning a generator instead of the keys themselves.

torchmultimodal/models/flava/flava_image_encoder.py
```diff
@@ -21,9 +20,7 @@
 from torchmultimodal.modules.losses.flava import Pooler
 
 
-def to_2tuple(x):
-    if isinstance(x, collections.abc.Iterable):
-        return x
+def to_2tuple(x: int) -> Tuple[int, int]:
```
RdoubleA, line 23: do we even need this function? it's used only twice and can be easily replaced with `image_size = (image_size, image_size)`

torchmultimodal/modules/layers/attention.py
```diff
@@ -85,7 +85,12 @@ def _combine_multihead(self, x: Tensor) -> Tensor:
         return shift_dim(x, 1, -2).flatten(start_dim=-2)
 
     def forward(
-        self, q: Tensor, k: Tensor, v: Tensor, decode_step=None, decode_idx=None
+        self,
+        q: Tensor,
+        k: Tensor,
+        v: Tensor,
+        decode_step: int = None,
+        decode_idx: Iterable[int] = None,
```
RdoubleA, line 93: I would do this: 
```
decode_step: Optional[int] = None,
decode_idx: Optional[List[int]] = None,
```
this is actually a change I've made in a separate PR: https://github.com/facebookresearch/multimodal/pull/101/files

torchmultimodal/utils/common.py
RdoubleA: I've tried setting the return type to Generator but mypy gets quite angry:
```
torchmultimodal/utils/common.py:168:5: error: Return type "Generator[Any, Any, Any]" of "keys" incompatible with return type
"_OrderedDictKeysView[Any, Any]" in supertype "OrderedDict"  [override]
        def keys(self) -> Generator:
        ^
torchmultimodal/utils/common.py:168:5: error: Return type "Generator[Any, Any, Any]" of "keys" incompatible with return type
"dict_keys[Any, Any]" in supertype "dict"  [override]
        def keys(self) -> Generator:
        ^
torchmultimodal/utils/common.py:168:5: error: Return type "Generator[Any, Any, Any]" of "keys" incompatible with return type
"KeysView[Any]" in supertype "Mapping"  [override]
        def keys(self) -> Generator:
```
I think setting it to Any is the simplest although not ideal solution

torchmultimodal/models/flava/flava_image_encoder.py
ankitade: yes get rid of it in another pr

torchmultimodal/modules/layers/transformer.py
```diff
@@ -115,10 +121,10 @@ def __init__(
 
     def forward(
         self,
-        hidden_states,
-        attention_mask=None,
-        head_mask=None,
-    ):
+        hidden_states: Tensor,
+        attention_mask: Tensor = None,
```
ankitade, line 125: should be optional no? why isnt it getting flagged tho?

torchmultimodal/modules/losses/contrastive_loss_with_temperature.py
```diff
@@ -181,7 +181,7 @@ def forward(
         image_embeddings: torch.Tensor,
         text_embeddings: torch.Tensor,
         backprop_in_gather: bool = True,
-    ):
+    ) -> torch.Tensor:
```
ankitade, line 184: do u want to make it tensor like everywhere else?

torchmultimodal/modules/losses/flava.py
```diff
@@ -438,7 +442,7 @@ def forward(
                 else -(text_masked_sequence.size(1) - 1)
             )
             sequence_for_text = multimodal_masked_sequence[:, start_index:, :]
-            outputs.mmm_text_output = self.mmm_loss.mlm(
+            outputs.mmm_text_output = self.mmm_loss["mlm"](
```
ankitade, line 445: why?

torchmultimodal/models/flava/flava_model.py
```diff
@@ -636,24 +631,32 @@ def forward(self, x: torch.Tensor) -> torch.Tensor:
...
+        else:
+            self.encoder.load_state_dict(encoder.state_dict())
         return self.state_dict()
 
     def get_codebook_indices(self, images: Tensor) -> Tensor:
         z_logits = self.encoder(images)
-        return torch.argmax(z_logits, axis=1)
+        return torch.argmax(z_logits, dim=1)
 
     def get_codebook_probs(self, images: Tensor) -> Tensor:
```
ankitade, line 661: why are we touching this logic

ankitade (commented):
thanks, appreciate the hard work. Next time, please break up such PRs  into smaller ones so its easier for reviewer to check if core logic is being touched or its just typing changes.

torchmultimodal/models/flava/flava_image_encoder.py
```diff
@@ -34,22 +31,30 @@ class PatchEmbeddings(nn.Module):
...
         super().__init__()
         image_size = to_2tuple(image_size)
         patch_size = to_2tuple(patch_size)
         num_patches = (image_size[1] // patch_size[1]) * (
             image_size[0] // patch_size[0]
         )
-        self.image_size = image_size
-        self.patch_size = patch_size
+        self.image_size: Tuple[int, int] = (image_size[0], image_size[1])
+        self.patch_size: Tuple[int, int] = (patch_size[0], patch_size[1])
```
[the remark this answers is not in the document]
ebsmothers: This is the replacement for to_2tuple. I will take it out for now if we're replacing that in another PR

torchmultimodal/models/flava/flava_model.py
```diff
@@ -382,12 +233,107 @@ def encode_mm(
...
+
+
 class FLAVAForPreTraining(nn.Module, PretrainedMixin):
     # TODOs:
     # 1. Expose logit scale
     # 2. For FLAVA model, allow interpolating the embeddings to
     # for patch embeddings
-    def __init__(self, model: FLAVAModel, image_codebook: nn.Module, loss: nn.Module):
+    def __init__(
+        self, model: FLAVAModel, image_codebook: nn.Module, loss: FLAVAPretrainingLoss
```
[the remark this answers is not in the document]
ebsmothers: Not sure what the right thing to do here is exactly. The args passed to loss are pretty specific, so I doubt we will be using anything besides FLAVAPretrainingLoss. But I'm fine to keep the type a bit more general

torchmultimodal/modules/layers/transformer.py
ebsmothers: There is a [flag](https://mypy.readthedocs.io/en/stable/command_line.html#cmdoption-mypy-no-strict-optional) for this that we have set right now. Turning it off triggers a ton of errors so I'm gonna leave it as is for now. We can handle that in a follow-up if we want

torchmultimodal/modules/losses/flava.py
ebsmothers: When I use dot access mypy thinks it's a tensor

torchmultimodal/modules/layers/transformer.py
RdoubleA: oh no there's more? :(

torchmultimodal/models/flava/flava_model.py
RdoubleA: mypy gave an error on this, and the [documentation](https://pytorch.org/docs/stable/generated/torch.argmax.html) also uses `dim` instead of `axis`

RdoubleA:
@ebsmothers there is one last mypy error in mdetr in these lines:
```
if embedding_dim is None:
        embedding_dim = text_encoder.embedding_dim
```
not sure the best way to fix this

torchmultimodal/modules/losses/flava.py
ankitade: just put ignore for now and add that comment

ankitade (approved):

PR #116, opened 2022-06-24
[ALBEF] multimodal encoder

katrina433:
Summary:

This PR implements the multimodal encoder for the ALBEF model. 
* Modifies the `ALBEFTransformerAttention` class in [albef_text_encoder.py](https://github.com/facebookresearch/multimodal/compare/albef-multimodal-encoder?expand=1#diff-3932bbbc9188801f7a2310b6c7d9461bf744f9ac5be55ab613ded71f83a062d6) to include an optional `encoder_hidden_states` input. When this input is passed in, cross attention is computed. Otherwise self attention is computed. 
* The multimodal encoder is similar to the text encoder, with the cross attention added to each of the encoder layers. 
* Fixed some bugs in `ALBEFModel` found when adding embedding tests to the [test_albef.py](https://github.com/facebookresearch/multimodal/compare/albef-multimodal-encoder?expand=1#diff-d5a1ba66dc8cd1d757cd40b8676f6c2a07866a432fae6f94f5f61a12561a1e8f) (removed `text_atts` from `ALBEFOutput` because it's not used, buffer queues were updated in the wrong place, and `vl_embeddings` only needed the cls tokens. 

Test plan:

* Wrote unit tests for the multimodal encoder. 
* Added more tests for the ALBEFModel to test for expected encoder embeddings

### DIFF
--- a/test/models/test_albef.py
+++ b/test/models/test_albef.py
@@ -145,13 +145,15 @@ def test_dequeue_and_enqueue(dummy_albef_model):
 def test_momentum_update(dummy_albef_model):
     init_weight = Tensor([[1, 2, 3], [4, 5, 6]])
     init_weight_m = Tensor([[6, 5, 4], [3, 2, 1]])
-    dummy_albef_model.models[0].weight = nn.Parameter(init_weight)
-    dummy_albef_model.models_m[0].weight = nn.Parameter(init_weight_m)
+    dummy_albef_model.vision_encoder.weight = nn.Parameter(init_weight)
+    dummy_albef_model.vision_encoder_m.weight = nn.Parameter(init_weight_m)
     dummy_albef_model._momentum_update()
     expected_weight_m = Tensor([[4.75, 4.25, 3.75], [3.25, 2.75, 2.25]])
-    assert_expected(dummy_albef_model.models[0].weight, init_weight, rtol=0, atol=1e-4)
     assert_expected(
-        dummy_albef_model.models_m[0].weight, expected_weight_m, rtol=0, atol=1e-4
+        dummy_albef_model.vision_encoder.weight, init_weight, rtol=0, atol=1e-4
+    )
+    assert_expected(
+        dummy_albef_model.vision_encoder_m.weight, expected_weight_m, rtol=0, atol=1e-4
     )
 
 
--- a/test/models/test_albef.py
+++ b/test/models/test_albef.py
@@ -52,6 +52,7 @@ def dummy_albef_model():
         nn.Linear(3, 2),
         embed_dim=2,
         queue_size=4,
+        momentum=0.75,
     )
 
 
@@ -147,7 +148,7 @@ def test_momentum_update(dummy_albef_model):
     dummy_albef_model.models[0].weight = nn.Parameter(init_weight)
     dummy_albef_model.models_m[0].weight = nn.Parameter(init_weight_m)
     dummy_albef_model._momentum_update()
-    expected_weight_m = Tensor([[5.9750, 4.9850, 3.9950], [3.0050, 2.0150, 1.0250]])
+    expected_weight_m = Tensor([[4.75, 4.25, 3.75], [3.25, 2.75, 2.25]])
     assert_expected(dummy_albef_model.models[0].weight, init_weight, rtol=0, atol=1e-4)
     assert_expected(
         dummy_albef_model.models_m[0].weight, expected_weight_m, rtol=0, atol=1e-4
--- a/test/models/test_albef.py
+++ b/test/models/test_albef.py
@@ -4,14 +4,47 @@
 # This source code is licensed under the BSD-style license found in the
 # LICENSE file in the root directory of this source tree.
 
+from functools import partial
+
+import pytest
 import torch
 from test.test_utils import assert_expected, set_rng_seed
 from torch import nn, Tensor
 from torchmultimodal.models.albef import ALBEFModel, ALBEFSimilarity
+from torchmultimodal.modules.encoders.albef_multimodal_encoder import (
+    ALBEFMultimodalEncoder,
+)
+from torchmultimodal.modules.encoders.albef_text_encoder import ALBEFTextEncoder
+from torchmultimodal.modules.encoders.albef_vision_encoder import ALBEFVisionEncoder
 
 
-class TestALBEFModel:
-    albef = ALBEFModel(
+@pytest.fixture(autouse=True)
+def vision_encoder():
+    set_rng_seed(0)
+    return ALBEFVisionEncoder(
+        image_size=4,
+        patch_size=4,
+        num_layers=2,
+        num_heads=1,
+        hidden_dim=3,
+        mlp_dim=6,
+        norm_layer=partial(nn.LayerNorm, eps=1e-6),
+    )
+
+
+@pytest.fixture(autouse=True)
+def text_encoder():
+    return ALBEFTextEncoder(hidden_size=3, num_attention_heads=1)
+
+
+@pytest.fixture(autouse=True)
+def multimodal_encoder():
+    return ALBEFMultimodalEncoder(hidden_size=3, num_attention_heads=1)
+
+
+@pytest.fixture(autouse=True)
+def dummy_albef_model():
+    return ALBEFModel(
         nn.Linear(3, 2),
         nn.Linear(3, 2),
         nn.Linear(3, 2),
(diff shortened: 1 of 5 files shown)


test/models/test_albef.py
```diff
@@ -4,113 +4,220 @@
...
-            [[0.568431, -1.084522, -1.398595], [1.540996, -0.293429, -2.178789]]
-        ).unsqueeze(1)
-        expected_text_embeds_neg = Tensor(
-            [[-0.403344, -0.596635, 0.182036], [0.403347, 0.838026, -0.719258]]
-        ).unsqueeze(1)
-        expected_text_atts_neg = Tensor([1.100604, -0.856675]).unsqueeze(1)
-        assert_expected(image_embeds_neg, expected_image_embeds_neg, rtol=0, atol=1e-4)
-        assert_expected(text_embeds_neg, expected_text_embeds_neg, rtol=0, atol=1e-4)
-        assert_expected(text_atts_neg, expected_text_atts_neg, rtol=0, atol=1e-4)
+
```
ebsmothers, line 70: Way to write such a thorough test for the end-to-end ALBEF model! Glad to see this helped you to find a couple subtle things we missed on the first pass

test/models/test_albef.py
```diff
@@ -4,113 +4,220 @@
...
+    assert_expected(
+        dummy_albef_model.text_queue[:, 0:2], text_feat_m.T, rtol=0, atol=1e-4
+    )
+
+
+def test_momentum_update(dummy_albef_model):
+    init_weight = Tensor([[1, 2, 3], [4, 5, 6]])
+    init_weight_m = Tensor([[6, 5, 4], [3, 2, 1]])
+    dummy_albef_model.models[0].weight = nn.Parameter(init_weight)
+    dummy_albef_model.models_m[0].weight = nn.Parameter(init_weight_m)
```
ebsmothers, line 148: Kind of a nit, but this indexing to get the vision encoder out of the ALBEFModel class makes me think that a list is probably not the best data structure for `self.models` and `self.models_m`...

test/models/test_albef.py
```diff
@@ -4,113 +4,220 @@
...
         nn.Linear(3, 2),
         nn.Linear(3, 2),
         nn.Linear(3, 2),
+        embed_dim=2,
+        queue_size=4,
+    )
+
+
+@pytest.fixture(autouse=True)
+def albef_model(vision_encoder, text_encoder, multimodal_encoder):
```
ebsmothers, line 59: I thought autouse meant you don't have to add any of these as args? (Sorry, this may just be my noob pytest understanding...)

test/models/test_albef.py
```diff
@@ -4,113 +4,220 @@
...
-        expected_text_embeds_neg = Tensor(
-            [[-0.403344, -0.596635, 0.182036], [0.403347, 0.838026, -0.719258]]
-        ).unsqueeze(1)
-        expected_text_atts_neg = Tensor([1.100604, -0.856675]).unsqueeze(1)
-        assert_expected(image_embeds_neg, expected_image_embeds_neg, rtol=0, atol=1e-4)
-        assert_expected(text_embeds_neg, expected_text_embeds_neg, rtol=0, atol=1e-4)
-        assert_expected(text_atts_neg, expected_text_atts_neg, rtol=0, atol=1e-4)
+
+@pytest.fixture(autouse=True)
+def albef_model_output(albef_model):
```
ebsmothers, line 72: Similar comment for autouse here, and on the usage of `albef_model_output` and `dummy_albef_model` below

test/models/test_albef.py
```diff
@@ -4,113 +4,220 @@
...
+    )
+
+
+def test_momentum_update(dummy_albef_model):
+    init_weight = Tensor([[1, 2, 3], [4, 5, 6]])
+    init_weight_m = Tensor([[6, 5, 4], [3, 2, 1]])
+    dummy_albef_model.models[0].weight = nn.Parameter(init_weight)
+    dummy_albef_model.models_m[0].weight = nn.Parameter(init_weight_m)
+    dummy_albef_model._momentum_update()
+    expected_weight_m = Tensor([[5.9750, 4.9850, 3.9950], [3.0050, 2.0150, 1.0250]])
```
ebsmothers, line 150: I like this test a lot because it shows how the momentum update works. One nit suggestion is to set the momentum value in your dummy ALBEF model to a nice number like 0.75 to make it even more obvious what's happening here. (Just don't set it to 0.5 🙂)

torchmultimodal/modules/encoders/albef_multimodal_encoder.py
```diff
@@ -0,0 +1,106 @@
...
+    ) -> None:
+        super().__init__()
+        self.attention = ALBEFTransformerAttention(
+            hidden_size, num_attention_heads, layer_norm_eps
+        )
+        self.cross_attention = ALBEFTransformerAttention(
+            hidden_size, num_attention_heads, layer_norm_eps
+        )
+        self.dense1 = nn.Linear(hidden_size, intermediate_size)
+        self.transform_act_fn = nn.GELU()
```
ebsmothers, line 88: You can consider adding this as a param, not super important though

torchmultimodal/modules/encoders/albef_multimodal_encoder.py
```diff
@@ -0,0 +1,106 @@
...
+from torch import nn, Tensor
+from torchmultimodal.modules.encoders.albef_text_encoder import (
+    ALBEFTransformerAttention,
+)
+from torchmultimodal.utils.common import get_extended_attention_mask
+
+
+class ALBEFMultimodalEncoder(nn.Module):
+    """
+    Construct multimodal embeddings from image embeddings, text embeddings, and text attention mask.
```
ebsmothers, line 17: Maybe add a bit more detail here explaining how it compares to the text encoder and the use of cross-attention

torchmultimodal/modules/encoders/albef_text_encoder.py
```diff
@@ -226,10 +230,17 @@ def forward(
         self,
         hidden_states: Tensor,
         attention_mask: Tensor,
+        encoder_hidden_states: Optional[Tensor] = None,
     ) -> Tensor:
         mixed_query_layer = self.query(hidden_states)
-        mixed_key_layer = self.key(hidden_states)
-        mixed_value_layer = self.value(hidden_states)
+
+        if encoder_hidden_states is None:
```
ebsmothers, line 237: Add a comment explaining what's going on here

torchmultimodal/modules/encoders/albef_multimodal_encoder.py
```diff
@@ -0,0 +1,106 @@
...
+        num_attention_heads: int,
+        layer_norm_eps: float,
+    ) -> None:
+        super().__init__()
+        self.attention = ALBEFTransformerAttention(
+            hidden_size, num_attention_heads, layer_norm_eps
+        )
+        self.cross_attention = ALBEFTransformerAttention(
+            hidden_size, num_attention_heads, layer_norm_eps
+        )
```
ebsmothers, line 86: So the combination of attention and cross-attention written the same way can potentially be confusing here (I understand why you're doing it). Either way, might be good to leave a comment explaining what's going on, either here or in the forward.

ebsmothers (commented):
This looks great! I really like the thorough testing you added for the ALBEF model class.

test/models/test_albef.py
katrina433: All models can still be accessed directly (e.g. `dummy_albef_model.vision_encoder`). The two lists are used to iterate through all pairs of `(model, model_m)` (which happens in `_copy_params_momentum_models` and `_momentum_update`). Should I rename them to `_models` and `_models_m`? Or try other data structures (dict `{'vision_encoder': (vision_encoder, vision_encoder_m), ...}`?

test/models/test_albef.py
katrina433: I still need to pass them in as args (or else `albef_model` refers to the function, not what's returned in the function)

YosuaMichael (commented):
Overall looks good, just have some idea on how to refactor AlbefModel and some nit

test/models/test_albef.py
ebsmothers: @katrina433 good point about just accessing via `.vision_encoder`. Personally I would do that here just to be more explicit about what you're modifying. I think @YosuaMichael's suggestion makes sense, feel free to address this in a separate PR though

ebsmothers (approved):
Looks good, thanks for addressing all the comments!

test/models/test_albef.py
```diff
@@ -4,113 +4,221 @@
...
+from torchmultimodal.modules.encoders.albef_multimodal_encoder import (
+    ALBEFMultimodalEncoder,
+)
+from torchmultimodal.modules.encoders.albef_text_encoder import ALBEFTextEncoder
+from torchmultimodal.modules.encoders.albef_vision_encoder import ALBEFVisionEncoder
+
+
+@pytest.fixture(autouse=True)
+def vision_encoder():
+    set_rng_seed(0)
```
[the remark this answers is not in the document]
YosuaMichael: I see, in that case it is probably fine to just leave the `set_rng_seed` inside vision_encoder for now. I am not sure how it work, but I guess probably because it is the first function or the first one being called.

PR #124, opened 2022-06-29
add video-text CLIP model builder

sophiazhi:
Summary:
Add video-text CLIP model builder using MUGEN's architecture.

Test plan:
`python -m pytest examples/mugen/test/retrieval/test_video_clip.py::TestVideoCLIPModel -rP`

### DIFF
--- a/examples/mugen/retrieval/video_clip.py
+++ b/examples/mugen/retrieval/video_clip.py
@@ -7,8 +7,10 @@
 from typing import Any, Dict, Optional
 
 import torch
+import torch.nn.functional as F
 from examples.mugen.retrieval.s3d import S3D
 from torch import nn
+from torchmultimodal.architectures.clip import CLIPArchitecture
 from transformers import DistilBertConfig, DistilBertModel
 
 
@@ -83,3 +85,55 @@ class VideoEncoder(nn.Module):
 
     def forward(self, x):
         return self.model(x)
+
+
+class Projection(nn.Module):
+    """Project embeddings to a fixed dimension.
+
+    Args:
+        dim_in (int): dimension of input
+        dim_out (int): dimension of output
+            Defaults to 256
+        dropout_prob (float): dropout probability
+            Defaults to 0.1
+
+    Inputs:
+        x (Tensor): embeddings (batch, dim_in)
+
+    Returns:
+        Tensor: projected embeddings (batch, dim_out)
+
+    """
+
+    def __init__(self, dim_in, dim_out=256, dropout_prob=0.1) -> None:
+        super().__init__()
+        self.linear1 = nn.Linear(dim_in, dim_out, bias=False)
+        self.linear2 = nn.Linear(dim_out, dim_out, bias=False)
+        self.drop = nn.Dropout(dropout_prob)
+        self.layer_norm = nn.LayerNorm(dim_out)
+
+    def forward(self, x: torch.Tensor) -> torch.Tensor:
+        embed1 = self.linear1(x)
+        embed2 = self.drop(self.linear2(F.gelu(embed1)))
+        embeds = self.layer_norm(embed1 + embed2)
+        return embeds
+
+
+def videoclip(dim_out: int = 256) -> CLIPArchitecture:
+    """Create MUGEN's video-text CLIP model with a S3D-backed video encoder and DistilBERT-backed text encoder.
+        MUGEN paper: https://arxiv.org/abs/2204.08058
+
+    Args:
+        dim_out (int): output dimension to project both encoders' outputs to.
+            Defaults to 256, the value used by MUGEN.
+
+    Returns:
+        CLIPArchitecture: CLIP model using MUGEN's video encoder and text encoder.
+
+    """
+    text_encoder = nn.Sequential(TextEncoder(), Projection(768, dim_out=dim_out))
+    video_encoder = nn.Sequential(
+        VideoEncoder(),
+        Projection(1024, dim_out=dim_out),
+    )
+    return CLIPArchitecture(encoder_a=text_encoder, encoder_b=video_encoder)
--- a/examples/mugen/test/retrieval/test_video_clip.py
+++ b/examples/mugen/test/retrieval/test_video_clip.py
@@ -6,7 +6,12 @@
 
 import pytest
 import torch
-from examples.mugen.retrieval.video_clip import TextEncoder, VideoEncoder
+from examples.mugen.retrieval.video_clip import (
+    Projection,
+    TextEncoder,
+    videoclip,
+    VideoEncoder,
+)
 
 from test.test_utils import assert_expected, set_rng_seed
 
@@ -35,6 +40,7 @@ class TestTextEncoder:
         assert_expected(
             actual=out.sum(), expected=torch.as_tensor(expected_sum), rtol=0, atol=1e-4
         )
+        print(encoder.state_dict().keys())
 
     def test_forward_untrained(self, utils):
         input_ids = utils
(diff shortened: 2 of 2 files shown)


examples/mugen/retrieval/video_clip.py
```diff
@@ -72,14 +81,115 @@ class VideoEncoder(nn.Module):
     """Encode videos to a fixed size vector. Adapted from VideoCLIP
```
langong347, line 81: "Fixed size vector" can be ambiguous. Can we elaborate on what `VideoEncoder` does here?

examples/mugen/retrieval/video_clip.py
```diff
@@ -23,6 +24,8 @@ class TextEncoder(nn.Module):
     Args:
         pretrained (bool): whether to use a pretrained model or not.
             Defaults to True.
+        trainable (bool): whether the encoder's weights should be trainable (not frozen).
+            Defaults to True. Ignored if ``pretrained`` is False.
```
langong347, line 28: Would add a warning statement for "no-surprise" reason, e.g.,  when a user says `pretrained = False` and `trainable = False` but instead the behavior is `trainable = True`.

examples/mugen/retrieval/video_clip.py
```diff
@@ -72,14 +81,115 @@ class VideoEncoder(nn.Module):
...
+    def forward(self, x: torch.Tensor) -> torch.Tensor:
+        embed1 = self.linear1(x)
+        embed2 = self.gelu(embed1)
+        embed2 = self.linear2(embed2)
+        embed2 = self.drop(embed2)
+        embeds = self.layer_norm(embed1 + embed2)
+        return embeds
+
+
+def videoclip(
```
langong347, line 141: `build_videoclip`

examples/mugen/retrieval/video_clip.py
```diff
@@ -72,14 +81,115 @@ class VideoEncoder(nn.Module):
...
+        text_model,
+        Projection(
+            text_model.model.config.dim, out_dim=proj_out_dim, dropout_prob=proj_dropout
+        ),
+    )
+    video_encoder = nn.Sequential(
+        VideoEncoder(
+            trainable=video_trainable,
+        ),
+        Projection(1024, out_dim=proj_out_dim, dropout_prob=proj_dropout),
```
langong347, line 193: Hard coding intermediate dimensions can be error prone as it assumes what's returned by the previous layer. 
Is `1024` the last dim of `VideoEncoder` (the concat dim of [`Mixed5c`](https://github.com/pytorch/vision/blob/6512146e447b69cc9fb379eb05e447a17d7f6d1c/test/test_datasets_utils.py#L20-L29))? 
What if we create an attribute `VideoEncoder.out_dim` or `S3D.out_dim` so if there's any inconsistency the error will be self-contained within the module that causes the error.

examples/mugen/test/retrieval/test_video_clip.py
```diff
@@ -82,3 +100,62 @@ def test_forward(self, start):
...
+        return proj, input
+
+    def test_forward(self, utils):
+        proj, input = utils
+        out = proj(input)
+        expected = torch.Tensor([[-1.2214, -0.0066, 1.2280], [-1.3886, 0.4626, 0.9260]])
+        assert_expected(actual=out, expected=expected, rtol=0, atol=1e-4)
+
+
+class TestVideoCLIPModel:
```
langong347, line 128: nit: `TestVideoCLIPBuilder` ?

examples/mugen/test/retrieval/test_video_clip.py
```diff
@@ -82,3 +100,62 @@ def test_forward(self, start):
...
+                [101, 6315, 3793, 7099, 2005, 5604, 19204, 17629, 102],
+                [101, 2117, 7820, 3793, 102, 0, 0, 0, 0],
+            ]
+        ).to(dtype=int)
+        input_video = torch.randint(10, [2, 3, 32, 32, 32]).float()
+        return input_text, input_video
+
+    def test_forward(self, utils):
+        input_text, input_video = utils
+        clip = videoclip()
```
langong347, line 146: nit: `model` instead of `clip`?

examples/mugen/test/retrieval/test_video_clip.py
```diff
@@ -82,3 +100,62 @@ def test_forward(self, start):
...
+        clip_output = clip(features_a=input_text, features_b=input_video)
+        assert_expected(
+            actual=clip_output.embeddings_a.shape, expected=torch.Size([2, 256])
+        )
+        assert_expected(
+            actual=clip_output.embeddings_b.shape, expected=torch.Size([2, 256])
+        )
+
+    def test_text_dim(self, utils):
```
langong347, line 155: why not adding `test_video_dim`?

langong347 (commented):
I assume this PR will change after #125 . If that's the case, use `ghstack` or make #125 your base commit will include the unmerged changes.

examples/mugen/retrieval/video_clip.py
```diff
@@ -72,14 +81,115 @@ class VideoEncoder(nn.Module):
...
+
     """
 
-    def __init__(self):
+    def __init__(self, trainable: bool = True):
         super().__init__()
         self.model = S3D(400)
         self.model.fc = nn.Identity()
 
+        if not trainable:
```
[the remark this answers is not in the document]
sophiazhi: Right, whichever of #124 and #125 lands first, I'll edit the other one so VideoEncoder has trainable and pretrained args, and line 98 will be ``if pretrained and not trainable``

examples/mugen/retrieval/video_clip.py
sophiazhi: yup, 1024 is the size of the output of Mixed5c. I'll add the out_dim attributes to both encoders

examples/mugen/retrieval/video_clip.py
sophiazhi: "Encode videos to the last layer before the fully-connected layer of S3D"?

examples/mugen/test/retrieval/test_video_clip.py
sophiazhi: I originally added `test_text_dim` because the text encoder output dim is configurable and `build_videoclip` relied on that logic. Now that the output dim logic is contained within the text encoder, I don't think this test case is needed anymore (dimensions are checked in `test_forward` already)

sophiazhi: * to clarify I don't think `test_text_dim` is needed anymore and similarly no `test_video_dim`

Justinyu1618:
lgtm

ebsmothers (changes requested):
How come changes from other PRs are getting picked up here? Maybe you need to rebase your changes on latest main branch

examples/mugen/retrieval/video_clip.py
```diff
@@ -155,13 +169,15 @@ def forward(self, x: torch.Tensor) -> torch.Tensor:
         return embeds
 
 
-def videoclip(
+def build_videoclip(
```
langong347, line 172: Sorry my bad. I checked again that we didn't use `build` prefix in other builders. 

https://github.com/facebookresearch/multimodal/blob/main/torchmultimodal/models/clip.py

examples/mugen/test/retrieval/test_video_clip.py
```diff
@@ -128,7 +129,31 @@ def make_input_video(c_dim=1):
...
+        out = encoder(input_video)
+        expected_sum = 1829.8858
+        assert_expected(
+            actual=out.shape, expected=torch.Size([2, 1024])
+        )  # batch x embedding
+        assert_expected(
+            actual=out.sum(), expected=torch.as_tensor(expected_sum), rtol=0, atol=1e-3
+        )
+        assert next(encoder.parameters()).requires_grad
+        assert encoder.out_dim == 1024
```
langong347, line 147: why need to enforce this value here? are there any subsequent routines that depend on out_dim=1024?

examples/mugen/retrieval/video_clip.py
```diff
@@ -76,36 +87,153 @@ def forward(self, input_ids: torch.Tensor) -> torch.Tensor:
...
+        trainable: bool = True,
         pretrain_path: str = PRETRAINED_S3D_KINETICS400_URL,
     ):
         super().__init__()
         self.model = S3D(400)
+        self.out_dim = self.model.fc.in_channels
         self.model.fc = nn.Identity()
 
         if pretrained:
             self.load_model(pretrain_path)
```
langong347, line 123: nit: Add a prompt here for the user (the same for TextEncoder as for distilledBERT)

https://github.com/mugen-org/MUGEN_baseline/blob/main/lib/models/videoclip/modules.py#L80

examples/mugen/retrieval/video_clip.py
```diff
@@ -76,36 +87,153 @@ def forward(self, input_ids: torch.Tensor) -> torch.Tensor:
...
 
     Inputs:
         x (Tensor): batch of videos with dimensions (batch, channel, time, height, width)
+            Size of ``channel`` dimension must be 3.
+
     """
 
     def __init__(
         self,
         pretrained: bool = True,
```
langong347, line 113: if pretrained is True, the user likely don't want to train the model. It looks over complicated here. Apart from MUGEN, have you seen elsewhere that requires both flags?

examples/mugen/test/retrieval/test_video_clip.py
```diff
@@ -72,8 +77,14 @@ def test_forward_pretrained(self, utils):
         assert_expected(
             actual=out.sum(), expected=torch.as_tensor(expected_sum), rtol=0, atol=1e-4
         )
+        assert next(encoder.parameters()).requires_grad
```
langong347, line 80: Think about whether we indeed need two knobs (pretrained, trainable). Do they really serve the user? If not, we could probably simplify our API as well as reduce complexity in unit tests.

examples/mugen/retrieval/video_clip.py
```diff
@@ -76,36 +87,153 @@ def forward(self, input_ids: torch.Tensor) -> torch.Tensor:
...
+        trainable: bool = True,
         pretrain_path: str = PRETRAINED_S3D_KINETICS400_URL,
     ):
         super().__init__()
         self.model = S3D(400)
+        self.out_dim = self.model.fc.in_channels
         self.model.fc = nn.Identity()
 
         if pretrained:
```
langong347, line 122: Comment about elevating pretrained related logic to the builder level. See #136

examples/mugen/test/retrieval/test_video_clip.py
sophiazhi: wanted to check that the out_dim logic was correct

examples/mugen/retrieval/video_clip.py
sophiazhi: as far as i know it's common to use a pretrained model and then continue training it, but I'm open to other suggestions for simplifying the API if possible

examples/mugen/test/retrieval/test_video_clip.py
sophiazhi: resolving since we can continue discussion in the comment above

langong347 (approved):

PR #126, opened 2022-06-30
[ALBEF] Losses

katrina433:
Summary:
This PR implements the ITC, ITM, and MLM losses used for the ALBEF model
* `ImageTextContrastiveLoss` (ITC) uses image-text similarity from the ALBEF model output to compute the contrastive loss.   [(ALBEF repo implementation)](https://github.com/salesforce/ALBEF/blob/6224e78e5292757ca9c7d5add3e415010955a58b/models/model_pretrain.py#L116-L128)
* `ImageTextMatchingLoss` (ITM) uses positive and negative image-text pairs to predict whether an image and a text are matched or not. [(ALBEF repo implementation)](https://github.com/salesforce/ALBEF/blob/6224e78e5292757ca9c7d5add3e415010955a58b/models/model_pretrain.py#L180-L185)
* `MaskedLanguageModelingLoss` (MLM) computes the loss by predicting the next token auto-regressively. [(ALBEF repo implementation)](https://github.com/salesforce/ALBEF/blob/6224e78e5292757ca9c7d5add3e415010955a58b/models/xbert.py#L1417-L1431)

Test plan:
Wrote unit tests

### DIFF
--- a/torchmultimodal/modules/losses/albef.py
+++ b/torchmultimodal/modules/losses/albef.py
@@ -9,7 +9,6 @@ from typing import Callable, Optional
 import torch
 import torch.nn.functional as F
 from torch import nn, Tensor
-from torch.nn import CrossEntropyLoss
 
 
 class ImageTextContrastiveLoss(nn.Module):
@@ -160,7 +159,6 @@ class MaskedLanguageModelingLoss(nn.Module):
         self.prediction_head = PredictionHead(
             vocab_size, hidden_size, layer_norm_eps, transform_act_fn
         )
-        self.loss_fn = CrossEntropyLoss(reduction="none")
 
     def forward(
         self,
@@ -173,8 +171,10 @@ class MaskedLanguageModelingLoss(nn.Module):
         # shift prediction scores and labels by one for next-token prediction
         prediction_scores = prediction_scores[:, :-1, :].contiguous()
         labels = labels[:, 1:].contiguous()
-        mlm_loss = self.loss_fn(
-            prediction_scores.view(-1, self.vocab_size), labels.view(-1)
+        mlm_loss = F.cross_entropy(
+            prediction_scores.view(-1, self.vocab_size),
+            labels.view(-1),
+            reduction="none",
         )
         mlm_loss = mlm_loss.view(batch_size, -1).sum(1)
 
--- a/torchmultimodal/modules/losses/albef.py
+++ b/torchmultimodal/modules/losses/albef.py
@@ -21,11 +21,11 @@ class ImageTextContrastiveLoss(nn.Module):
         alpha (float): The interpolation value of momentum similarity and sim_targets. Default is 0.
 
     Inputs:
-        sim_i2t (Tensor): Image to text similarity.
-        sim_t2i (Tensor): Text to image similarity.
-        sim_i2t_m (Optional[Tensor]): Image to text similarity from momentum models.
+        image_to_text_sim (Tensor): Image to text similarity.
+        text_to_image_sim (Tensor): Text to image similarity.
+        image_to_text_sim_m (Optional[Tensor]): Image to text similarity from momentum models.
             Required if alpha is non-zero.
-        sim_t2i_m (Optional[Tensor]): Text to image similarity from momentum models.
+        text_to_image_sim_m (Optional[Tensor]): Text to image similarity from momentum models.
             Required if alpha is non-zero.
         sim_targets (Optional[Tensor]): Similarity pseudo-targets from momentum models. Default is the diagonal matrix.
             Requires all inputs to have the same size.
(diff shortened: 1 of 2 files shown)


torchmultimodal/modules/losses/albef.py
```diff
@@ -0,0 +1,216 @@
...
+        weights (Tensor): The weight for each output.
+    """
+
+    def __init__(
+        self,
+        mask_token_id: int = -100,
+        vocab_size: int = 30522,
+        hidden_size: int = 768,
+        layer_norm_eps: float = 1e-12,
+        alpha: float = 0.0,
```
ebsmothers, line 140: Just curious about the choice of 0.0 as the default value here. I know they use it as the default in xbert.py, but I think in the [VQA config](https://github.com/salesforce/ALBEF/blob/main/configs/VQA.yaml#L16) they have it set as 0.4. (I actually like the choice of 0.0 as it means we are dealing with "true" MLM loss with no distillation, but just want to call out that it won't match the default model values.)

torchmultimodal/modules/losses/albef.py
```diff
@@ -0,0 +1,216 @@
...
+        sim_t2i (Tensor): Text to image similarity.
+        sim_i2t_m (Tensor): Image to text similarity from momentum models.
+        sim_t2i_m (Tensor): Text to image similarity from momentum models.
+        sim_targets (Optional[Tensor]): Similarity pseudo-targets from momentum models. Default is the diagonal matrix.
+            Requires all inputs to have the same size.
+    """
+
+    def __init__(
+        self,
+        alpha: float = 0.0,
```
ebsmothers, line 33: My comments on alpha in MLM loss can be applied here as well

torchmultimodal/modules/losses/albef.py
```diff
@@ -0,0 +1,216 @@
...
+            prediction_scores.view(-1, self.vocab_size), labels.view(-1)
+        )
+        mlm_loss = mlm_loss.view(batch_size, -1).sum(1)
+        mlm_loss = (1 - self.alpha) * mlm_loss + self.alpha * loss_distill
+        mlm_loss = weights * mlm_loss
+        mlm_loss = mlm_loss.sum() / batch_size
+        return mlm_loss
+
+
+class PredictionHead(nn.Module):
```
ebsmothers, line 183: This is quite similar to our existing [MLP class](https://github.com/facebookresearch/multimodal/blob/main/torchmultimodal/modules/layers/mlp.py#L13), but the activation and normalization are swapped. A possible follow-up can be to generalize MLP so we can use it here, cc @RdoubleA

torchmultimodal/modules/losses/albef.py
```diff
@@ -0,0 +1,216 @@
...
+        hidden_size: int,
+        layer_norm_eps: float,
+        transform_act_fn: Callable[[Tensor], Tensor],
+    ) -> None:
+        super().__init__()
+        self.dense = nn.Linear(hidden_size, hidden_size)
+        self.transform_act_fn = transform_act_fn
+        self.layer_norm = nn.LayerNorm(hidden_size, eps=layer_norm_eps)
+        self.decoder = nn.Linear(hidden_size, vocab_size, bias=False)
+        self.decoder.bias = nn.Parameter(torch.zeros(vocab_size))
```
ebsmothers, line 209: Why are we setting the bias after initializing with `bias=False`?

torchmultimodal/modules/losses/albef.py
```diff
@@ -0,0 +1,216 @@
...
+            ],
+            dim=0,
+        ).to(vl_embeddings.device)
+        loss_itm = F.cross_entropy(vl_output, itm_labels)
+        return loss_itm
+
+
+class MaskedLanguageModelingLoss(nn.Module):
+    """
+    Compute the autoregressive masked language modeling loss by predicting the next token.
```
ebsmothers, line 117: Maybe call out that this is "as used in VQA" just to make it clear

torchmultimodal/modules/losses/albef.py
```diff
@@ -0,0 +1,216 @@
...
+        hidden_size (int): The hidden size of the prediction_head. Default is 768.
+        layer_norm_eps (float): The epsilon used by the prediction_head normalization layer. Default is 1e-12.
+        alpha (float): The interpolation value between mlm_loss and loss_distill. Default is 0.
+        transform_act_fn (Callable[[Tensor], Tensor]): The activation function in the prediction_head. Default is GELU.
+
+    Inputs:
+        labels (Tensor): The masked output tokens.
+        hidden_states (Tensor): The hidden states of preceding tokens.
+        hidden_states_m (Tensor): The hidden states of preceding tokens from momentum models.
+        weights (Tensor): The weight for each output.
```
ebsmothers, line 131: Can you add shapes for these like you did with the other losses?

torchmultimodal/modules/losses/albef.py
```diff
@@ -0,0 +1,216 @@
...
+        labels = labels[:, 1:].contiguous()
+        with torch.no_grad():
+            prediction_scores_m = self.prediction_head(hidden_states_m)
+            prediction_scores_m = prediction_scores_m[:, :-1, :].contiguous()
+        loss_distill = -torch.sum(
+            F.log_softmax(prediction_scores, dim=-1)
+            * F.softmax(prediction_scores_m, dim=-1),
+            dim=-1,
+        )
+        loss_distill = (loss_distill * (labels != self.mask_token_id)).sum(1)
```
ebsmothers, line 172: If alpha = 0.0, we don't need to compute this, right? And also `hidden_states_m` becomes optional. Might be a nice way to make this work well with a more standard kind of MLM loss. (Though it will make the class a bit more convoluted, so I'll leave it up to you if you wanna do this or not)

torchmultimodal/modules/losses/albef.py
```diff
@@ -0,0 +1,216 @@
...
+
+    def forward(
+        self,
+        embeddings_pos: Tensor,
+        embeddings_neg: Tensor,
+    ) -> Tensor:
+        batch_size = embeddings_pos.size(0)
+        vl_embeddings = torch.cat(
+            [embeddings_pos[:, 0, :], embeddings_neg[:, 0, :]], dim=0
+        )
```
ebsmothers, line 102: Just wondering, why do we want to do the slicing in here? Seems we could just ditch the sequence dimension altogether here.

torchmultimodal/modules/losses/albef.py
```diff
@@ -0,0 +1,216 @@
...
+        vl_output = self.itm_head(vl_embeddings)
+        itm_labels = torch.cat(
+            [
+                torch.ones(batch_size, dtype=torch.long),
+                torch.zeros(2 * batch_size, dtype=torch.long),
```
ebsmothers, line 107: Does the assumption that neg_labels batch size = 2 * pos_labels batch size have to hold? Or can we just allow embeddings_neg to be an arbitrary size and infer the labels here based on its size

torchmultimodal/modules/losses/albef.py
```diff
@@ -0,0 +1,216 @@
...
+
+import torch
+import torch.nn.functional as F
+from torch import nn, Tensor
+from torch.nn import CrossEntropyLoss
+
+
+class ImageTextContrastiveLoss(nn.Module):
+    """
+    Compute the image-text contrastive loss from image-text similarity, as used in ALBEF.
```
ebsmothers, line 17: Would add the additional information that this supports the use of pseudo-targets and so is different from your vanilla image/text contrastive loss. Similar to with MLM, when alpha = 0.0 I think this should just reduce to standard contrastive loss. You can think about making the momentum similarities optional to support that case, and if so it's worth mentioning that as well here.

torchmultimodal/modules/losses/albef.py
```diff
@@ -0,0 +1,216 @@
...
+            dim=-1,
+        )
+        loss_distill = (loss_distill * (labels != self.mask_token_id)).sum(1)
+        mlm_loss = self.loss_fn(
+            prediction_scores.view(-1, self.vocab_size), labels.view(-1)
+        )
+        mlm_loss = mlm_loss.view(batch_size, -1).sum(1)
+        mlm_loss = (1 - self.alpha) * mlm_loss + self.alpha * loss_distill
+        mlm_loss = weights * mlm_loss
+        mlm_loss = mlm_loss.sum() / batch_size
```
ebsmothers, line 179: Do we need to normalize by batch size here? One thing to watch out for is that in distributed training you will have different local and global batch sizes, so make sure you are using the right one

ebsmothers (commented):
Looks good! Lots of small comments but overall I like the way you wrote each of these losses

torchmultimodal/modules/losses/albef.py
katrina433: removed

ebsmothers (approved):
Looks good!

torchmultimodal/modules/losses/albef.py
```diff
@@ -0,0 +1,230 @@
...
+        embeddings_neg (Tensor of shape (batch_size_neg, hidden_size)):
+            The multimodal embeddings for negative image-text pairs.
+    """
+
+    def __init__(
+        self,
+        hidden_size: int = 768,
+    ) -> None:
+        super().__init__()
+        self.itm_head = nn.Linear(
```
ankitade, line 101: why make the head part of loss ?

torchmultimodal/modules/losses/albef.py
```diff
@@ -0,0 +1,230 @@
...
+        transform_act_fn: Callable[[Tensor], Tensor] = nn.functional.gelu,
+    ) -> None:
+        super().__init__()
+        self.mask_token_id = mask_token_id
+        self.vocab_size = vocab_size
+        self.alpha = alpha
+        self.prediction_head = PredictionHead(
+            vocab_size, hidden_size, layer_norm_eps, transform_act_fn
+        )
+        self.loss_fn = CrossEntropyLoss(reduction="none")
```
ankitade, line 161: maybe make it consistent in all place in the file, either use F or nn.module

PR #130, opened 2022-07-04
[Omnivore] Add more variant and pretrained_weight capability

YosuaMichael:
Summary:
Adding new variant: swin_s and swin_b with pretrained weight for omnivore.

Test plan:
Create a test for each variant, also I change the test to use pytest instead of unittest.

Usage:
Here are example script to get the pretrained weight:
```
import torchmultimodal.models.omnivore as omnivore

model = omnivore.omnivore_swin_s(pretrained=True)
```


torchmultimodal/models/omnivore.py
```diff
@@ -104,11 +115,97 @@ def _omnivore_swin_t_encoder() -> SwinTransformer3d:
...
+        progress (bool): If true then there will be a progress bar for downloading weight (default: True)
+    """
     encoder = _omnivore_swin_t_encoder()
+    heads = _multimodal_head(input_dim=encoder.num_features)
+    model = OmnivoreArchitecture(encoder, heads)
+    if pretrained:
+        model.load_model(_OMNIVORE_PRETRAINED_URLS["swin_t"])
     if encoder_only:
-        return encoder
+        return model.encoder
```
ankitade, line 165: instead of this flag, people can just call _omnivore_swin_t_encoder() right if u allow ckpt loading in that function ?

ankitade: i think same ckpt should work if u set strict to false. try?

YosuaMichael: I have tried using `strict=False` and it still failed because the checkpoint key has `encoder.` prefix compared to what encoder actually expected.
For this case, I am thinking of storing `encoder` and `head` weight separately, and let the encoder and head load separately. But for this, I am thinking of creating utils for `load_model_from_url` since both model and encoder would not have `Mixin` (since the head dont have its own class, and encoder will be upstreamed and it shouldn't inherit PretrainedMixin).
What do you think about this approach @ankitade ?

torchmultimodal/architectures/omnivore.py
```diff
@@ -6,9 +6,10 @@
 
 import torch
 from torch import nn
+from torchmultimodal.utils.common import PretrainedMixin
 
 
-class OmnivoreArchitecture(nn.Module):
+class OmnivoreArchitecture(nn.Module, PretrainedMixin):
```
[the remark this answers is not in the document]
YosuaMichael: Thanks, I didn't know we deprecate the architecture before. I will put the architecture on `models/omnivore.py` then

test/models/test_omnivore.py
```diff
@@ -4,46 +4,110 @@
...
+    return omnivore.omnivore_swin_b().to(device)
+
+
+def test_omnivore_swin_t_forward(omnivore_swin_t_model):
+    model = omnivore_swin_t_model
+
+    image = torch.randn(1, 3, 1, 112, 112)  # B C D H W
+    image_score = model(image, input_type="image")
+
+    assert_expected(torch.tensor(image_score.size()), torch.tensor((1, 1000)))
```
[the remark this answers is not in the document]
YosuaMichael: This because `assert_expected` expect tensor for its input and I thought `torch.Size` is not really tensor

torchmultimodal/models/omnivore.py
```diff
@@ -104,11 +110,97 @@ def _omnivore_swin_t_encoder() -> SwinTransformer3d:
...
+        encoder_only (bool): If true then the builder will only return encoder without head (default: False)
+        pretrained (bool): If true then the it will load pretrained weight,
+            otherwise it will have random weight (default: False)
+        progress (bool): If true then there will be a progress bar for downloading weight (default: True)
+    """
     encoder = _omnivore_swin_t_encoder()
+    heads = _multimodal_head(input_dim=encoder.num_features)
+    model = OmnivoreArchitecture(encoder, heads)
+    if pretrained:
+        model.load_model(_OMNIVORE_PRETRAINED_URLS["swin_t"])
```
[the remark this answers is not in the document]
YosuaMichael: I think in general we dont really test loading pretrained weight in torchvision now since it is time consuming (the weight is usually large). In fact in some test, we disable the `load_state_dict` function explicitly ([see this function](https://github.com/pytorch/vision/blob/main/test/test_models.py#L34)).

> how do you unit test this? In our case, we create small files to store under test/assets/ and load them locally into the model to test. I've seen torchvision also has asset files, wonder if you consider this a viable approach?

I dont really get what you mean here. I think in general the model file is quite big (several hundred MBs or few GBs). So it is not ideal to store on the repo.

test/models/test_omnivore.py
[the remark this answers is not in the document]
YosuaMichael: Let me check if just passing `torch.Size` will be okay. If yes, I will change accordingly

torchmultimodal/models/omnivore.py
ankitade: I see the issue. its fine to leave as is for this PR. 
util is a good idea but how will u generalize it? would be good to put up a draft pr/notebook

test/models/test_omnivore.py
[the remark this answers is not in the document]
YosuaMichael: Using `torch.Size` will cause complaint by mypy if I did `mypy --config=mypy.ini test/models/test_omnivore.py`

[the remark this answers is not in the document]
YosuaMichael: Resolve by changing the `assert_expected` input type to Any

torchmultimodal/utils/common.py
```diff
@@ -133,6 +133,19 @@ def transpose_for_scores(
     return x.permute(0, 2, 1, 3)
 
 
+def load_module_from_url(m: torch.nn.Module, url: str, progress=True):
```
langong347, line 136: This function duplicates what `PretrainedMixin` does? Is it possible to also refactor `PretrainedMixin` such that `load_model` would call the utility?

torchmultimodal/models/omnivore.py
```diff
@@ -8,20 +8,53 @@
...
 }
 
 
+class OmnivoreArchitecture(nn.Module):
+    """Omnivore is a model that accept multiple vision modality.
+
+    Omnivore (https://arxiv.org/abs/2201.08377) is a single model that able to do classification
+    on images, videos, and single-view 3D data using the same shared parameters of the encoder.
+
+    Args:   encoder (nn.Module): Instantiated encoder.
```
langong347, line 34: start a new line for the list of args and inputs?

https://google.github.io/styleguide/pyguide.html#383-functions-and-methods

torchmultimodal/models/omnivore.py
```diff
@@ -137,70 +182,70 @@ def _omnivore_swin_b_encoder() -> SwinTransformer3d:
...
+    encoder = omnivore_swin_t_encoder(pretrained=pretrained)
     heads = _multimodal_head(input_dim=encoder.num_features)
-    model = OmnivoreArchitecture(encoder, heads)
     if pretrained:
-        model.load_model(_OMNIVORE_PRETRAINED_URLS["swin_t"])
-    if encoder_only:
-        return model.encoder
-    else:
-        return model
+        common_utils.load_module_from_url(
```
langong347, line 205: Ok. I see why you are pivoting towards the util because now you are loading the encoder and heads separately.

torchmultimodal/models/omnivore.py
YosuaMichael: Thanks for noticing!

torchmultimodal/models/omnivore.py
YosuaMichael: yes exactly

torchmultimodal/utils/common.py
ankitade: yosua can u just create a task and send me. we should try moving to this util for other models if possible and deprecate the mixin

torchmultimodal/models/omnivore.py
```diff
@@ -89,7 +130,7 @@ def forward(self, x: torch.Tensor) -> torch.Tensor:
         return x
 
 
-def _omnivore_swin_t_encoder() -> SwinTransformer3d:
+def omnivore_swin_t_encoder(pretrained=False, progress=True) -> SwinTransformer3d:
```
ankitade, line 133: do u really need to expose the progress param here?

torchmultimodal/utils/common.py
ankitade: nit : can u change m to "model"

ankitade (approved):

torchmultimodal/utils/common.py
YosuaMichael: I will change `m` to be `module`, and will change `model_dir` to `module_dir` as well. Because I think the term `module` is more generic and suitable for this case.

torchmultimodal/models/omnivore.py
YosuaMichael: Actually no need to expose, I just follow [torchvision style](https://github.com/pytorch/vision/blob/main/torchvision/models/vision_transformer.py#L699) but no strong opinion on this.

If you think not exposing is better, then I will remove it. One question though, if we are not exposing, should we do `progress=False` or `progress=True` on the underlying call?

langong347: I am pro exposing them as the callsite may not want to default `pretrained=False`.

YosuaMichael: I think overall no one has strong opinion on removing `progress=False` hence I will resolve this

PR #135, opened 2022-07-06
Refactor S3D model

sophiazhi:
Summary:
Followup to #104: Refactor S3D model to reduce code repetition

Test plan:
`python -m pytest examples/mugen/test/retrieval/test_s3d.py -rP`
Coverage: 
```python -m pytest --cov=examples/mugen/retrieval examples/mugen/test/retrieval/test_s3d.py -vv

examples/mugen/test/retrieval/test_s3d.py::TestS3D::test_forward PASSED                     [ 25%]
examples/mugen/test/retrieval/test_s3d.py::TestS3D::test_basicconv3d PASSED                 [ 50%]
examples/mugen/test/retrieval/test_s3d.py::TestS3D::test_sepconv3d PASSED                   [ 75%]
examples/mugen/test/retrieval/test_s3d.py::TestS3D::test_mixedconvsblock PASSED             [100%]

Name                                     Stmts   Miss  Cover
examples/mugen/retrieval/s3d.py             66      0   100%
examples/mugen/retrieval/video_clip.py      28     28     0%
TOTAL                                       94     28    70%


======================================== 4 passed in 5.03s ========================================

### DIFF
--- a/examples/mugen/retrieval/s3d.py
+++ b/examples/mugen/retrieval/s3d.py
@@ -30,17 +30,17 @@ class S3D(nn.Module):
             BasicConv3d(64, 64, kernel_size=1, stride=1),
             SepConv3d(64, 192, kernel_size=3, stride=1, padding=1),
             nn.MaxPool3d(kernel_size=(1, 3, 3), stride=(1, 2, 2), padding=(0, 1, 1)),
-            Mixed3b(),
-            Mixed3c(),
+            MixedConvsBlock(192, 64, 96, 128, 16, 32, 32),
+            MixedConvsBlock(256, 128, 128, 192, 32, 96, 64),
             nn.MaxPool3d(kernel_size=(3, 3, 3), stride=(2, 2, 2), padding=(1, 1, 1)),
-            Mixed4b(),
-            Mixed4c(),
-            Mixed4d(),
-            Mixed4e(),
-            Mixed4f(),
+            MixedConvsBlock(480, 192, 96, 208, 16, 48, 64),
+            MixedConvsBlock(512, 160, 112, 224, 24, 64, 64),
+            MixedConvsBlock(512, 128, 128, 256, 24, 64, 64),
+            MixedConvsBlock(512, 112, 144, 288, 32, 64, 64),
+            MixedConvsBlock(528, 256, 160, 320, 32, 128, 128),
             nn.MaxPool3d(kernel_size=(2, 2, 2), stride=(2, 2, 2), padding=(0, 0, 0)),
-            Mixed5b(),
-            Mixed5c(),
+            MixedConvsBlock(832, 256, 160, 320, 32, 128, 128),
+            MixedConvsBlock(832, 384, 192, 384, 48, 128, 128),
         )
         self.fc = nn.Conv3d(1024, num_class, kernel_size=1, stride=1, bias=True)
 
@@ -111,25 +111,36 @@ class SepConv3d(nn.Module):
         return x
 
 
-class Mixed3b(nn.Module):
-    def __init__(self):
-        super(Mixed3b, self).__init__()
+class Branch0(nn.Sequential):
+    def __init__(self, in_planes, out_planes):
+        super(Branch0, self).__init__(
+            BasicConv3d(in_planes, out_planes, kernel_size=1, stride=1)
+        )
 
-        self.branch0 = nn.Sequential(
-            BasicConv3d(192, 64, kernel_size=1, stride=1),
+
+class Branch1or2(nn.Sequential):
+    def __init__(self, in_planes, mid_planes, out_planes):
+        super(Branch1or2, self).__init__(
+            BasicConv3d(in_planes, mid_planes, kernel_size=1, stride=1),
+            SepConv3d(mid_planes, out_planes, kernel_size=3, stride=1, padding=1),
         )
-        self.branch1 = nn.Sequential(
-            BasicConv3d(192, 96, kernel_size=1, stride=1),
-            SepConv3d(96, 128, kernel_size=3, stride=1, padding=1),
-        )
-        self.branch2 = nn.Sequential(
-            BasicConv3d(192, 16, kernel_size=1, stride=1),
-            SepConv3d(16, 32, kernel_size=3, stride=1, padding=1),
-        )
-        self.branch3 = nn.Sequential(
+
+
+class Branch3(nn.Sequential):
+    def __init__(self, in_planes, out_planes):
+        super(Branch3, self).__init__(
             nn.MaxPool3d(kernel_size=(3, 3, 3), stride=1, padding=1),
-            BasicConv3d(192, 32, kernel_size=1, stride=1),
+            BasicConv3d(in_planes, out_planes, kernel_size=1, stride=1),
         )
+
+
+class MixedConvsBlock(nn.Module):
+    def __init__(self, in_planes, b0_out, b1_mid, b1_out, b2_mid, b2_out, b3_out):
+        super().__init__()
+        self.branch0 = Branch0(in_planes, b0_out)
+        self.branch1 = Branch1or2(in_planes, b1_mid, b1_out)
+        self.branch2 = Branch1or2(in_planes, b2_mid, b2_out)
+        self.branch3 = Branch3(in_planes, b3_out)
 
     def forward(self, x):
         x0 = self.branch0(x)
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: examples/mugen/retrieval/s3d.py [hunk 1]
```diff
@@ -111,262 +111,42 @@ def forward(self, x):
...
-    def __init__(self):
-        super(Mixed4b, self).__init__()
-
-        self.branch0 = nn.Sequential(
-            BasicConv3d(480, 192, kernel_size=1, stride=1),
-        )
-        self.branch1 = nn.Sequential(
-            BasicConv3d(480, 96, kernel_size=1, stride=1),
-            SepConv3d(96, 208, kernel_size=3, stride=1, padding=1),
+class Branch1or2(nn.Sequential):
```
ebsmothers, line 121: nit: this naming feels a bit clunky. Not sure I have a better suggestion (MiddleBranch? Idk). Ultimately will leave it up to you if it's better to just stick with the names from the original repo though

examples/mugen/retrieval/s3d.py
```diff
@@ -111,262 +111,42 @@ def forward(self, x):
...
+    def __init__(self, in_planes, mid_planes, out_planes):
```
ebsmothers, line 122: another nit: would maybe call this `hidden_planes` instead of `mid_planes`

examples/mugen/retrieval/s3d.py
```diff
@@ -111,262 +111,42 @@ def forward(self, x):
...
-        self.branch2 = nn.Sequential(
-            BasicConv3d(512, 24, kernel_size=1, stride=1),
-            SepConv3d(24, 64, kernel_size=3, stride=1, padding=1),
-        )
-        self.branch3 = nn.Sequential(
-            nn.MaxPool3d(kernel_size=(3, 3, 3), stride=1, padding=1),
-            BasicConv3d(512, 64, kernel_size=1, stride=1),
-        )
+class MixedConvsBlock(nn.Module):
+    def __init__(self, in_planes, b0_out, b1_mid, b1_out, b2_mid, b2_out, b3_out):
```
ebsmothers, line 138: similar comment here around naming

examples/mugen/retrieval/s3d.py
ebsmothers, line 122: nit: add type info (here and in other classes). Mypy won't complain about examples/ so it's not a hard requirement, but still will make the code clearer.

ebsmothers (approved):
Looks great! Thanks for refactoring these classes, the way you're doing it now is really nice. Left a few nits but otherwise looks good.

langong347:
@YosuaMichael Would love to get your eyes on this. This way of implementing inception blocks is different from [torchvision](https://github.com/pytorch/vision/blob/main/torchvision/models/inception.py). This version of using S3D as the backbone may as well be a good addition to TorchVision' inception class in the future.

examples/mugen/retrieval/s3d.py
```diff
@@ -30,17 +30,17 @@ def __init__(self, num_class):
             BasicConv3d(64, 64, kernel_size=1, stride=1),
```
YosuaMichael, line 30: I think we can make a more generic class that able to handle both `S3D` and `I3D` (see https://arxiv.org/pdf/1705.07750.pdf). The idea (just to illustrate, the naming might not be good) is to enable user to choose the `class` for the `conv_block (S3D: SepConv3D, I3D: Conv3D)`, `max_pool_block`, and `inception_block (S3D: SepInceptionModule3D, I3D: InceptionModule3D)` (examples: [here](https://github.com/pytorch/vision/blob/main/torchvision/models/video/mvit.py#L352) or [here](https://github.com/pytorch/vision/blob/main/torchvision/models/efficientnet.py#L306)). 

It is also good to let the parameters also be configurable on the generic class, this will be useful if there will be a new variant with different parameters in the future. This is not a must, however since there are a lot of parameters, it might be a good idea to create a config class to store the parameters ([example](https://github.com/pytorch/vision/blob/main/torchvision/models/efficientnet.py#L377)).

Once we have the generic module, we can have a builder to build the s3d model ([example](https://github.com/pytorch/vision/blob/main/torchvision/models/video/mvit.py#L464))

examples/mugen/retrieval/s3d.py
YosuaMichael: NIT: just a suggestion on the parameters, we might consider to single parameter called: `branch_layers` and it is a list of list of integers representing the layer parameters of the branch, i.e:
```
[[b0_out], [b1_mid, b1_out], [b2_mid, b2_out], [b3_out]]
```
with this the constructor can become: `def __init__(self, in_planes, branch_layers)`

YosuaMichael (commented):
Hi @sophiazhi , thanks the PR! I have a few comments on making more general class for S3D and some suggestions for the implementation.

YosuaMichael:
I also want to suggest to reuse [Conv3dNormActivation](https://github.com/pytorch/vision/blob/main/torchvision/ops/misc.py#L175)  from torchvision. I think this can be used to replace `BasicConv3d` by using 
```
Conv3dNormActivation(64, 64, kernel_size=1, stride=1, norm_layer=partial(nn.BatchNorm3d, eps=1e-3, momentum=1e-3))
```

Also `SepConv3d` can also be simplified:
```
class SepConv3d(nn.Sequential):
    def __init__(self, in_planes, out_planes, kernel_size, stride, padding=0):
        super(SepConv3d, self).__init__(
            Conv3dNormActivation(
                in_planes,
                out_planes,
                kernel_size=(1, kernel_size, kernel_size),
                stride=(1, stride, stride),
                padding=(0, padding, padding),
                bias=False,
            ),
            Conv3dNormActivation(
                out_planes,
                out_planes,
                kernel_size=(kernel_size, 1, 1),
                stride=(stride, 1, 1),
                padding=(padding, 0, 0),
                bias=False,
            ),
        )
```

Another thing is that the `padding` is usually constructor with [common formula](https://github.com/pytorch/vision/blob/main/torchvision/ops/misc.py#L92) of `(kernel_size - 1) // 2 * dilation` and I observe that it is the case here too. In this case, you can also not passing the padding manually :)

examples/mugen/retrieval/s3d.py
sophiazhi: changed the param format to `List[List[int]]`, but keeping the "mid" naming for consistency with torchvision [video resnet](https://pytorch.org/vision/stable/_modules/torchvision/models/video/resnet.html) naming

examples/mugen/retrieval/s3d.py
sophiazhi: resolving, see comment below

examples/mugen/retrieval/s3d.py
```diff
@@ -111,262 +111,42 @@ def forward(self, x):
...
-            BasicConv3d(192, 16, kernel_size=1, stride=1),
-            SepConv3d(16, 32, kernel_size=3, stride=1, padding=1),
-        )
-        self.branch3 = nn.Sequential(
-            nn.MaxPool3d(kernel_size=(3, 3, 3), stride=1, padding=1),
-            BasicConv3d(192, 32, kernel_size=1, stride=1),
+class Branch0(nn.Sequential):
+    def __init__(self, in_planes, out_planes):
+        super(Branch0, self).__init__(
+            BasicConv3d(in_planes, out_planes, kernel_size=1, stride=1)
```
datumbox, line 117: TorchVision has a standard block for `BasicConv3d`. It's called `Conv3dNormActivation`.

`SepConv3d` can also be rewritten based on it. Not sure if we should call them separable conv because there are normalization and activation layers in between them but we can review naming later.

examples/mugen/retrieval/s3d.py
```diff
@@ -111,262 +111,42 @@ def forward(self, x):
...
-            nn.MaxPool3d(kernel_size=(3, 3, 3), stride=1, padding=1),
-            BasicConv3d(512, 64, kernel_size=1, stride=1),
-        )
+class MixedConvsBlock(nn.Module):
+    def __init__(self, in_planes, b0_out, b1_mid, b1_out, b2_mid, b2_out, b3_out):
+        super().__init__()
+        self.branch0 = Branch0(in_planes, b0_out)
+        self.branch1 = Branch1or2(in_planes, b1_mid, b1_out)
+        self.branch2 = Branch1or2(in_planes, b2_mid, b2_out)
+        self.branch3 = Branch3(in_planes, b3_out)
```
datumbox, line 143: Can we avoid the introduction of the `Branch*` classes? From what I understand, we can just use Sequential directly and simplify the code structure.

examples/mugen/retrieval/s3d.py
```diff
@@ -30,17 +30,17 @@ def __init__(self, num_class):
...
+            MixedConvsBlock(480, 192, 96, 208, 16, 48, 64),
+            MixedConvsBlock(512, 160, 112, 224, 24, 64, 64),
+            MixedConvsBlock(512, 128, 128, 256, 24, 64, 64),
+            MixedConvsBlock(512, 112, 144, 288, 32, 64, 64),
+            MixedConvsBlock(528, 256, 160, 320, 32, 128, 128),
             nn.MaxPool3d(kernel_size=(2, 2, 2), stride=(2, 2, 2), padding=(0, 0, 0)),
-            Mixed5b(),
-            Mixed5c(),
+            MixedConvsBlock(832, 256, 160, 320, 32, 128, 128),
+            MixedConvsBlock(832, 384, 192, 384, 48, 128, 128),
```
datumbox, line 43: Sorry putting a comment on a random position because Github doesn't let me comment on the constructor of the `S3D` class. We need to align the arguments you pass to make them similar to what TorchVision is currently using (see [Video Resnet](https://github.com/pytorch/vision/blob/main/torchvision/models/video/resnet.py#L197) class, nothing to different).

datumbox (commented):
@sophiazhi In continuation to our discussion at https://github.com/pytorch/vision/issues/6402, I think this implementation is quite good and worth upstreaming to TorchVision.

I've added a few comments to indicate parts that would probably need to change to make it easier to upstream. My comments are minor and don't really affect your implementation; rather just suggest specific idioms that TorchVision uses.

There is a bit of boilerplate code that would also be required for creating model builder methods and weights (we will need to port the paper weights) but again nothing too much. We can guide you through the process during the PR. My recommendation is to open one on the repo and I can guide you from there.

sophiazhi:
Closing due to pytorch/vision#6412

PR #137, opened 2022-07-07
Make contrastive loss temperature clamping configurable

sophiazhi:
Summary:
Add args to ContrastiveLossWithTemperature to make temperature clamping min and max configurable. This change will allow us to replicate the MUGEN contrastive loss function (investigated in [colab notebook](https://colab.research.google.com/drive/1QQFi96ENNR2njcp-4rXpIw5rBtL-2Sj9)).

- `contrastive_loss_with_temperature.py`: add `logit_scale_min` and `logit_scale_max` optional args
- `test_contrastive_loss_with_temperature.py`: add test of clamping functionality

Test plan:
`python -m pytest --cov=torchmultimodal/modules/losses test/modules/losses/test_contrastive_loss_with_temperature.py -rP`
```test/modules/losses/test_contrastive_loss_with_temperature.py .ss.                                                            [100%]

============================================================== PASSES ===============================================================

---------- coverage: platform linux, python 3.8.13-final-0 -----------
Name                                                                  Stmts   Miss  Cover   Missing
---------------------------------------------------------------------------------------------------
torchmultimodal/modules/losses/contrastive_loss_with_temperature.py      55     14    75%   37-64, 120-122
torchmultimodal/modules/losses/flava.py                                 194    194     0%   7-479
torchmultimodal/modules/losses/vqvae.py                                  10     10     0%   7-30
---------------------------------------------------------------------------------------------------
TOTAL                                                                   259    218    16%

=================================================== 2 passed, 2 skipped in 5.01s ====================================================
```
Missing lines are either for GPU or not accessible through the main loss module (`mask=None`).


torchmultimodal/modules/losses/contrastive_loss_with_temperature.py
```diff
@@ -158,6 +158,10 @@ class ContrastiveLossWithTemperature(nn.Module):
         logit_scale (Union[float, nn.Module]): Log of the learnable temperature parameter value
             A nn.Parameter instantiation can also be passed directly in case parent class
             is handling the initialization.
+        logit_scale_min (Optional[float]): Log of the minimum temperature value.
+            If None, then temperature will not be clamped to a minimum value.
```
ebsmothers, line 162: Is this true? I thought it will be clamped to the default value

torchmultimodal/modules/losses/contrastive_loss_with_temperature.py
```diff
@@ -158,6 +158,10 @@ class ContrastiveLossWithTemperature(nn.Module):
...
+        logit_scale_max (Optional[float]): Log of the maximum temperature value.
+            If None, then temperature will not be clamped to a maximum value.
```
ebsmothers, line 164: same here

test/modules/losses/test_contrastive_loss_with_temperature.py
```diff
@@ -53,6 +53,20 @@ def test_local_loss(self):
         self.assertEqual(loss.size(), torch.Size([]))
         self.assertAlmostEqual(loss.item(), 9.8753, 3)
 
+    def test_temperature_clamp(self):
```
ebsmothers, line 56: nit: this is really just testing that the existing clamping works as expected. Would be nice to instead explicitly pass values for logit_scale_min/max here

torchmultimodal/modules/losses/contrastive_loss_with_temperature.py
sophiazhi: From [pytorch docs](https://pytorch.org/docs/stable/generated/torch.clamp.html#:~:text=If%20min%20is%20None%2C%20there%20is%20no%20lower%20bound.%20Or%2C%20if%20max%20is%20None%20there%20is%20no%20upper%20bound.): "If min is None, there is no lower bound. Or, if max is None there is no upper bound."

ebsmothers: I see, I was referring more to the default behavior here. In other words, if the user does not pass logit_scale_min or logit_scale_max we will clamp to the default values from CLIP. Would be worth mentioning this in the docstring as well.

Btw might want to do a check that at least one of logit_scale_min and logit_scale_max is not None, otherwise clamp will throw an error.

sophiazhi: oh ok, I added defaults for all args in the docstring and a ValueError for the both None case

torchmultimodal/modules/losses/contrastive_loss_with_temperature.py
```diff
@@ -167,9 +174,21 @@ class ContrastiveLossWithTemperature(nn.Module):
...
+        self,
+        logit_scale: Union[float, nn.Parameter] = DEFAULT_LOGIT_SCALE,
+        logit_scale_min: Optional[float] = 0.0,
+        logit_scale_max: Optional[float] = 4.6052,
+    ):
         super().__init__()
 
+        if not logit_scale_min and not logit_scale_max:
+            raise ValueError(
+                "At least one of `logit_scale_min` and `logit_scale_max` must not be None."
```
RdoubleA, line 187:
```suggestion
                "Only one of `logit_scale_min` and `logit_scale_max` can be None."
```

RdoubleA (approved):

torchmultimodal/modules/losses/contrastive_loss_with_temperature.py
```diff
@@ -167,9 +174,21 @@ class ContrastiveLossWithTemperature(nn.Module):
                 all_gather to all workers (versus just the local worker).
     """
 
-    def __init__(self, logit_scale: Union[float, nn.Parameter] = DEFAULT_LOGIT_SCALE):
+    def __init__(
+        self,
+        logit_scale: Union[float, nn.Parameter] = DEFAULT_LOGIT_SCALE,
+        logit_scale_min: Optional[float] = 0.0,
+        logit_scale_max: Optional[float] = 4.6052,
```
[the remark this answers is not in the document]
sophiazhi: I was thinking about that too but wasn't sure if there's a reason the original clamp code used `4.6052` instead (git blame just says "initial commit" so i'm not sure who wrote it?). I'll change to `math.log` unless someone has a reason to keep 4.6052

ebsmothers (approved):

langong347 (approved):

PR #138, opened 2022-07-07
[MDETR] Add contrastive alignment loss

ebsmothers:
Summary:
This PR adds support for the contrastive alignment loss from MDETR. This loss also necessitates
adding two projections to the MDETR model, and returning some intermediate outputs. I modified the
return types of MDETRTransformer and MDETR to be NamedTuples in the process of making these changes.

Note that the contrastive alignment loss uses the mapping between character positions and token IDs
and so needs access to the transformers tokenizer. For this reason I am putting this loss in examples/
instead of in core.

Test plan:
Added a new test for the contrastive loss, updated the existing MDETR model tests
to account for different random initialization orders and new params.
Confirmed in LoadAndComparePretrainedWeights.ipynb that the outputs line up with the MDETR checkpoint.

```
python -m pytest -v
...
========================= 175 passed, 2 skipped, 2 xfailed, 6 warnings in 46.57s ================
```

### DIFF
--- a/examples/mdetr/contrastive_alignment_loss.py
+++ b/examples/mdetr/contrastive_alignment_loss.py
@@ -19,7 +19,7 @@ def contrastive_alignment_loss(
     num_boxes: int,
     tokenized: BatchEncoding,
     temperature: float = 0.07,
-):
+) -> Tensor:
     """Contrastive alignment loss.
 
     Enforces alignment between the text representations after cross encoder and the
@@ -81,7 +81,12 @@ def contrastive_alignment_loss(
     return tot_loss / num_boxes
 
 
-def construct_positive_map(logits: Tensor, target_tokens, indices, tokenized):
+def construct_positive_map(
+    logits: Tensor,
+    target_tokens: List[List[List[int]]],
+    indices: List[Tuple[Tensor, Tensor]],
+    tokenized: BatchEncoding,
+):
     # construct a map such that positive_map[k, i,j] = True iff query i is associated to token j in batch item k
     # For efficency, the construction happens on CPU, then the whole matrix is transferred to GPU in one go.
     positive_map = torch.zeros(logits.shape, dtype=torch.bool)
@@ -91,23 +96,10 @@ def construct_positive_map(logits: Tensor, target_tokens, indices, tokenized):
             for (beg, end) in tok_list:
                 beg_pos = tokenized.char_to_token(i, beg)
                 end_pos = tokenized.char_to_token(i, end - 1)
-                if beg_pos is None:
-                    try:
-                        beg_pos = tokenized.char_to_token(beg + 1)
-                        if beg_pos is None:
-                            beg_pos = tokenized.char_to_token(beg + 2)
-                    except Exception:
-                        beg_pos = None
-                if end_pos is None:
-                    try:
-                        end_pos = tokenized.char_to_token(end - 2)
-                        if end_pos is None:
-                            end_pos = tokenized.char_to_token(end - 3)
-                    except Exception:
-                        end_pos = None
-                if beg_pos is None or end_pos is None:
-                    continue
 
-                assert beg_pos is not None and end_pos is not None
+                if beg_pos is None and end_pos is None:
+                    raise ValueError(
+                        "At least one of beg_pos and end_pos must not be None"
+                    )
                 positive_map[i, idx_src[j], beg_pos : end_pos + 1].fill_(True)
     return positive_map.to(logits.device)
--- a/examples/mdetr/contrastive_alignment_loss.py
+++ b/examples/mdetr/contrastive_alignment_loss.py
@@ -1,0 +1,113 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+from typing import List, Tuple
+
+import torch
+from torch import Tensor
+from transformers.tokenization_utils_base import BatchEncoding
+
+
+def contrastive_alignment_loss(
+    projected_queries: Tensor,
+    projected_tokens: Tensor,
+    target_tokens: List[List[List[int]]],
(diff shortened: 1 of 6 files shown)


langong347 (commented):
Would be possible to add comments about dimensions for at least the inputs and the outputs of each class? Image (and video) data are high dimensional and often we rearrange the dimensions within a module. One can quickly lose track of what's going on.

examples/mdetr/contrastive_alignment_loss.py
```diff
@@ -0,0 +1,113 @@
...
+    tokens_to_boxes_loss = (
+        ((pos_term / nb_pos + neg_term)).masked_fill(~tokens_with_pos, 0).sum()
+    )
+
+    tot_loss = (box_to_token_loss + tokens_to_boxes_loss) / 2
+
+    return tot_loss / num_boxes
+
+
+def construct_positive_map(logits: Tensor, target_tokens, indices, tokenized):
```
RdoubleA, line 84: type annotations and return type

examples/mdetr/contrastive_alignment_loss.py
```diff
@@ -0,0 +1,113 @@
...
+
+def contrastive_alignment_loss(
+    projected_queries: Tensor,
+    projected_tokens: Tensor,
+    target_tokens: List[List[List[int]]],
+    indices: List[Tuple[Tensor, Tensor]],
+    num_boxes: int,
+    tokenized: BatchEncoding,
+    temperature: float = 0.07,
+):
```
RdoubleA, line 22: return type

examples/mdetr/contrastive_alignment_loss.py
```diff
@@ -0,0 +1,113 @@
...
+        for j, tok_list in enumerate(cur_tokens):
+            for (beg, end) in tok_list:
+                beg_pos = tokenized.char_to_token(i, beg)
+                end_pos = tokenized.char_to_token(i, end - 1)
+                if beg_pos is None:
+                    try:
+                        beg_pos = tokenized.char_to_token(beg + 1)
+                        if beg_pos is None:
+                            beg_pos = tokenized.char_to_token(beg + 2)
+                    except Exception:
```
RdoubleA, line 99: why are we checking offsets of beg and end? is it to find multiple tokens that are associated with the query? then why only upto offset of 3?

examples/mdetr/contrastive_alignment_loss.py
```diff
@@ -0,0 +1,113 @@
...
+
+
+def construct_positive_map(logits: Tensor, target_tokens, indices, tokenized):
+    # construct a map such that positive_map[k, i,j] = True iff query i is associated to token j in batch item k
+    # For efficency, the construction happens on CPU, then the whole matrix is transferred to GPU in one go.
+    positive_map = torch.zeros(logits.shape, dtype=torch.bool)
+    for i, ((idx_src, idx_tgt), tgt) in enumerate(zip(indices, target_tokens)):
+        cur_tokens = [tgt[j] for j in idx_tgt]
+        for j, tok_list in enumerate(cur_tokens):
+            for (beg, end) in tok_list:
```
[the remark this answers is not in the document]
ebsmothers: So the outer loop is for each sample in the batch. The inner two are both related to how the tokens are associated to boxes. In general a given box may be represented by multiple phrases in the caption, which is why they are represented as a set of disjoint intervals in the token space. So the middle loop iterates over boxes, while the innermost loop iterates over these disjoint intervals. In practice, it is pretty rare that a given box has multiple intervals like this, but it does happen.

TLDR: I don't think we can condense it 😅, but it is most often a no-op anyways

examples/mdetr/contrastive_alignment_loss.py
ebsmothers: I think this is just some heuristic they use for error handling. Actually I just tested on the phrase grounding eval code and it never even hits this portion of the code (the first char_to_token call always works). So I will take it out to keep things simpler, if we find that we need it for VQA or something we can always add it back later on

examples/mdetr/contrastive_alignment_loss.py
```diff
@@ -0,0 +1,105 @@
...
+    tokens_to_boxes_loss = (
+        ((pos_term / nb_pos + neg_term)).masked_fill(~tokens_with_pos, 0).sum()
+    )
+
+    tot_loss = (box_to_token_loss + tokens_to_boxes_loss) / 2
+
+    return tot_loss / num_boxes
+
+
+def construct_positive_map(
```
RdoubleA, line 84: add a test for this?

examples/mdetr/test/test_contrastive_alignment_loss.py
```diff
@@ -0,0 +1,131 @@
...
+# LICENSE file in the root directory of this source tree.
+
+import pytest
+import torch
+from examples.mdetr.contrastive_alignment_loss import contrastive_alignment_loss
+from test.test_utils import assert_expected, set_rng_seed
+from transformers import RobertaTokenizerFast
+
+
+@pytest.fixture(scope="class", autouse=True)
```
RdoubleA, line 14: we decided on using function scope as default right? but if this changes all the output values then let's leave it

RdoubleA (approved):

examples/mdetr/contrastive_alignment_loss.py
```diff
@@ -0,0 +1,105 @@
...
+    positive_logits = -logits.masked_fill(~positive_map, 0)
+    negative_logits = logits
+
+    # Calculate the contrastive loss for all objects
+    boxes_with_pos = positive_map.any(2)
+    pos_term = positive_logits.sum(2)
+    neg_term = negative_logits.logsumexp(2)
+    nb_pos = positive_map.sum(2) + 1e-6
+    box_to_token_loss = (
+        ((pos_term / nb_pos + neg_term)).masked_fill(~boxes_with_pos, 0).sum()
```
ankitade, line 67: why does nb_pos not divide neg_term also?

ankitade (approved):

PR #140, opened 2022-07-08
Add HF ImageNet dataset authentication to README

RdoubleA:
## Summary

Running FLAVA pre-training in examples led to authentication failure when trying to download the ImageNet dataset. For a short term solution, we've added details to the README describing how users can generate their own HuggingFace access tokens for successful authentication.

## Test plan

Ran the train script from the README: `python -m flava.train config=flava/configs/pretraining/debug.yaml` to ensure datasets were downloaded successfully and script runs.

### DIFF
--- a/examples/flava/README.md
+++ b/examples/flava/README.md
@@ -27,7 +27,7 @@ pip install -r flava/requirements.txt
 
 ### Access ImageNet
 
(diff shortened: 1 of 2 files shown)


ankitade (approved):
thanks. land after doing the sanity check and updating test plan

ankitade (approved):

PR #141, opened 2022-07-09
[MDETR] Matcher class

ebsmothers:
Summary: Added the Hungarian matcher class used in MDETR to perform
the bipartite matching between prediction and target bounding boxes.

Test plan: added a unit test

```
python -m pytest examples/mdetr/test/test_matcher.py
...
============= 2 passed in 5.75s ===============
```


examples/mdetr/matcher.py
```diff
@@ -0,0 +1,131 @@
...
+        # We flatten to compute the cost matrices in a batch
+        out_prob = F.softmax(
+            pred_logits.flatten(0, 1), dim=-1
+        )  # [batch_size * num_queries, num_classes]
+        out_bbox = pred_boxes.flatten(0, 1)  # [batch_size * num_queries, 4]
+        assert target_boxes.size(0) == positive_map.size(
+            0
+        ), "Total of target boxes should match first dim of positive map"
+
+        # Compute the soft-cross entropy between the predicted token alignment and the GT one for each box
```
langong347, line 84: `GT` refers to ? Use the full name here unless it was denoted somewhere else or convention?

examples/mdetr/matcher.py
```diff
@@ -0,0 +1,131 @@
...
+class HungarianMatcher(nn.Module):
+    """This class computes an assignment between the targets and the predictions of the network.
+
+    For efficiency reasons, the targets don't include the no_object. Because of this,
+    in general, there are more predictions than targets. In this case, we do a 1-to-1
+    matching of the best predictions, while the others are un-matched (and thus treated
+    as non-objects). This implementation is based on the MDETR repo:
+    https://github.com/ashkamath/mdetr/blob/main/models/matcher.py#L13
+
+    Params: cost_class (float): Relative weight of the classification error in the
```
langong347, line 25: `Params` --> `Attributes`
`Inputs` --> `Args`
Start a new line after those keywords too.

Add `Raises`.

https://google.github.io/styleguide/pyguide.html#384-classes

examples/mdetr/test/test_matcher.py
```diff
@@ -0,0 +1,127 @@
...
+import random
+
+import pytest
+import torch
+from examples.mdetr.matcher import HungarianMatcher
+from test.test_utils import assert_expected, set_rng_seed
+from torchvision.ops.boxes import box_convert
+
+
+@pytest.fixture(scope="class", autouse=True)
```
langong347, line 16: I suppose here you want to refresh the seed only at class level?
This means the order of the methods making use of random within a class matters. 
The latter can cause some surprise later on, e.g., when someone wants to add a new method with randomness into the same class.

examples/mdetr/matcher.py
```diff
@@ -0,0 +1,131 @@
...
+from typing import List, Tuple
+
+import torch
+import torch.nn.functional as F
+from scipy.optimize import linear_sum_assignment
+from torch import nn, Tensor
+from torchvision.ops.boxes import box_convert, generalized_box_iou
+
+
+class HungarianMatcher(nn.Module):
```
langong347, line 16: Why putting this in examples not in `torchmultimodal/modules/loss/`?

examples/mdetr/matcher.py
ebsmothers: Oh it means ground truth, but agree that the shortening here is unnecessary

examples/mdetr/matcher.py
ebsmothers: It is pretty specific to MDETR in that it is performing a bipartite matching between predicted bounding boxes and ground truth bounding boxes. So it is not really a loss per se, even though it is used as part of the final loss calculation. Also, we already have to put one of the MDETR losses in examples (see #138), so we do not lose anything by having this under examples.

examples/mdetr/matcher.py
```diff
@@ -0,0 +1,131 @@
...
+        return [
+            (
+                torch.as_tensor(i, dtype=torch.int64),
+                torch.as_tensor(j, dtype=torch.int64),
+            )
+            for i, j in indices
+        ]
+
+
+def build_matcher(args):
```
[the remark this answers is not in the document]
ebsmothers: On second thought I will just remove this altogether for now, it is not adding much value as we can equivalently instantiate the class directly

examples/mdetr/matcher.py
```diff
@@ -0,0 +1,131 @@
...
+            target_boxes_per_sample (List[Tensor]): A list of target bounding boxes.
+                Length = batch_size.
+                Each element is a tensor of size (n_boxes_for_sample, 4).
+            positive_map (Tensor): positive_map[i,j] = 1 when box i maps to class j.
+                Size: (total_boxes, num_classes) where total_boxes is the sum of
+                n_boxes_for_sample over every sample in the batch.
+
+    Returns:
+            A list of size batch_size, containing tuples of (index_i, index_j) where:
+                - index_i is the indices of the selected predictions (in order)
```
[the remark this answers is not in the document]
ebsmothers: Leaving out the math formatting for the len(index_i) = len(index_j) ... equation because ultimately I will just wind up wrapping almost everything in \text anyways 😅

langong347 (approved):
LGTM. Q: should we add a CI workflow to run tests under `examples` that only gets triggered during PR time + when there're changes made to `examples`? We can discuss during standup / weekly meeting.

PR #146, opened 2022-07-13
[MDETR] Add VQA loss module

ebsmothers:
Summary: This commit adds a module for MDETR's VQA loss.
We combine the cross-entropy loss with the linear layers which
were in the main MDETR model class in the original repo. The VQA
loss has two components: loss for predicting the answer type, and
filtered losses within each answer type. We use masking to avoid
extra computations, and return a dictionary of all the different losses.

Test plan: Added unit test

```
python -m pytest -v test/modules/losses/test_mdetr_losses.py
...
============== 4 passed in 4.38s =====================
```

### DIFF
--- a/test/modules/losses/test_mdetr_losses.py
+++ b/test/modules/losses/test_mdetr_losses.py
@@ -9,7 +9,12 @@ import random
 import pytest
 import torch
 from test.test_utils import assert_expected, set_rng_seed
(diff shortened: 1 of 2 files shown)


ebsmothers:
test

ebsmothers:
testing

PR #147, opened 2022-07-13
[MUGEN] Add caching to MultiheadAttention

langong347:
* (to be filled)

### DIFF
--- a/test/modules/layers/test_attention.py
+++ b/test/modules/layers/test_attention.py
@@ -17,6 +17,10 @@ from torchmultimodal.modules.layers.attention import (
 )
 
 
+# TODO: Port the entire script to pytest
+# TODO: Split into tests by attention module
+
+
 class TestAttention(unittest.TestCase):
     """
     Test all Attention classes
@@ -47,6 +51,17 @@ class TestAttention(unittest.TestCase):
             causal=False,
             attn_module=self.full,
         )
+        """
+        self.mha_causal = MultiHeadAttention(
+            self.input_shape,
+            self.hidden_dim,
+            self.hidden_dim,
+            self.n_heads,
+            1,
+            causal=True,
+            attn_module=self.full,
+        )
+        """
         self.block = AxialAttentionBlock(
             len(self.input_shape), self.hidden_dim, self.n_heads
         )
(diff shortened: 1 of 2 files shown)


test/modules/layers/test_attention.py
```diff
@@ -17,164 +19,300 @@
...
+
+@pytest.fixture
+def kv(n_heads, input_shape, hidden_dim):
+    return torch.randn(1, n_heads, *input_shape, hidden_dim // n_heads)
+
+
+@pytest.fixture
+def full_attn(input_shape):
+    # TODO: retire causal once mask generation is moved out of FullAttention
+    #   causal inside FullAttention does not affect caching of k, v
```
RdoubleA, line 60: nit: move this TODO comment outside of test and in FullAttention for easier tracking?

test/modules/layers/test_attention.py
```diff
@@ -17,164 +19,300 @@
...
+            return MultiHeadAttention(
+                input_shape, hidden_dim, hidden_dim, n_heads, 1, causal, attn_module
+            )
+
+        return create_multihead_attn
+
+    def test_split_multihead(self, input_shape, multihead_attn, full_attn):
+        mha = multihead_attn(False, full_attn)
+        x = torch.randn(1, *input_shape, 6)  # (b, d1, ..., dn, c)
+        mha.n_head = 2
```
RdoubleA, line 82: you could make `n_head` a parameter in `create_multihead_attn` so you can remove fixture and you don't have to modify it here

torchmultimodal/modules/layers/attention.py
```diff
@@ -131,63 +137,50 @@ def __init__(
...
-
-        a = self.attn(q, k, v, decode_step, decode_idx)
+                    # append present k, v to past k, v
+                    # for autoregressive decoding inputs are flattened as 1D sequences
+                    # so are the cached tensors: (b, n_heads, seq_len, c)
+                    k_, v_ = self.cache["k"], self.cache["v"]
+                    self.cache["k"] = torch.cat([k_, k], dim=2)
+                    self.cache["v"] = torch.cat([v_, v], dim=2)
+                # override the present k, v with the cache
+                k, v = self.cache["k"], self.cache["v"]
```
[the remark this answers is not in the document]
langong347: In the non-causal case, `k` and `v` will be on the start the entire sequence. We should just compute their projection once (this logic is missing in MUGEN) and cache them once to be used later.

In the causal case, `k` and `v` will be step-wise so the cache will be incremental.

torchmultimodal/modules/layers/attention.py
```diff
@@ -224,17 +217,14 @@ def forward(
         q: Tensor,
         k: Tensor,
         v: Tensor,
-        decode_step: Optional[int] = None,
-        decode_idx: Optional[Iterable[int]] = None,
     ) -> Tensor:
         mask = torch.Tensor(self.mask) if self.causal else None
```
RdoubleA, line 221: The mask is already created for you when you're decoding. what about the case when a user specifies an attention mask during encoding? I am thinking of making `FullAttention` the base self-attention class for all models. But in FLAVA, ALBEF there is attention masking used in the encoders. how do we reconcile this?

langong347: By "mask created while decoding" Are you referring to 

https://github.com/facebookresearch/multimodal/blob/e274207e5c091d681bcb790d561fca0764148549/torchmultimodal/modules/layers/attention.py#L231-L232

?

I think mask should be created by a top level module (either vanilla transformer or that in the decoding mode) and passed in to the attention modules instead of being created inside the attention module which is not only inflexible but incurring the additional argument `shape`.

Is this consistent with "user specifies an attention mask during encoding"?

RdoubleA: Yes I agree with that, I think it's better design and it's consistent with other models. FLAVA passes in the attention mask at the highest level at `FLAVAModel`'s forward, for example. I can move the mask out of the attention modules in a separate PR for unifying `SelfAttention` and `MultiHeadAttention`

RdoubleA (approved):

torchmultimodal/modules/layers/attention.py
langong347: Feel free to move the mask out of the attention module. I was thinking about doing that previously as the transformer decoder layers will depend on that.

langong347:
@langong347 has imported this pull request.  If you are a Facebook employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D37865352).

PR #151, opened 2022-07-13
[Unification] Unify attention utils

RdoubleA:
* #172
* #170
* #169
* #168
* #162
* #152
* __->__ #151

## Summary
Small diffs building up to a unified `SelfAttention` class used across all models by generalizing one piece at a time.

Replaces `transpose_for_scores` and shape permutations in `FLAVASelfAttention` and `ALBEFTransformerSelfAttention` with `split_multihead`, 'merge_multihead`.

## Test plan
`pytest test -vv`
`pytest examples -vv`
[Notebook](https://colab.research.google.com/drive/1qc0njVryPWuzCsEaaVupMzU1pcXwZjhG?usp=sharing) with FLAVA checkpoint loading functions. Tested checkpoints on random inputs to ensure they align with previous version before unification and that loading is successful. A separate [PR](https://github.com/facebookresearch/multimodal/pull/161) adds this check as an optional test.

Differential Revision: [D37929520](https://our.internmc.facebook.com/intern/diff/D37929520)

### DIFF
--- a/torchmultimodal/modules/layers/transformer.py
+++ b/torchmultimodal/modules/layers/transformer.py
@@ -13,9 +13,12 @@ from typing import Any, Callable, Optional, Tuple
 
 import torch
 from torch import nn, Tensor
-from torchmultimodal.modules.layers.attention import scaled_dot_product_attention
+from torchmultimodal.modules.layers.attention import (
+    merge_multihead,
+    scaled_dot_product_attention,
+    split_multihead,
+)
 from torchmultimodal.modules.layers.normalizations import Fp32LayerNorm
-from torchmultimodal.utils.common import transpose_for_scores
 
 FLAVATransformerOutput = namedtuple(
     "FLAVATransformerOutput",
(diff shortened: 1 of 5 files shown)


torchmultimodal/modules/layers/transformer.py
```diff
@@ -83,9 +81,7 @@ def forward(
             self.attn_dropout,
         )
 
-        context_layer = context_layer.permute(0, 2, 1, 3).contiguous()
```
langong347, line 86: non blocking Q: This `contiguous` call is replaced by `make_contiguous` option in `shift_dim`. But in general calling `contiguous` after `permute` may not be necessary:

https://discuss.pytorch.org/t/call-contiguous-after-every-permute-call/13190/3

I guess we can still default `make_contiguous` as `True` just to be safe.

langong347 (approved):

torchmultimodal/modules/layers/transformer.py
RdoubleA: I kept it because I was previously using `view` instead of `flatten` and `unflatten` right after `shift_dim` and I ran into errors if I didn't make contiguous. It may not be necessary anymore but I would lean towards keeping to be safe, unless we find a clear drawback in doing so.

torchmultimodal/utils/common.py
```diff
@@ -121,13 +121,6 @@ def tensor_slice(x: Tensor, begin: List[int], size: List[int]) -> Tensor:
     return x[slices]
 
 
-def transpose_for_scores(
-    num_attention_heads: int, attention_head_size: int, x: Tensor
-) -> Tensor:
-    x = x.unflatten(-1, (num_attention_heads, attention_head_size))
-    return x.permute(0, 2, 1, 3)
```
ankitade, line 128: why do we need to replace the simple permute logic with the shift dim logic?

RdoubleA: shift_dim has built in error checking, allows use of negative dimensions, and calls `contiguous()` after permuting which is important if you call `view()` right after. so although it's replacing a single line with slightly more code it's a lot more flexible to use.

RdoubleA:
@RdoubleA has imported this pull request.  If you are a Facebook employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D37929520).

RdoubleA:
@RdoubleA has imported this pull request.  If you are a Facebook employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D37929520).

PR #152, opened 2022-07-13
[Unification] Move get_extended_attention_mask to utils

RdoubleA:
* #172
* #170
* #169
* #168
* #162
* __->__ #152
* #151

## Summary
Small diffs building up to a unified `SelfAttention` class used across all models by generalizing one piece at a time.

This diff moves `get_extended_attention_mask` from FLAVA and ALBEF into utils.

## Test plan
`pytest test -vv`
`pytest examples -vv`
[Notebook](https://colab.research.google.com/drive/1qc0njVryPWuzCsEaaVupMzU1pcXwZjhG?usp=sharing) with FLAVA checkpoint loading functions. Tested checkpoints on random inputs to ensure they align with previous version before unification and that loading is successful. A separate [PR](https://github.com/facebookresearch/multimodal/pull/161) adds this check as an optional test.

Differential Revision: [D37935055](https://our.internmc.facebook.com/intern/diff/D37935055)

### DIFF
--- a/torchmultimodal/utils/attention.py
+++ b/torchmultimodal/utils/attention.py
@@ -1,0 +1,33 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+from torch import Tensor
+
+
+def get_extended_attention_mask(attention_mask: Tensor) -> Tensor:
+    """
+    Makes broadcastable attention and causal masks so that future and masked tokens are ignored.
+
+    Args:
+        attention_mask (Tensor): Mask with ones indicating tokens to attend to, zeros for tokens to ignore.
+    Returns:
+        extended_attention_mask (Tensor): extended attention mask with the same dtype as attention_mask.dtype.
+    """
+
+    if attention_mask.dim() == 3:
+        extended_attention_mask = attention_mask[:, None, :, :]
+    elif attention_mask.dim() == 2:
+        extended_attention_mask = attention_mask[:, None, None, :]
+    else:
+        raise ValueError(
+            "Wrong shape for attention_mask (shape {})".format(attention_mask.shape)
+        )
+
+    extended_attention_mask = extended_attention_mask.to(
+        dtype=attention_mask.dtype
+    )  # fp16 compatibility
+
+    return extended_attention_mask
--- a/torchmultimodal/models/flava/flava_text_encoder.py
+++ b/torchmultimodal/models/flava/flava_text_encoder.py
@@ -5,11 +5,11 @@
 # LICENSE file in the root directory of this source tree.
 
 from functools import partial
-from typing import Any, Callable, Optional, Tuple
+from typing import Any, Callable, Optional
 
 import torch
 from packaging import version
-from torch import device, nn, Tensor
+from torch import nn, Tensor
 from torchmultimodal.modules.layers.normalizations import Fp32LayerNorm
 from torchmultimodal.modules.layers.transformer import (
     FLAVATransformerEncoder,
@@ -17,6 +17,7 @@ from torchmultimodal.modules.layers.transformer import (
     init_transformer_weights,
 )
 from torchmultimodal.modules.losses.flava import Pooler
+from torchmultimodal.utils.attention import get_extended_attention_mask
 
 
 class TextEmbeddings(nn.Module):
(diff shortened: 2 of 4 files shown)


torchmultimodal/utils/attention.py
```diff
@@ -0,0 +1,33 @@
...
+    """
+    Makes broadcastable attention and causal masks so that future and masked tokens are ignored.
+
+    Args:
+        attention_mask (Tensor): Mask with ones indicating tokens to attend to, zeros for tokens to ignore.
+    Returns:
+        extended_attention_mask (Tensor): extended attention mask with the same dtype as attention_mask.dtype.
+    """
+
+    if attention_mask.dim() == 3:
```
langong347, line 20: In each case, add comments for: 
1) original shape 
2) extended shape 
?
The original comments you left out seem to be right on point?

torchmultimodal/utils/attention.py
```diff
@@ -0,0 +1,33 @@
...
+    elif attention_mask.dim() == 2:
+        extended_attention_mask = attention_mask[:, None, None, :]
+    else:
+        raise ValueError(
+            "Wrong shape for attention_mask (shape {})".format(attention_mask.shape)
+        )
+
+    extended_attention_mask = extended_attention_mask.to(
+        dtype=attention_mask.dtype
+    )  # fp16 compatibility
```
langong347, line 31: Do you know why after adding `None` the data type might change (such that here it's forced back to that of the input attn mask)? 

Is it related to the comment "fp16 compatibility"?

langong347 (approved):

torchmultimodal/utils/attention.py
RdoubleA: oh, I didn't notice that the FLAVA version had more comments

torchmultimodal/utils/attention.py
RdoubleA: I am not sure why this is needed, the default data type for torch tensors is `float32`. I supposed if you passed in an attention mask as fp16 this would preserve it, but throughout the code base I see attention_mask be instantiated with `float32` ([here](https://github.com/facebookresearch/multimodal/blob/df2f952d7751bf6517ee85686802a72a8b5d432e/torchmultimodal/models/flava/flava_text_encoder.py#L180) or even an `IntTensor` ([here](https://github.com/facebookresearch/multimodal/blob/df2f952d7751bf6517ee85686802a72a8b5d432e/test/models/flava/test_flava_text_encoder.py#L79)). 

There is also the `Fp32LayerNorm` in the normalization folder, which I'm not sure is really doing anything different than `nn.LayerNorm` since weights and biases should already be `float32` by default.

torchmultimodal/utils/attention.py
```diff
@@ -0,0 +1,33 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
```
ankitade, line 1: why do we need a attention.py under utils? just add these to the existing attention module?

torchmultimodal/utils/attention.py
ankitade: probably needed for mixed precision training

torchmultimodal/utils/attention.py
RdoubleA: I envisioned `modules/layers/attention.py` to contain primarily `nn.Modules`/classes intended to be used as layer/components/buildling blocks for models. Utility functions such as `get_extended_attention_mask` seemed to belong more in the `utils/` folders as opposed to `modules/layers/attention.py`.

torchmultimodal/utils/common.py
```diff
@@ -24,16 +24,21 @@ def get_current_device() -> Union[str, torch.device]:
...
     Makes broadcastable attention and causal masks so that future and masked tokens are ignored.
+    Arguments:
+        attention_mask (`torch.Tensor`):
+            Mask with ones indicating tokens to attend to, zeros for tokens to ignore.
 
-    Args:
-        attention_mask (Tensor): Mask with ones indicating tokens to attend to, zeros for tokens to ignore.
     Returns:
-        extended_attention_mask (Tensor): extended attention mask with the same dtype as attention_mask.dtype.
+        `torch.Tensor` The extended attention mask, with the same dtype as `attention_mask.dtype`.
```
langong347, line 32: Double back tick for code sample inline expansion. Single back tick has reserved meaning in reStructuredText.
"Tensor" instead of "torch.Tensor"?

torchmultimodal/utils/attention.py
langong347: If the idea is to enforce fp16 for mixed precision training, then should we set `dtype` equal to `fp16` instead of inheriting that of the input attention_mask?

PR #158, opened 2022-07-15
videoCLIP lightning module for evaluation

sophiazhi:
Summary:
Add pytorch lightning module for videoCLIP evaluation.
(Recall metrics will be added in a different PR.)

Test plan:
First, made a mini dataset from MUGEN's test dataset.
Test/example in [notebook](https://colab.research.google.com/drive/1aWX4dxJd4HBiz9QYpzGsqI5evI5VxwNJ)

### DIFF
--- a/examples/mugen/retrieval/model.py
+++ b/examples/mugen/retrieval/model.py
@@ -1,0 +1,51 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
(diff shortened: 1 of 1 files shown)


langong347 (approved):
Looks good as a first step. Eventually, we should have the full lightning module with the training related logic, `train.py`, `eval.py` and the datamodule altogether. That means incorporate some content from the notebook.

PR #174, opened 2022-07-19
[ALBEF] vqa finetune

katrina433:
Summary:

ALBEF VQA finetuning logic draft

`examples/albef/model.py`: has the training model for vqa task
`examples/albef/finetune.py`: has the methods for training and validation

Test plan:

Ran the training validation loops (saw decreasing loss and improved outputs)

### DIFF
--- a/examples/albef/finetune.py
+++ b/examples/albef/finetune.py
@@ -1,0 +1,197 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+import argparse
+import datetime
+import os
+import random
+import time
+
+import examples.utils.distributed as dist_utils
+
+import ruamel.yaml as yaml
+import torch
+import torch.backends.cudnn as cudnn
+from examples.albef.data.vqa_datamodules import VQADataModule
+from examples.albef.model import albef_model_for_vqa
+from examples.albef.utils import add_weight_decay
+from torch.optim import AdamW
+from torch.optim.lr_scheduler import CosineAnnealingWarmRestarts
+
+
+def train(model, datamodule, args, checkpoint_root, device):
+    model_without_ddp = (
+        model.module if dist_utils.is_dist_avail_and_initialized() else model
+    )
+    model.train()
+
+    optimizer_params = add_weight_decay(model, args["weight_decay"])
+    optimizer = AdamW(optimizer_params, lr=args["lr"])
+    scheduler = CosineAnnealingWarmRestarts(
+        optimizer, T_0=args["max_epochs"], eta_min=args["min_lr"]
+    )
+
+    step_size = args["step_size"]
+    warmup_steps = args["warmup_steps"]
+    warmup_iterations = warmup_steps * step_size
+
+    data_loader = datamodule.train_dataloader(
+        is_distributed=dist_utils.is_dist_avail_and_initialized(),
+        num_tasks=dist_utils.get_world_size(),
+        global_rank=dist_utils.get_rank(),
+    )
+
+    start_time = time.time()
+    log_every_n_steps = 100
+    dataset_length = len(datamodule.train_dataset)
+
+    for epoch in range(args["max_epochs"]):
+        if epoch > 0:
+            scheduler.step(epoch + warmup_steps)
+
+        for batch, (
+            images,
+            questions,
+            questions_atts,
+            answers,
+            answers_atts,
+            ans_weights,
+            ans_lengths,
+        ) in enumerate(data_loader):
+            if epoch > 0:
+                alpha = args["alpha"]
+            else:
+                alpha = args["alpha"] * min(1, batch / dataset_length)
+
+            images = images.to(device, non_blocking=True)
+            questions = questions.to(device)
+            questions_atts = questions_atts.to(device)
+            answers = answers.to(device)
+            answers_atts = answers_atts.to(device)
+            ans_weights = ans_weights.to(device)
+
+            loss = model(
+                images,
+                questions,
+                questions_atts,
+                answers,
+                answers_atts,
+                ans_weights=ans_weights,
+                ans_lengths=ans_lengths,
+                alpha=alpha,
+                is_train=True,
+            )
+
+            optimizer.zero_grad()
+            loss.backward()
+            optimizer.step()
+
+            if epoch == 0 and batch % step_size == 0 and batch <= warmup_iterations:
+                scheduler.step(batch // step_size)
+
+            if batch % log_every_n_steps == 0:
+                total_time = time.time() - start_time
+                time_str = "time {},".format(
+                    datetime.timedelta(seconds=int(total_time))
+                )
+                epoch_str = "epoch {}/{},".format(epoch, args["max_epochs"])
+                batch_str = "batch {}/{},".format(batch, len(data_loader))
+                loss_str = "loss {}".format(loss.item())
+                print(time_str, epoch_str, batch_str, loss_str)
+
+        save_obj = {
+            "model": model_without_ddp.state_dict(),
+            "optimizer": optimizer.state_dict(),
+            "lr_scheduler": scheduler.state_dict(),
+            "epoch": epoch,
+        }
+        torch.save(
+            save_obj, os.path.join(checkpoint_root, "vqa_checkpoint_%02d.pth" % epoch)
+        )
+
+
+@torch.no_grad()
+def evaluation(model, datamodule, k, device):
+    model.eval()
+
+    result = []
+
+    answer_list = datamodule.test_dataset.answer_list
+    answer_input_ids = datamodule.test_dataset.answer_input_ids.to(device)
+    answer_atts = datamodule.test_dataset.answer_attention_mask.to(device)
+    data_loader = datamodule.test_dataloader(
+        is_distributed=dist_utils.is_dist_avail_and_initialized(),
+        num_tasks=dist_utils.get_world_size(),
+        global_rank=dist_utils.get_rank(),
+    )
+
+    start_time = time.time()
+    log_every_n_steps = 100
+
+    for batch, (img, ques, ques_atts, ques_ids) in enumerate(data_loader):
+        img = img.to(device, non_blocking=True)
+        ques = ques.to(device)
+        ques_atts = ques_atts.to(device)
+
+        topk_ids, topk_probs = model(
+            img, ques, ques_atts, answer_input_ids, answer_atts, k=k, is_train=False
+        )
+
+        for ques_id, topk_id, topk_prob in zip(ques_ids, topk_ids, topk_probs):
+            _, pred = topk_prob.max(dim=0)
+            result.append(
+                {"question_id": ques_id, "answer": answer_list[topk_id[pred]]}
+            )
+
+        if batch % log_every_n_steps == 0:
+            total_time = time.time() - start_time
+            total_time_str = str(datetime.timedelta(seconds=int(total_time)))
+            print(
+                "time {}, batch {}/{}".format(total_time_str, batch, len(data_loader))
+            )
+
+    return result
+
+
+def main():
+    parser = argparse.ArgumentParser()
+    parser.add_argument("--config", default="./examples/albef/configs/vqa.yaml")
+    parser.add_argument("--device", default="cuda")
+    parser.add_argument(
+        "--world_size", default=1, type=int, help="number of distributed processes"
+    )
+    parser.add_argument(
+        "--dist_url", default="env://", help="url used to set up distributed training"
+    )
+    args = parser.parse_args()
+    config = yaml.load(open(args.config, "r"), Loader=yaml.Loader)
+
+    dist_utils.init_distributed_mode(args)
+    device = torch.device(args.device)
+
+    seed = config["seed"] + dist_utils.get_rank()
+    torch.manual_seed(seed)
+    random.seed(seed)
+    cudnn.benchmark = True
+
+    datamodule = VQADataModule(**config["datamodule_args"])
+    model = albef_model_for_vqa(config)
+
+    pretrained_checkpoint_path = os.path.join(
+        config["checkpoint_root"], config["pretrained_checkpoint"]
+    )
+    model.load_state_dict(torch.load(pretrained_checkpoint_path, map_location="cpu"))
+    model = model.to(device)
+    if args.distributed:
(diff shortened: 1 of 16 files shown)


torchmultimodal/modules/losses/albef.py
```diff
@@ -127,106 +127,56 @@ class MaskedLanguageModelingLoss(nn.Module):
...
-        )
 
     def forward(
         self,
         labels: Tensor,
-        hidden_states: Tensor,
-        hidden_states_m: Optional[Tensor] = None,
+        prediction_scores: Tensor,
+        prediction_scores_m: Optional[Tensor] = None,
+        alpha: Optional[float] = 0.0,
```
RdoubleA, line 154: what was the reason for moving alpha outside of the `__init__`? seems more suitable for an attribute

torchmultimodal/utils/attention.py
```diff
@@ -26,8 +29,13 @@ def get_extended_attention_mask(attention_mask: Tensor) -> Tensor:
         extended_attention_mask = attention_mask[:, None, :, :]
     elif attention_mask.dim() == 2:
         # Provided a padding mask of dimensions [batch_size, seq_length]
-        # - if the model is an encoder, make the mask broadcastable to [batch_size, num_heads, seq_length, seq_length]
-        extended_attention_mask = attention_mask[:, None, None, :]
+        if is_decoder:
+            extended_attention_mask = get_extended_attention_mask_for_decoder(
```
RdoubleA, line 33: I would just keep these two as separate functions and remove this logic, I don't see a reason to tie them together like this

examples/albef/model.py
```diff
@@ -0,0 +1,385 @@
...
+            for b, n in enumerate(ans_lengths):
+                encoder_hidden_states_m += [multimodal_embeds_m[b]] * n
+            encoder_hidden_states_m = torch.stack(encoder_hidden_states_m)
+
+        return encoder_hidden_states, encoder_hidden_states_m, encoder_attention_mask
+
+
+def albef_vision_encoder(
+    hidden_size: int,
+    args: dict,
```
ebsmothers, line 281: Just curious, why do we treat hidden_size differently than all the other args in these builders?

examples/albef/model.py
```diff
@@ -0,0 +1,385 @@
...
+        loss: MaskedLanguageModelingLoss,
+    ) -> None:
+        super().__init__()
+        self.model = model
+        self.answer_decoder = answer_decoder
+        self.loss = loss
+        self.answer_decoder_m = copy.deepcopy(self.answer_decoder)
+        remove_grad(self.answer_decoder_m)
+
+    def forward(
```
ebsmothers, line 107: nit: I think it's common to put forward method last in the class body

examples/albef/model.py
```diff
@@ -0,0 +1,385 @@
...
+        question: Tensor,
+        question_atts: Tensor,
+        answers: Tensor,
+        answers_atts: Tensor,
+        ans_weights: Optional[Tensor] = None,
+        ans_lengths: Optional[List[int]] = None,
+        alpha: Optional[float] = 0.0,
+        k: Optional[int] = 128,
+        is_train: Optional[bool] = True,
+    ) -> Union[Tensor, Tuple[Tensor, Tensor]]:
```
ebsmothers, line 119: You could also return a named tuple with optional fields to make the return type a bit cleaner

examples/albef/model.py
```diff
@@ -0,0 +1,385 @@
...
+            encoder_attention_mask=encoder_attention_mask,
+        )
+
+        labels = answers.masked_fill(answers == 0, -100)
+        loss = self.loss(labels, prediction_scores, prediction_scores_m, alpha)
+        loss = ans_weights * loss
+        loss = loss.sum() / image.size(0)
+        return loss
+
+    def _test_forward(
```
ebsmothers, line 188: nit: maybe call this _inference_forward or something

examples/albef/model.py
```diff
@@ -0,0 +1,385 @@
...
+        )
+
+        answer_loss = self.loss(targets_ids, prediction_scores)
+        answer_loss = answer_loss.view(input_ids.size(0), -1)
+
+        # topk_prob: first token probability
+        topk_probs = topk_probs.view(-1, 1)
+        log_probs = torch.cat([topk_probs.log(), -answer_loss], dim=1)
+
+        # re-calculate log probabilities for the answer sequences using chain rule
```
ebsmothers, line 247: Not sure I understand this comment. Is this just about the product of conditional probabilities in the answer sequence?

torchmultimodal/modules/losses/albef.py
katrina433: The training loop passes in different `alpha` values during warm up so I thought it makes sense for it to be part of the func call. Another option is to keep it in `__init__` and change `model.loss.alpha` in the training loop. Which one is more preferable?

examples/albef/finetune.py
```diff
@@ -0,0 +1,146 @@
...
+
+    return result
+
+
+def main():
+    seed = 42
+    torch.manual_seed(seed)
+    np.random.seed(seed)
+    random.seed(seed)
+    cudnn.benchmark = True
```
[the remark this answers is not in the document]
katrina433: This finds the optimal algorithm for the hardware when the input sizes stay constant so it speeds up the model.

examples/albef/utils.py
```diff
@@ -0,0 +1,23 @@
...
+
+from torch import nn
+
+
+def add_weight_decay(model: nn.Module, weight_decay: float = 1e-5) -> None:
+    decay = []
+    no_decay = []
+    for name, param in model.named_parameters():
+        if not param.requires_grad:
+            continue  # frozen weights
```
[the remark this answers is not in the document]
katrina433: This skips weight decay for the momentum models

examples/albef/utils.py
```diff
@@ -0,0 +1,23 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+from torch import nn
+
+
+def add_weight_decay(model: nn.Module, weight_decay: float = 1e-5) -> None:
```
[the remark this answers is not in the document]
katrina433: Should I actually remove the default value for `weight_decay` since a different value is passed in from the training loop?

examples/albef/model.py
katrina433: Other args are encoder-specific, but `hidden_size` should be the same for the 3 encoders so the hidden states can be passed around.

torchmultimodal/modules/losses/albef.py
RdoubleA: ok that makes sense, let's keep it in forward

RdoubleA (commented):
This is quite the heavy lift - props to you for getting this all in. In the future, I would suggest breaking these diffs down smaller as much as you can so it's easier to find any issues, even if it is all related to one task :)

examples/albef/model.py
```diff
@@ -0,0 +1,385 @@
...
+            )
+
+        prediction_scores = self.answer_decoder(
+            input_ids=answers,
+            attention_mask=answers_atts,
+            encoder_hidden_states=encoder_hidden_states,
+            encoder_attention_mask=encoder_attention_mask,
+        )
+
+        labels = answers.masked_fill(answers == 0, -100)
```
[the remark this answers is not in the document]
katrina433: The value just has to match the `mask_token_id` used in the loss module. I updated this to directly pass in the value from the loss module.

examples/albef/model.py
```diff
@@ -0,0 +1,385 @@
...
+        loss = self.loss(labels, prediction_scores, prediction_scores_m, alpha)
```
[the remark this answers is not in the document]
katrina433: the ALBEFModel is also finetuned

torchmultimodal/utils/attention.py
```diff
@@ -38,3 +41,34 @@ def get_extended_attention_mask(attention_mask: Tensor) -> Tensor:
     )  # fp16 compatibility
 
     return extended_attention_mask
+
+
+def get_extended_attention_mask_for_decoder(attention_mask: Tensor) -> Tensor:
```
ebsmothers, line 46: Can you add some more detail on what this function is doing? It seems quite a bit more complicated than get_extended_attention_mask, which is really just broadcasting the existing mask to extra dimensions

examples/albef/finetune.py
```diff
@@ -0,0 +1,199 @@
...
+            total_time_str = str(datetime.timedelta(seconds=int(total_time)))
+            print(
+                "time {}, batch {}/{}".format(total_time_str, batch, len(data_loader))
+            )
+
+    return result
+
+
+def main():
+    seed = 42
```
ebsmothers, line 161: Do you need to set a separate seed per rank?

examples/albef/model.py
```diff
@@ -0,0 +1,385 @@
...
+
+    def forward(self, hidden_states: Tensor) -> Tensor:
+        hidden_states = self.dense(hidden_states)
+        hidden_states = self.transform_act_fn(hidden_states)
+        hidden_states = self.layer_norm(hidden_states)
+        hidden_states = self.decoder(hidden_states)
+        return hidden_states
+
+
+class ALBEFDecoder(nn.Module):
```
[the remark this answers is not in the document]
ankitade: but why are the models in example though?

examples/albef/model.py
```diff
@@ -0,0 +1,357 @@
...
+    """
+    Generate the prediction scores for answers from image and question hidden states.
+
+    Args:
+        text_embeddings (ALBEFTextEmbeddings): Instantiated ALBEFTextEmbeddings.
+        multimodal_decoder (ALBEFMultimodalEncoder): Instantiated ALBEFMultimodalEncoder.
```
ankitade, line 71: this part is confusing to lay man. why is encoder being passed to decoder? please add some explanation in comments

examples/albef/model.py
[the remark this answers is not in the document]
katrina433: The models in examples are used for VQA only (since other tasks do not use a decoder). They just provide an example of how to put albef model, encoders, and losses together for a downstream task.

examples/albef/model.py
katrina433: Originally there was a flag that turns the multimodal encoder to a decoder. But now the functionality is refactored out from the class. I renamed the variable name to multimodal_encoder to avoid confusion.

examples/albef/finetune.py
```diff
@@ -0,0 +1,198 @@
...
+            )
+
+    return result
+
+
+def main():
+    parser = argparse.ArgumentParser()
+    parser.add_argument("--config", default="./examples/albef/configs/vqa.yaml")
+    parser.add_argument("--device", default="cuda")
+    parser.add_argument("--distributed", default=True, type=bool)
```
sophiazhi, line 163: fyi I found [this warning](https://docs.python.org/3/library/argparse.html#type:~:text=The%20bool()%20function%20is%20not%20recommended%20as%20a%20type%20converter.%20All%20it%20does%20is%20convert%20empty%20strings%20to%20False%20and%20non%2Dempty%20strings%20to%20True.%20This%20is%20usually%20not%20what%20is%20desired.) against using type=bool. An alternative is to use `action='store_true'` or `action='store_false'`

examples/albef/model.py
```diff
@@ -0,0 +1,361 @@
...
+    Args:
+        vocab_size (int): The number of different tokens the prediction_head can predict.
+        hidden_size (int): The hidden size of the prediction_head.
+        layer_norm_eps (float): The epsilon used by the prediction_head normalization layer.
+        transform_act_fn (Callable[[Tensor], Tensor]): The activation function in the prediction_head.
+
+    Inputs:
+        hidden_states (Tensor): The hidden states of preceding tokens.
+
+    Return:
```
langong347, line 40:
```suggestion
    Returns:
```

torchmultimodal/utils/attention.py
```diff
@@ -25,11 +25,13 @@ def get_extended_attention_mask(attention_mask: Tensor) -> Tensor:
     # We can provide a self-attention mask of dimensions [batch_size, from_seq_length, to_seq_length]
     # ourselves in which case we just need to make it broadcastable to all heads,
     # [batch_size, num_heads, from_seq_length, to_seq_length].
-    if attention_mask.dim() == 3:
+    if attention_mask.dim() == 4:
```
langong347, line 28: I'm not sure if we need this. I suppose people who use this function are aware whether the original masks have the right dims or not.

torchmultimodal/utils/attention.py
```diff
@@ -43,6 +45,49 @@ def get_extended_attention_mask(attention_mask: Tensor) -> Tensor:
     return extended_attention_mask
 
 
+def get_extended_attention_mask_for_decoder(attention_mask: Tensor) -> Tensor:
```
langong347, line 48: This seems to be a very specific use case where you want to derive the causal mask from the padding mask.
Instead, one can also create the causal mask "from first principle" by knowing the sequence lengths and other dimensions, right?
Also, I would look at how we can unify with `get_causal_attention_mask`.

test/models/test_albef.py
```diff
@@ -30,11 +29,10 @@ def vision_encoder():
     return ALBEFVisionEncoder(
         image_size=4,
         patch_size=4,
-        num_layers=2,
-        num_heads=1,
-        hidden_dim=3,
+        num_hidden_layers=2,
```
langong347, line 32: Unrelated to this PR: I think the typical way of using `autouse` is to automatically enable every test method to use the fixture without actually requesting it in their arguments. But here you are still requesting `vision_encoder` in your test method (e.g., `albef_model`) which defeats the purpose of `autouse`.

Do you want the seed or the encoder object to be autoused? If the former, you'd want to create a fixture for the seed with `autouse` and another for the encoder without `autouse`.

langong347 (commented):
1.

examples/albef/finetune.py
katrina433: Thanks for the info! I actually reset this arg in a separate method so I removed it from the command line args.

torchmultimodal/utils/attention.py
katrina433: This is added because in the decoder case, `get_extended_attention_mask_for_decoder` is called at a higher stack level, so `get_extended_attention_mask` will essentially be a noop.

test/models/test_albef.py
katrina433: For some reason I have to set seed in the same method I initialize the vision encoder, or else the vision encoder output is not deterministic every every I run the tests.

PR #176, opened 2022-07-19
[MUGEN] Add checkpointing wrapper

langong347:
* #205
* #213
* #204
* __->__ #176

Summary:
- Utility to wrap a function with PT Core's [checkpointing API](https://pytorch.org/docs/stable/checkpoint.html) when in training mode.
- Checkpointing avoids saving gradient info in the forward pass and recomputes the gradient in the backward pass. 
- Wrapping computation heavy module such as Attentions with checkpointing API can speed up training.

Test Plan:
```
$ python -m pytest test/utils/test_common.py -vv
platform darwin -- Python 3.8.13, pytest-7.1.2, pluggy-1.0.0 -- /Users/langong/local/miniconda3/envs/torchmm/bin/python
cachedir: .pytest_cache
rootdir: /Users/langong/gpt_attention, configfile: pyproject.toml
plugins: mock-3.8.2, cov-3.0.0
collected 10 items

test/utils/test_common.py::test_shift_dim PASSED                                                                 [ 10%]
test/utils/test_common.py::TestTensorSlice::test_default PASSED                                                  [ 20%]
test/utils/test_common.py::TestTensorSlice::test_size_minus_one PASSED                                           [ 30%]
test/utils/test_common.py::TestTensorSlice::test_uneven_begin_size PASSED                                        [ 40%]
test/utils/test_common.py::TestTensorSlice::test_invalid_begin XFAIL (Invalid begin)                             [ 50%]
test/utils/test_common.py::TestTensorSlice::test_invalid_size XFAIL (Invalid size)                               [ 60%]
test/utils/test_common.py::TestToTupleTuple::test_int PASSED                                                     [ 70%]
test/utils/test_common.py::TestToTupleTuple::test_tuple PASSED                                                   [ 80%]
test/utils/test_common.py::TestCheckpointWrapper::test_training_mode PASSED                                      [ 90%]
test/utils/test_common.py::TestCheckpointWrapper::test_eval_model PASSED                                         [100%]

============================================= 8 passed, 2 xfailed in 1.47s ===========
```

Differential Revision: [D38283963](https://our.internmc.facebook.com/intern/diff/D38283963)


test/utils/test_common.py
```diff
@@ -70,3 +76,90 @@ def test_int(self, expected):
...
+        def _compute_grad(output):
+            output.sum().backward()
+            grad_checkpointed = {}
+            for name, param in model.named_parameters():
+                grad_checkpointed[name] = param.grad.data.clone()
+            return grad_checkpointed
+
+        return _compute_grad
+
+    def test_training_mode(self, model, inputs, compute_grad):
```
ebsmothers, line 144: So in these two tests it seems you are checking that use_cache gets disabled in train mode, but are you actually testing that the checkpointing gets applied in some way? Btw might be nice to add a quick comment explaining why the grad is the same as the model output, I was a bit confused by this at first glance.

ebsmothers (commented):
Looks good overall, just a few minor comments. Also looks like the CI is complaining about lint

test/utils/test_common.py
```diff
@@ -70,3 +76,90 @@ def test_int(self, expected):
     def test_tuple(self, expected):
         actual = to_tuple_tuple((2, 2, 2), 3, 3)
         assert actual == expected, "tuple -> tuple[tuple] incorrect"
+
+
+class TestCheckpointWrapper:
+    @pytest.fixture
+    def device(self):
+        return torch.device("cuda:0" if torch.cuda.is_available() else "cpu")
```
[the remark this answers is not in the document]
langong347: `device` should not be needed here as I was just experimenting with this to get rid of a device related warning (a bug from pytorch that will be fixed in their next release).

test/utils/test_common.py
```diff
@@ -70,3 +76,90 @@ def test_int(self, expected):
...
+                out = self.identity(out)
+                return out
+
+        return DummyModel().to(device)
+
+    @pytest.fixture
+    def inputs(self, device):
+        # Set inputs to be requires_grad = True or else it will break autograd
+        # during the backward pass,
+        # "RuntimeError: element 0 of tensors does not require grad and does not have a grad_fn"
```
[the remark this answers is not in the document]
langong347: Good question here.

- Short answer: 
Checkpointed layers run in `torch.no_grad()` mode but to compute gradients in the backward pass either the input or the output of the checkpointed layer needs to have `grad_fn`.

- Long answer: 
In order for checkpointing to compute the gradient during the backward pass, the input to the first checkpointed layer needs to have `grad_fn` so that the checkpointed layers are not disconnected from autograd. While this condition is almost guaranteed if the checkpointed layers are the middle layers, in this unit test `DummyAttention` is the first layer of the model so I had to make its inputs `requires_grad`. 

Also that checkpointing runs in `torch.no_grad()` model: if the inputs are not attached to gradient, neither will the outputs. In that case, the "RuntimeError: element 0 of tensors does not require grad and does not have a grad_fn" will be thrown.

- Update
Adding a dummy identity layer before the checkpointed attention layer so that inputs to the later will have gradient recorded.

test/utils/test_common.py
langong347: Set `use_cache=False` is in the same codepath as `torch.utils.checkpoint.checkpoint` in a way that the former is a proxy of the latter. But I can see how this may be less obvious in regard to what's being tested. I have not seen a good way of testing triggering of `checkpoint` API elsewhere. 

I can also patch `checkpoint` and inspect the mock that it's called.

ebsmothers: Thanks for clarifying, I think that makes sense. Up to you if you want to patch checkpoint or not, but if not I would at least add a comment explaining this point.

test/utils/test_common.py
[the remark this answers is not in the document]
ebsmothers: Thanks for the explanation. At first glance the use of grads for the input layer is definitely counterintuitive, but now I see why we would need it based on how grads get handled by checkpointing. Your solution to prepend with an Identity layer sounds good to me

torchmultimodal/utils/common.py
```diff
@@ -183,3 +186,35 @@ def to_tuple_tuple(
...
+            # By default the checkpoint API stashes and restores the RNG state during each checkpointed
+            # segment such that checkpointed passes making use of RNG (e.g., through dropout, batch norm)
+            # have deterministic outputs as compared to non-checkpointed passes. This can incur a moderate
+            # performance hit which we mitigate by checkpointing either before and after the layer that
+            # requires RNG.
+            if "use_cache" in kwargs and kwargs["use_cache"] is True:
+                warnings.warn(
+                    "Using `cache` is incompatible with gradient checkpointing. Setting `use_cache=False`..."
+                )
+                kwargs["use_cache"] = False
```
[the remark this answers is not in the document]
langong347: for now it's only needed for decoding has "use_cache" for the need to save re-computing key/value (if computed in a previous step).

test/utils/test_common.py
```diff
@@ -70,3 +77,108 @@ def test_int(self, expected):
...
+                self.param = nn.Parameter(torch.ones(3))
+
+            def forward(self, x, y, attn_mask=None, use_cache=None):
+                grad = x if use_cache else y
+                grad = grad * self.param
+                if attn_mask is not None:
+                    grad = grad * attn_mask
+                return grad
+
+        class DummyIdentity(nn.Module):
```
ebsmothers, line 97: Just wondering, why do we use DummyIdentity instead of nn.Identity here? (I know you had nn.Identity previously, so assuming there is some reason for the change)

ebsmothers (approved):
Looks good to me. One final question, but accepting to unblock

torchmultimodal/utils/common.py
```diff
@@ -183,3 +185,34 @@ def to_tuple_tuple(
...
+
+            def create_custom_forward(fn: Callable) -> Callable:
+                # Specifies what should the checkpoint API run in forward pass
+                def custom_forward(*inputs: Any) -> Callable:
+                    return fn(cls, *inputs, **kwargs)
+
+                return custom_forward
+
+            return checkpoint(create_custom_forward(fn), *inputs)
```
RdoubleA, line 213: any way to simplify three levels of nested functions? can this just be `checkpoint(fn(cls, *inputs, **kwargs))`? or do you need to format a specific way for `checkpoint`?

test/utils/test_common.py
langong347: Good Q. Because `nn.Identity` does not render the output to be `requires_grad`. In order for the checkpointed attention layer to receive a "gradient-ed" input, I have to manually craft the identity layer.

torchmultimodal/utils/common.py
langong347: We need to wrap `checkpoint` around `fn(**kwargs)` and apply it to `*inputs` where `kwargs` could be constants (such as booleans) and `inputs` are what we actually re-compute against during the backward pass. 
So it'd be like:
```
checkpoint(fn(use_cache), x, mask)
```

`cls` is for us to access the training status `cls.training`. It's not consumed by `checkpoint`.

So the three levels are:
1.  branch by training mode, handle `use_cache`
2.  tell `checkpoint` what to compute from
Maybe a better interpretation from PT core [here](https://pytorch.org/docs/stable/checkpoint.html):

> function – describes what to run in the forward pass of the model or part of the model.

3.  the actual forward logic

RdoubleA: ok I see. but it just seems like `create_custom_forward` and `custom_forward` achieve the same thing as just `checkpoint(fn(cls, *inputs, **kwargs), *inputs)`. Isn't that what `create_custom_forward` is returning anyway?

langong347: `create_custom_forward` is trying to tease out the boolean arguments (or whatever provided as `**kwargs`) so that we only forward pass on the tensor arguments (see my additional comment below). 

```
def create_custom_forward(fn):
                # Specifies what should the checkpoint API run in forward pass
                def custom_forward(*inputs):    # custom_forward only interfaces *inputs
                    return fn(cls, *inputs, **kwargs)

                return custom_forward
```

But I am trying to validate if this separation is necessary and will update once I can confirm it.

langong347: So yeah I "asked a pytorch developer" and both yours and mine approaches are ok. For simplicity, I will amend this either in-place or a follow-up diff.

RdoubleA: cool, thanks for making that post about it

langong347: Unfortunately, I am not able to make the simplification as `checkpoint` does not accept kwargs other than those defined in their API. So I had to create a custom function to hide kwargs such as `use_cache` or `attn_mask` from `checkpoint`.

RdoubleA (approved):

langong347:
@langong347 has imported this pull request.  If you are a Facebook employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D38283963).

langong347:
@langong347 has imported this pull request.  If you are a Facebook employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D38283963).

PR #177, opened 2022-07-19
Add recall metrics to videoCLIP lightning module

sophiazhi:
Summary:
Add top-k recall and reverse recall metrics to VideoCLIP lightning module, with logging.

Test plan:
Compared my recall function to MUGEN's recall function to check for equal outputs ([notebook](https://colab.research.google.com/drive/190ZD0NadfXTI0OUWO-KVthP8zN1tniyq))
Calculated MUGEN's recall outputs on my mini test set of 100 samples ([notebook](https://colab.research.google.com/drive/10pXZZ9Nz5vy5SC237Brhy1pd5XZS4XNc))
Confirmed that my recall outputs matches MUGEN's after adding recall to the lightning module ([notebook](https://colab.research.google.com/drive/1uS8w0TIsc4hsGLBDdzxif97pAr9uPi6y))

### DIFF
--- a/examples/mugen/retrieval/model.py
+++ b/examples/mugen/retrieval/model.py
@@ -16,6 +16,7 @@ from torchmetrics import Recall
 from torchmultimodal.modules.losses.contrastive_loss_with_temperature import (
     ContrastiveLossWithTemperature,
 )
+from torchmultimodal.utils.common import get_current_device
 
 
 class VideoCLIPLightningModule(LightningModule):
@@ -67,7 +68,7 @@ class VideoCLIPLightningModule(LightningModule):
     def _compute_recall(self, text_embedding, video_embedding):
         similarity_matrix = text_embedding @ video_embedding.T
         num_samples = similarity_matrix.shape[0]
-        target_matrix = torch.diag(torch.ones(num_samples)).to(dtype=int)
+        target_matrix = torch.eye(n=num_samples, dtype=int, device=get_current_device())
 
         for k in self.recall_ks:
             v2t_recall = self.metrics[f"v2t_recall_{k}"]
--- a/examples/mugen/retrieval/model.py
+++ b/examples/mugen/retrieval/model.py
@@ -1,0 +1,92 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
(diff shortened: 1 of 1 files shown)


examples/mugen/retrieval/model.py
```diff
@@ -4,30 +4,37 @@
...
+from typing import Any, Tuple
+
+import torch
 
 from examples.mugen.retrieval.video_clip import videoclip
 from pytorch_lightning import LightningModule
+from torchmetrics import Recall
+
 from torchmultimodal.modules.losses.contrastive_loss_with_temperature import (
     ContrastiveLossWithTemperature,
```
langong347, line 17: it's not used anymore?

examples/mugen/retrieval/model.py
```diff
@@ -4,30 +4,37 @@
...
 )
 
 
 class VideoCLIPLightningModule(LightningModule):
     """PyTorch Lightning module for evaluating VideoCLIP model.
-
     Args:
         logit_scale (float): Initial log-temperature value for contrastive loss funtion.
-            Defaults to ``3.8282``, MUGEN's log-temperature value after training.
+            Defaults to ``0.07``, MUGEN's log-temperature value at initialization.
```
langong347, line 25: Where did the ``3.8282`` come from previously? Why "after training" --> "at initialization"?

langong347:
Did you update the last [notebook](https://colab.research.google.com/drive/1uS8w0TIsc4hsGLBDdzxif97pAr9uPi6y)? For recall I only saw a type error thrown in the end:

```
from pytorch_lightning import Trainer

trainer = Trainer()
trainer.test(pl_videoclip, dataloaders=datamodule.test_dataloader())
```

examples/mugen/retrieval/model.py
sophiazhi: 3.8282 came from the checkpoint as the trained temperature parameter, but mugen initialized temperature to 0.07 so i think it makes more sense for the lightning module to initialize to 0.07 and only use 3.8282 when using the checkpoint

examples/mugen/retrieval/model.py
sophiazhi: it's not, since i replaced measuring loss during evaluation to measuring recall (i kept the import and args because it will be used when i add training code). i think loss isn't usually reported during evaluation

RdoubleA (approved):

PR #186, opened 2022-07-20
onboard mugen dataset

sophiazhi:
Summary:
Onboard MUGEN dataset. Changes from the original MUGEN repo include:
- remove dependencies on `jukebox` and MUGEN `audio_vqvae` and place into `data/audio_utils.py`
- replace cmd line `argparse` with `MUGENDatasetArgs` and `DataModuleArgs`
- `data.py` --> `mugen_datamodules.py`. Rename `VideoData` to `MUGENDataModule` and add data transforms.
- `mugen_data.py` --> `mugen_dataset.py`
- remove `data/coinrun/assets` folder and replace with README instructions on how to download assets and a user-configurable asset path argument

Test plan:
In colab [notebook](https://colab.research.google.com/drive/1vzGm8TN-91gMlVr_skLbzW2oUWOyzJtE#scrollTo=cHk_CuGPC1sc)
Recall results match those from original MUGEN implementation.


examples/mugen/data/README.md
```diff
@@ -0,0 +1,9 @@
+This folder contains code for interfacing the MUGEN dataset (https://mugen-org.github.io/data).
```
langong347, line 1: Perhaps add a summary of what MUGEN dataset is here?

langong347 (commented):
I suppose you are still working on removing the assets and figure out instructions for people to download them, right?

examples/mugen/data/mugen_datamodules.py
```diff
@@ -0,0 +1,113 @@
...
+
+import pytorch_lightning as pl
+import torch
+import torch.distributed as dist
+import torch.utils.data as data
+
+from .mugen_dataset import MUGENDataset, MUGENDatasetArgs
+
+
+class MUGENDataModuleArgs(NamedTuple):
```
RdoubleA, line 17: why make this class, how about just using these args in `MUGENDataModule` itself? [FLAVA](https://github.com/facebookresearch/multimodal/blob/a1c136853ebee87bb148e2442991f5430cb40fa2/examples/flava/data/datamodules.py#L60) does something similar

examples/mugen/data/mugen_datamodules.py
```diff
@@ -0,0 +1,113 @@
...
+    """
+
+    def __init__(
+        self,
+        mugen_dataset_args: MUGENDatasetArgs,
+        data_module_args: MUGENDataModuleArgs = MUGENDataModuleArgs(),
+        text_transform: Optional[Callable] = None,
+        video_transform: Optional[Callable] = None,
+        audio_transform: Optional[Callable] = None,
+        shuffle=True,
```
RdoubleA, line 49: nit: we don't run type checking on examples but since you defined the type for every other argument might as well add it for `shuffle` as well

RdoubleA (approved):

examples/mugen/data/audio_utils.py
```diff
@@ -0,0 +1,67 @@
...
+# LICENSE file in the root directory of this source tree.
+
+import av
+import numpy as np
+
+
+AUDIO_SAMPLE_RATE = 22050
+AUDIO_SAMPLE_LENGTH = AUDIO_SAMPLE_RATE * 3.2  # each video is 3.2 seconds
+
+
```
langong347, line 14: Is this from Jukebox? Can we add a url to the reference code for tracking purpose?

langong347 (approved):
Looks good. Left a comment.

examples/mugen/data/audio_utils.py
sophiazhi: it's from MUGEN in their [audio vqvae](https://github.com/mugen-org/MUGEN_baseline/blob/main/lib/models/audio_vqvae/hparams.py#L7)

sophiazhi: the load_audio function is from jukebox so i'll add that url too

PR #184, opened 2022-07-20
Enable loading ckpts from manifold

ankitade:
Differential Revision: D37988804

PR #187, opened 2022-07-21
Remove embedding layer norm from clip text encoder and some fixes

ankitade:
Differential Revision: D38020819

### DIFF
--- a/test/modules/encoders/test_clip_text_encoder.py
+++ b/test/modules/encoders/test_clip_text_encoder.py
@@ -4,6 +4,8 @@
 # This source code is licensed under the BSD-style license found in the
 # LICENSE file in the root directory of this source tree.
 
+from math import inf
+
 import pytest
 import torch
 from test.test_utils import assert_expected, set_rng_seed
@@ -15,7 +17,7 @@ class TestCLIPTextEncoder:
     def start(self):
         set_rng_seed(1234)
         context_length = 77
-        batch_size, embedding_dim = 2, 4
+        batch_size, embedding_dim, heads = 2, 4, 2
 
         def build_text(text_length):
             return torch.randint(1, 10, (batch_size, text_length), dtype=torch.long)
@@ -24,34 +26,16 @@ class TestCLIPTextEncoder:
             embedding_dim=embedding_dim,
             use_clip_init=True,
             context_length=context_length,
+            heads=heads,
         ):
             return CLIPTextEncoder(
                 embedding_dim=embedding_dim,
                 use_clip_init=use_clip_init,
                 context_length=context_length,
+                heads=heads,
             )
 
         return build_encoder, build_text
-
-    def test_initialization(self, start):
-        """Test whether the ``use_clip_init`` parameter changes the initialized parameters."""
-        build_encoder, _ = start
-
-        # First, initialize without CLIP parameters and get the standard deviation of token embedding weights
-        # Use larger embedding size for stability in std
-        text_encoder = build_encoder(use_clip_init=False, embedding_dim=50)
-        assert isinstance(text_encoder, torch.nn.Module)
-        actual_std_no_init = torch.std(
-            text_encoder.encoder.token_embedding.weight
-        ).item()
-
-        # Get the standard deviation with CLIP params
-        text_encoder = build_encoder(embedding_dim=50)
-        actual_std_clip_init = torch.std(
-            text_encoder.encoder.token_embedding.weight
-        ).item()
-
-        assert abs(actual_std_clip_init - actual_std_no_init) > 1e-1
 
     def test_clip_parameters(self, start):
         build_encoder, _ = start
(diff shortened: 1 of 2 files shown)


torchmultimodal/modules/encoders/clip_text_encoder.py
```diff
@@ -39,42 +47,49 @@ def __init__(
...
+        self.encoder = TransformerEncoder(encoder_layer, num_layers=layers)
 
         self.width = width
         self.context_length = context_length
 
         self.ln_final = nn.LayerNorm(width)
         self.projection = nn.Linear(width, embedding_dim, bias=False)
 
-        if use_clip_init:
+        if pretrained_url:
```
langong347, line 72: Why not having `load_module_from_url` outside of the module so that:
-  you don't need to pass in `self` as the first argument? 
-  `pretrained` is more of an option to construct a model rather than part of the model logic

For example, in [Omnivore](https://github.com/facebookresearch/multimodal/blob/e274207e5c091d681bcb790d561fca0764148549/torchmultimodal/models/omnivore.py#L148) and [VideoCLIP](https://github.com/facebookresearch/multimodal/blob/main/examples/mugen/retrieval/video_clip.py#L163) we had `load_model` at the builder level.

torchmultimodal/modules/encoders/clip_text_encoder.py
```diff
@@ -39,42 +47,49 @@ def __init__(
...
+        )
+        encoder_layer = TransformerEncoderLayer(
+            d_model=width,
+            nhead=heads,
             dropout=0.0,
-            normalize_before=True,
+            activation=quick_gelu,
+            norm_first=True,
         )
+        self.encoder = TransformerEncoder(encoder_layer, num_layers=layers)
```
langong347, line 64: this is to utilize PT core's better transformer? I thought text team has or is going to adopt it?

torchmultimodal/modules/encoders/clip_text_encoder.py
```diff
@@ -87,14 +102,21 @@ def initialize_parameters(self) -> None:
 
     def build_attention_mask(self) -> torch.Tensor:
         # To support torchscripting, we have to pass an int as fill_value
-        mask = torch.full((self.context_length, self.context_length), int(True)).triu(1)
-        return mask.to(device=None, dtype=torch.bool)
+        mask = torch.full(
+            (self.context_length, self.context_length), float("-inf")
```
langong347, line 106: Why "-inf" here? Originally the mask was filled with "True" (or integer 1 for torchscript):

```
def build_attention_mask(self):
        mask = torch.full((self.context_length, self.context_length), True).triu(1)
        return mask.to(device=None, dtype=torch.bool)
```

cc @ebsmothers @sophiazhi

test/modules/encoders/test_clip_text_encoder.py
```diff
@@ -24,89 +26,69 @@ def build_encoder(
...
             return CLIPTextEncoder(
                 embedding_dim=embedding_dim,
                 use_clip_init=use_clip_init,
                 context_length=context_length,
+                heads=heads,
             )
 
         return build_encoder, build_text
 
-    def test_initialization(self, start):
```
langong347, line 36: Do you mean to remove this test or plan to modify it? This was originally proposed to validate the effect of enabling clip initialization.

cc @sophiazhi  @ebsmothers

ebsmothers: I chatted with @ankitade about this yesterday and seems this test is a bit flaky, making it difficult to make changes to other parts of CLIP code without triggering failures. Given that the specific initialization is not used except for the pretraining from the paper (which we are not doing), I think it's OK to remove.

torchmultimodal/utils/common.py
```diff
@@ -96,18 +97,14 @@ def tensor_slice(x: Tensor, begin: List[int], size: List[int]) -> Tensor:
 
 
 def load_module_from_url(
-    model: torch.nn.Module, url: str, progress: bool = True
+    model: torch.nn.Module, url: str, strict: bool = True, progress: bool = True
```
ebsmothers, line 100: Are you abandoning #184?

torchmultimodal/modules/encoders/clip_text_encoder.py
ebsmothers: They have, but they are also wrapping it with some other stuff (for instance their class also includes the embeddings).

torchmultimodal/modules/encoders/clip_text_encoder.py
ebsmothers: Previously TorchText's transformer encoder was calling into their own MHA [here](https://github.com/pytorch/text/blob/ed44dc88539b51785df03a276410cf71adfe6bfd/torchtext/models/roberta/modules.py#L135-L137). But with the change to transformer encoder they are no longer doing this. So I think we need to make this change regardless

torchmultimodal/modules/encoders/clip_text_encoder.py
RdoubleA: This should ultimately be updated by our unified TransformerEncoder

torchmultimodal/modules/encoders/clip_text_encoder.py
ankitade: this was more for user passing in any ckpt path (vs predefined ones). but i pushed the responsibility for such things to the user for now

torchmultimodal/modules/encoders/clip_text_encoder.py
ankitade: the goal here is to get rid of the layer norm after embedding, using core's transformer is just a side effect

ebsmothers (approved):

PR #194, opened 2022-07-23
Support training for VideoCLIP lightning module

sophiazhi:
Summary:
Add training step, validation step, and optimizer to VideoCLIP lightning module so the model can be trained.

Test plan:
In [notebook](https://colab.research.google.com/drive/1VE68J34YEfuVUN8KZ3ZYOrxfCCMsJeHf?usp=sharing)


ebsmothers (approved):
Looks good to me. Do you plan to add any other validation metrics besides loss?

sophiazhi:
> Looks good to me. Do you plan to add any other validation metrics besides loss?

Just added recall logging to both train and validation steps (same as original mugen repo) since i overlooked that originally

PR #219, opened 2022-07-29
[utils] Add test util for init model params to ones

RdoubleA:
* #223
* #222
* #221
* __->__ #219
* #218

### DIFF
--- a/test/test_utils.py
+++ b/test/test_utils.py
@@ -13,6 +13,7 @@ from typing import Any
 
 import torch
 import torch.distributed as dist
+from torch import nn
 
 
 def gpu_test(gpu_count: int = 1):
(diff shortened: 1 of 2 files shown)


test/test_utils.py
```diff
@@ -79,3 +80,8 @@ def assert_expected(actual: Any, expected: Any, rtol: float = None, atol: float
         atol=atol,
         msg=f"actual: {actual}, expected: {expected}",
     )
+
+
+def init_params_ones(model: nn.Module) -> None:
```
langong347, line 85: Would it be too specific to init everything to 1? Or do you plan to make this configurable?

RdoubleA: No plans at the moment, if we need to init to anything else we can make the change later or add a new function. Not sure what else we could init to for testing.

langong347: Non-blocking Q:

The need to have deterministic model params is to avoid output shift due to model structure changes, right?

In general, I am not too sure about this approach (to instill artificial values). Statistically, this may skew certain properties. I suppose once our models stabilize, there will be less of the need to do so.

Just some food for thought: [TorchVision](https://github.com/pytorch/vision/blob/main/test/test_models.py#L602)'s model test design:
- set rng seed
- init the model
- pass in random inputs
- compare with expected outputs (stored as pickles, i.e., `test/expect/*.pkl`)

PR #220, opened 2022-07-29
[Unification] Return attn probs by default in MHA

RdoubleA:
* #223
* #222
* #221
* __->__ #220
* #219
* #218

### DIFF
--- a/test/modules/layers/test_attention.py
+++ b/test/modules/layers/test_attention.py
@@ -80,7 +80,7 @@ class TestMultiheadAttention:
     ):
         mha = multihead_attn(1, self_attn)
         qkv = 2 * torch.ones(1, *input_shape, hidden_dim)
-        actual = mha(qkv)
+        actual, _ = mha(qkv)
         expected = torch.tensor(
             [
                 [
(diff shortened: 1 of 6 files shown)


langong347 (commented):
Is this a must-have that we return attn_weights by default? Perhaps for unification we can focus on the big refactoring as small changes like this can cause a lot of churns to call-sites that are not merged yet.

- huggingface: https://github.com/huggingface/transformers/blob/v4.20.1/src/transformers/models/gpt2/modeling_gpt2.py#L339
- PT core: https://pytorch.org/docs/stable/generated/torch.nn.MultiheadAttention.html#torch.nn.MultiheadAttention.forward

RdoubleA:
I didn't like having to repeat this check for the flag up the stack. But if it's gonna cause too much thrash I can remove it, since it seems standard to use the flag as you pointed out.

langong347:
> I didn't like having to repeat this check for the flag up the stack. But if it's gonna cause too much thrash I can remove it, since it seems standard to use the flag as you pointed out.

Leave aside the thrashes for a moment. So probably I should have asked this way:

The call-sites will need a flag to return the attn_probs returned by MHA unless we want to return them by default.
In other words, removing the flag at MHA level does not result in its removal at the call-sites.
For example, my transformer encoder may decide not to return the attn_probs by default because it will take up `num_layer * seq_len^2)` space. In this case, I will still need a flag to either return or discard what MHA returns.

Not sure how we may avoid repeat this check?

PR #224, opened 2022-07-29
albef vqa finetune

katrina433:
* #229
* #228
* #227
* #226
* #225
* __->__ #224

Differential Revision: [D38881553](https://our.internmc.facebook.com/intern/diff/D38881553)


katrina433:
Refer to pr #174 for previous reviews, but this one will be landed as part of the stack.

torchmultimodal/utils/attention.py
```diff
@@ -43,6 +45,32 @@ def get_extended_attention_mask(attention_mask: Tensor) -> Tensor:
     return extended_attention_mask
 
 
+def get_extended_attention_mask_for_decoder(attention_mask: Tensor) -> Tensor:
```
ebsmothers, line 48: I'm still not 100% sure why we need this. Why not just call get_causal_attention_mask and get_extended_attention_mask(attn_mask) then take the product in the decoder itself?

examples/albef/model.py
```diff
@@ -0,0 +1,361 @@
...
+        super().__init__()
+        self.model = model
+        self.answer_decoder = answer_decoder
+        self.loss = loss
+        self.answer_decoder_m = copy.deepcopy(self.answer_decoder)
+        remove_grad(
+            self.answer_decoder_m
+        )  # remove gradient for the momentum decoder model
+
+    def _train_forward(
```
ebsmothers, line 166: Both these forwards are hard to follow (I think this is unavoidable). But maybe a brief docstring for each method explaining at a high level the steps it is taking would help

ebsmothers (approved):
Looks good! Just a few more comments, but accepting to unblock you

examples/albef/configs/vqa.yaml
```diff
@@ -0,0 +1,71 @@
+hidden_size: &hidden_size 768
```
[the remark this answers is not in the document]
ebsmothers: Yeah the `_target_` approach is specific to Hydra. It could be a nice extension but for now I personally would not consider it mandatory. I think `&` is just used to mark an alias that can be reused later (i.e. if we want to use the same value for multiple configurables to constrain their values to match)

examples/albef/finetune.py
```diff
@@ -0,0 +1,197 @@
...
+
+    dist_utils.init_distributed_mode(args)
+    device = torch.device(args.device)
+
+    seed = config["seed"] + dist_utils.get_rank()
+    torch.manual_seed(seed)
+    random.seed(seed)
+    cudnn.benchmark = True
+
+    datamodule = VQADataModule(**config["datamodule_args"])
```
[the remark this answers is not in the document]
ebsmothers: The alternative would just be to use data loader directly. I think we are trending in that direction (already had some conversations with Ankita about doing this for MDETR, which is in a similar state)

ebsmothers:
@ebsmothers has imported this pull request.  If you are a Facebook employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D38881553).

torchmultimodal/utils/attention.py
```diff
@@ -27,14 +27,16 @@ def get_extended_attention_mask(attention_mask: Tensor) -> Tensor:
         extended_attention_mask (Tensor):
             The broadcastable attention mask, with the same dtype as ``attention_mask.dtype``.
     """
-    # We can provide a self-attention mask of dimensions [batch_size, query_seq_length, key_seq_length]
-    # ourselves in which case we just need to make it broadcastable to all heads,
-    # [batch_size, num_heads, query_seq_length, key_seq_length].
-    if attention_mask.dim() == 3:
+    if attention_mask.dim() == 4:
```
langong347, line 30: Can we add comments to this codepath as well? 
Also update the docstring since we are now accepting masks with head dim.

```
Accepting two types of attention masks:
        - Causal: masks that prevent attending to future positions of dimensions
            ``(batch_size, query_seq_len, key_seq_len)``
        - Padding: masks that prevent attending to token paddings of dimensions
            ``(batch_size, seq_len)``
```

langong347: Non blocking Q: this utility is ambiguous as it puts strong assumption around the original dims of the masks.
For example, dim=3 could mean both `(b, h, seq_len)` and `(b, q_seq_len, k_seq_len)`.
May introduce bugs unless we have a convention for the original dims. Better be careful with it.

ebsmothers: Yeah this is a fair point. I think we should not expect users to pass `(b, h, seq_len)` very often, since the inclusion of `h` implies they have already broadcasted (and if using this util the broadcasting will have already been to 4D). But agree there can be some ambiguity here. We could add a flag, but not sure that is the best approach either...

ebsmothers:
@ebsmothers has imported this pull request.  If you are a Facebook employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D38881553).

torchmultimodal/utils/attention.py
langong347: nit: If the more common case is passing without either batch or head dims, perhaps update the docstring to reflect that expectation? Like:

```
Accepting two types of attention masks:
        - Causal: masks that prevent attending to future positions of dimensions
            ``(query_seq_len, key_seq_len)``
        - Padding: masks that prevent attending to token paddings of dimensions
            ``(seq_len)``
```

ebsmothers: Oh sorry I should clarify... I mean that the common cases are `(b, query_seq_len, key_seq_len)` or `(b, seq_len)`. So in either case we still have `b`, it is just `h` that is missing

ebsmothers:
@ebsmothers has imported this pull request.  If you are a Facebook employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D38881553).

ebsmothers:
@ebsmothers has imported this pull request.  If you are a Facebook employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D38881553).

PR #226, opened 2022-07-29
move alpha to forward in itc loss

katrina433:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* #229
* #228
* #227
* __->__ #226
* #225
* #224

Differential Revision: [D38881554](https://our.internmc.facebook.com/intern/diff/D38881554)

PR #225, opened 2022-07-29
albef retrieval dataset

katrina433:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* #229
* #228
* #227
* #226
* __->__ #225
* #224

Differential Revision: [D38881555](https://our.internmc.facebook.com/intern/diff/D38881555)

PR #227, opened 2022-07-29
albef retrieval model

katrina433:
* #229
* #228
* __->__ #227
* #226
* #225
* #224

Differential Revision: [D38881557](https://our.internmc.facebook.com/intern/diff/D38881557)


torchmultimodal/models/albef.py
```diff
@@ -211,10 +211,10 @@ def forward(
         image_embeds_neg, text_embeds_neg, text_atts_neg = self._neg_embeddings(
             outputs.image_embeddings, outputs.text_embeddings, text_atts, similarity
         )
-        multimodal_embeddings_neg = self.model.multimodal_encoder(
-            torch.cat([image_embeds_neg, outputs.image_embeddings], dim=0),
+        multimodal_embeddings_neg = self.albef_model.multimodal_encoder(
             torch.cat([outputs.text_embeddings, text_embeds_neg], dim=0),
             torch.cat([text_atts, text_atts_neg], dim=0),
+            torch.cat([image_embeds_neg, outputs.image_embeddings], dim=0),
```
ebsmothers, line 217: This is just a continuation of the changes you made to multimodal encoder in #224, right?

examples/albef/model.py
```diff
@@ -344,6 +348,155 @@ def forward(
...
+                image is not None and text is not None and text_atts is not None
+            ), "image embeddings, text embeddings, and text attention mask cannot be None"
+            multimodal_embeds = (
+                self.model_with_similarity.albef_model.multimodal_encoder(
+                    text,
+                    text_atts,
+                    image,
+                )
+            )
+            score = self.itm_loss.itm_head(multimodal_embeds[:, 0, :])[:, 1]
```
ebsmothers, line 468: I'm confused by this, why does ITM loss have an ITM head?

examples/albef/model.py
```diff
@@ -344,6 +348,155 @@ def forward(
...
+        )
+
+        pos_embeddings = encoder_output.multimodal_embeddings[:, 0, :]
+        neg_embeddings = encoder_output.multimodal_embeddings_neg[:, 0, :]
+        itm_loss = self.itm_loss(pos_embeddings, neg_embeddings)
+
+        loss = itc_loss + itm_loss
+        return loss
+
+    def _eval_forward(
```
ebsmothers, line 432: Is there some way we can clean this method up? It feels like we are trying to do too much here

examples/albef/model.py
```diff
@@ -344,6 +348,155 @@ def forward(
...
+            similarity_targets,
+            alpha,
+        )
+
+        pos_embeddings = encoder_output.multimodal_embeddings[:, 0, :]
+        neg_embeddings = encoder_output.multimodal_embeddings_neg[:, 0, :]
+        itm_loss = self.itm_loss(pos_embeddings, neg_embeddings)
+
+        loss = itc_loss + itm_loss
+        return loss
```
ebsmothers, line 430: Do we need a separate class to do this? Seems like we could just return the encoder_outputs and calculate the losses directly from those without bundling them with ALBEFModelWithSImilarity (at least for train, not sure how it will work for eval)

examples/albef/model.py
```diff
@@ -448,5 +627,35 @@ def albef_model_for_vqa(config: dict, pretrained: bool = False) -> ALBEFModelFor
...
     if pretrained:
-        load_module_from_url(model, _ALBEF_PRETRAINED_URLS["vqa"])
+        checkpoint = torch.hub.load_state_dict_from_url(
+            _ALBEF_PRETRAINED_URLS["vqa"], map_location="cpu"
+        )
+        model.load_state_dict(checkpoint)
+    return model
+
+
+def albef_model_for_retrieval(
```
ebsmothers, line 637: I think we have been leaning more towards using explicit args in the builder functions instead of configs. Not a huge deal, but something worth thinking about for consistency (can always add a to-do to unify these)

examples/albef/model.py
```diff
@@ -431,6 +430,186 @@ def forward(
             )
 
 
+class ALBEFModelForRetrieval(nn.Module):
+    """
+    ALBEF Model for Retrieval finetuning and inference.
+
```
ebsmothers, line 436: Can you add more detail in the docstring explaining the different forward modes here (train vs eval, image vs text vs multimodal)

ebsmothers (approved):

examples/albef/model.py
RdoubleA: I'm on the fence with this, the closest comparison is [flava_model](https://github.com/facebookresearch/multimodal/blob/e73ad6db1d2f93b55a4290649470e7ce005dbb2b/torchmultimodal/models/flava/model.py#L418) which flattens each encoder's args, leading to a gigantic function signature. I would lean towards being consistent with the codebase, but I see the appeal of having just one dict with all the args.

another thing I would prioritize is the ability to just call this function and have it create everything with default parameters. at this high level you want to make it as easy as possible for the user. flattening the args let's you do that. another option is to break it down into kwargs for each encoder. so the function will have `vision_encoder_args`, `text_encoder_args`, etc as params that are optional. that way you can still take advantage of using default values if the user does not want to specify anything.

ebsmothers:
@ebsmothers has imported this pull request.  If you are a Facebook employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D38881557).

ebsmothers:
@ebsmothers has imported this pull request.  If you are a Facebook employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D38881557).

torchmultimodal/models/albef/model.py
```diff
@@ -167,12 +167,12 @@ def __init__(
         remove_grad(self.text_proj_m)
 
         self.queue_size = queue_size
-        self.temp = temp
+        self.temp = nn.Parameter(torch.ones([]) * temp)
```
langong347, line 170: Is the purpose to create a scalar tensor? 
Why not just do `torch.tensor(temp)`?
```
>>> nn.Parameter(torch.tensor(0.7))
Parameter containing:
tensor(0.7000, requires_grad=True)
>>> nn.Parameter(torch.ones([]) * 0.7)
Parameter containing:
tensor(0.7000, requires_grad=True)
```

ebsmothers: Yeah I often see scalar parameters constructed this way (see e.g. [CLIP](https://github.com/mlfoundations/open_clip/blob/1eb2fcfbc1e1c80b94fd63130f4811f09af20c8b/src/open_clip/model.py#L395)). I think it is more for legacy reasons than anything, see the discussion [here](https://discuss.pytorch.org/t/best-way-to-define-a-scalar-using-nn-parameter-in-pytorch/97209). Will change it for clarity

ebsmothers:
@ebsmothers has imported this pull request.  If you are a Facebook employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D38881557).

ebsmothers:
@ebsmothers has imported this pull request.  If you are a Facebook employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D38881557).

ebsmothers:
@ebsmothers has imported this pull request.  If you are a Facebook employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D38881557).

PR #230, opened 2022-07-30
[MDETR] Move model components into models/mdetr

ebsmothers:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* #246
* #232
* #231
* __->__ #230

Summary: Mostly just moving around existing components. Create a new directory models/mdetr
and put text and image encoders in there. Also move transformer components into their own file.
This is prep for the next PR, which will move task-specific models into mdetr/model.py, so taking
the transformer stuff out will make the file clearer.

Test plan:

```
python -m pytest -vv

```

Differential Revision: [D38467125](https://our.internmc.facebook.com/intern/diff/D38467125)

PR #231, opened 2022-07-30
[MDETR] Separate losses from model

ebsmothers:
* #246
* #232
* __->__ #231
* #230

Summary:
This PR takes the losses and matcher out of MDETRForVQA and MDETRForPhraseGrounding classes.
We keep contrastive projections in both, even though they are not technically necessary for eval.
The alternative is to add a flag but that introduces a lot of copypasta. Also modified the eval
and fine-tuning loops and added a builder for the loss class.

Test plan:
Phrase grounding eval:

```
CUBLAS_WORKSPACE_CONFIG=:4096:8 python -m torch.distributed.launch --nproc_per_node=2 --use_env phrase_grounding.py
--resume https://pytorch.s3.amazonaws.com/models/multimodal/mdetr/pretrained_resnet101_checkpoint.pth --ema --eval
--dataset_config phrase_grounding.json
```
Gives recall@1 82.4% as expected

VQA eval:

```
CUBLAS_WORKSPACE_CONFIG=:4096:8 python -m torch.distributed.launch --nproc_per_node=4 --use_env vqa_eval.py
--resume https://pytorch.s3.amazonaws.com/models/multimodal/mdetr/gqa_resnet101_checkpoint.pth --ema --eval
--dataset_config vqa.json
```
Gives total accuracy 61.6% as expected

VQA fine-tuning:
```
CUBLAS_WORKSPACE_CONFIG=:4096:8 python -m torch.distributed.launch --nproc_per_node=4 --use_env vqa_finetune.py
--dataset_config vqa.json --ema --epochs 125 --epoch_chunks 3 --lr_drop 150
--load /data/home/ebs/data/mdetr/checkpoints/pretrained_resnet101_checkpoint.pth --qa_loss_coef 25 --batch_size 4
--lr 1.4e-4 --lr_backbone 1.4e-5 --text_encoder_lr 7e-5
```
Logged losses from first ten iterations:
```
0 tensor(199.3950, device='cuda:0', grad_fn=<AddBackward0>)
1 tensor(178.6834, device='cuda:0', grad_fn=<AddBackward0>)
2 tensor(187.6183, device='cuda:0', grad_fn=<AddBackward0>)
3 tensor(164.5415, device='cuda:0', grad_fn=<AddBackward0>)
4 tensor(352.6355, device='cuda:0', grad_fn=<AddBackward0>)
5 tensor(220.8754, device='cuda:0', grad_fn=<AddBackward0>)
6 tensor(178.6945, device='cuda:0', grad_fn=<AddBackward0>)
7 tensor(342.7989, device='cuda:0', grad_fn=<AddBackward0>)
8 tensor(169.4948, device='cuda:0', grad_fn=<AddBackward0>)
9 tensor(186.1401, device='cuda:0', grad_fn=<AddBackward0>)
10 tensor(126.2118, device='cuda:0', grad_fn=<AddBackward0>),
```
which matches the results on main branch.

Differential Revision: [D38467124](https://our.internmc.facebook.com/intern/diff/D38467124)


torchmultimodal/models/mdetr/model.py
```diff
@@ -257,3 +260,236 @@ def mdetr_resnet101(
         extra_query_embeddings,
     )
     return mdetr
+
+
+class MultiHead(nn.Module):
```
RdoubleA, line 265: what was the reason for moving these out of examples?

torchmultimodal/models/mdetr/model.py
```diff
@@ -257,3 +260,236 @@ def mdetr_resnet101(
...
+            contrastive_query_embeddings, contrastive_token_embeddings
+        )
+
+        # Apply VQA heads to get answer predictions
+        answer_preds = self.vqa_heads(model_output.extra_embeddings.transpose(0, 1))
+
+        return MDETRVQAOutput(model_output, answer_preds, contrastive_outputs)
+
+
+def mdetr_for_vqa(
```
langong347, line 372: Consider adding an option to load from checkpoint to the builder (e.g., omnivore.py)?
From a user's perspective, I'd probably prefer a clean API such as `build_model(pretrained=True)`.

RE: this block in `eval.py`.
https://github.com/facebookresearch/multimodal/blob/9b3d1a17de0d9b20f9e83db423b24e26a4363e57/examples/mdetr/vqa_eval.py#L156-L163

ebsmothers: Yeah this is still a to-do, I am planning to address separately if that's good with you

torchmultimodal/models/mdetr/model.py
ebsmothers: Trying to follow the convention with FLAVA of having task-specific model classes in core (as long as it makes sense). For MultiHead specifically, it is just a ModuleDict wrapper and gets removed in the next PR

torchmultimodal/models/mdetr/model.py
```diff
@@ -257,3 +260,236 @@ def mdetr_resnet101(
...
+    token_embeddings: Tensor
+
+
+class MDETRVQAOutput(NamedTuple):
+    model_output: MDETRModelOutput
+    vqa_preds: Dict[str, Tensor]
+    contrastive_embeddings: ContrastiveEmbeddingsOutput
+
+
+class MDETRForVQA(nn.Module):
```
langong347, line 314: Are those task-specific modules more suitable to be placed inside `examples/`?

examples/mdetr/utils/args_parse.py
```diff
@@ -84,23 +84,29 @@ def get_args_parser():
         "--contrastive_loss_hdim",
         type=int,
         default=64,
-        help="Projection head output size before computing normalized temperature-scaled cross entropy loss",
+        help="Projection head output size before computing the contrastive alignment loss",
     )
-    # * Matcher
     parser.add_argument(
```
langong347, line 89: As a follow-up: use config system for this?

torchmultimodal/models/mdetr/model.py
ebsmothers: See my response to Rafi's comment 🙂 . Basically I am following the FLAVA convention here. If we do decide to keep them in core, I will add unit tests for each of these

torchmultimodal/models/mdetr/model.py
ebsmothers: @langong347 Actually after thinking about this a bit more, I am not sure if there will be a clean approach to the checkpoint loading API for MDETR. There are a couple reasons for this. 

1) The full MDETR model checkpoint contains both model weights and EMA model weights. We could modify the builder to return both model and EMA model, but I don't think this is the right approach. (For instance for inference we only use the EMA model, so we would not be able to use it then.) So otherwise we will have to pass a key to e.g. the `load_module_from_url` method. (I think `PretrainedMixin` already supports this, but iirc we are thinking of deprecating it?) 

2) MDETR contains a handful of configurable components, meaning that there is not really a canonical checkpoint for a given task. For instance, if we are loading a checkpoint to do VQA eval we expect VQA heads in the state dict, but if we are loading a checkpoint to do VQA fine-tuning we should load the pretrained checkpoint directly, constructing the VQA heads ourselves. There is also the question of contrastive projections, which can be used or omitted depending on whether we are performing training or not. Of course we can set `strict=False` to allow for each of these cases, but then there is a higher likelihood that a user playing around with various configurations would break something and not realize it. 

Hopefully this makes sense, please let me know if any part of it is unclear.

langong347: >MDETR contains a handful of configurable components, meaning that there is not really a canonical checkpoint for a given task. 

It sounds like your current checkpoint files are the superset of what you want for different purposes? 
Technically, is it possible to checkpoint different parts of the model separately?

torchmultimodal/models/mdetr/model.py
ankitade: makes sense to put in core since people finetuning (within their code) on these will have to import from examples?

torchmultimodal/models/mdetr/model.py
```diff
@@ -257,3 +260,236 @@ def mdetr_resnet101(
...
+class MultiHead(nn.Module):
+    def __init__(self, heads: nn.ModuleDict):
+        super().__init__()
+        self.heads = heads
+
+    def forward(
+        self,
+        embeddings: Tensor,
+    ) -> Dict[str, Tensor]:
+        if embeddings.size(0) != len(self.heads):
```
ankitade, line 274: just curious, why do u need a nn module for this

torchmultimodal/models/mdetr/model.py
ankitade: yeah i think we need to add the official ckpt loading option here (which u can hack around as we did for omnivore and flava). adding the key option to the function is reasonable too

examples/mdetr/vqa_eval.py
```diff
@@ -136,18 +160,20 @@ def main(args):
     else:
         checkpoint = torch.load(args.resume, map_location="cpu")
 
-    model_without_ddp.load_state_dict(checkpoint["model"])
+    model_without_ddp.load_state_dict(checkpoint["model"], strict=False)
```
ankitade, line 163: why not strict? how was it working till now?

ankitade (approved):
i like the test plan

torchmultimodal/models/mdetr/model.py
ebsmothers: It gets removed in the next PR 🙂

examples/mdetr/model.py
```diff
@@ -1,370 +0,0 @@
...
-        transformer_d_model,
-        transformer_num_heads,
-        transformer_encoder_layers,
-        transformer_decoder_layers,
-        transformer_dim_feedforward,
-        transformer_dropout,
-        return_intermediate_dec,
-        num_extra_query_embeddings=num_heads,
-    )
-    if contrastive_dim is not None:
```
ebsmothers, line 213: @ankitade in response to your question about strict checkpoint loading: the previous version of MDETRForVQA had optional contrastive projections, while the new one makes them mandatory. But these projections are not present in the GQA fine-tuned checkpoint. So we need to disable strict checkpoint loading to do VQA eval. The alternative is to make them optional again, but it makes the code kinda messy

examples/mdetr/vqa_eval.py
ebsmothers: tagged you in the old model.py file with the explanation

torchmultimodal/models/mdetr/model.py
ebsmothers: Regarding checkpointing different parts of the model separately, it's technically possible. But I'm not sure it will be intuitive for the end user. Let me give a bit of background first.

The existing model checkpoints look like `{"model": model_weights, "model_ema": model_ema_weights}`, where both model_weights and model_ema_weights have the same set of keys. Aside from the core MDETR model weights, these include
- VQA head weights for the fine-tuned VQA checkpoint, and
- Contrastive projections for the pretrained checkpoint.

(Note that `MDETRForVQA` can be used for either evaluation or fine-tuning, which means that our `MDETRForVQAClass` may load either one of these checkpoints.)

So how should we split the checkpoint? I think there are two ways to do this.

1) Create a separate checkpoint file for each of `model` and `model_ema`. 
2) Create a separate checkpoint file for each of e.g. `core_model_weights`, `contrastive_projections`, and `vqa_heads`, then mix and match as needed.

I don't like (1) because it can be solved already by adding the key option to `load_module_from_url`. Then there is no extra confusion to the user and we still have all the weights in one place.

For (2), I think it is very counterintuitive. It is not really a standard kind of encoder/head split, and as a user I would be very confused about why a couple projections necessitate their own separate checkpoint. There is also the question of how this would mix with (1) (since each of these is nested under the model/model_ema keys).

Assuming that we agree we don't want to split the checkpoint (maybe a big assumption 🙂), the remaining question is how to handle the case where the weights don't line up perfectly. This can happen if e.g. the user loads `MDETRForVQA` with pretrained weights to do fine-tuning on the GQA dataset (since the model class will have VQA heads, which won't be present in the checkpoint). In this case, I think we can just load with `strict=False`. We have control over the checkpoints and builder functions so it is unlikely that users will be able modify things in such a way that `strict=False` will mask some other checkpoint loading error. I think [FLAVA](https://github.com/facebookresearch/multimodal/blob/679f3596e4c44b483c68d4023b24e3c7f77292b3/torchmultimodal/models/flava/flava_model.py#L565) is also doing this for a similar case. This way we don't have to add a bunch of slightly different checkpoints or a bunch of branching logic to the model classes or builders.

Let me know if this makes sense and if you have any concerns with the proposed approach @langong347 @ankitade.

ebsmothers:
@ebsmothers has imported this pull request.  If you are a Facebook employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D38467124).

ebsmothers:
@ebsmothers has imported this pull request.  If you are a Facebook employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D38467124).

ebsmothers:
@ebsmothers has imported this pull request.  If you are a Facebook employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D38467124).

ebsmothers:
@ebsmothers has imported this pull request.  If you are a Facebook employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D38467124).

torchmultimodal/models/mdetr/model.py
langong347: Not breaking the checkpoint file into core_model, projection and vqa_heads to avoid confusing the user. This makes sense to me since you have more pieces than just `encoder` and `head`.

>I don't like (1) because it can be solved already by adding the key option to load_module_from_url. Then there is no extra confusion to the user and we still have all the weights in one place.

Do you mean `mdtr_model.ckpt` and `mdtr_model_ema.ckpt` are confusing? That might be a personal choice I think. If we keep them in one file, you'll need an additional option in `load_module_from_url` to branch out? This could be too specific.

torchmultimodal/models/mdetr/model.py
langong347: @ebsmothers Are you going to add unit tests for those finetuning modules in a separate PR?

torchmultimodal/models/mdetr/model.py
ebsmothers: Sorry for my delayed response as well 🙂 . Yeah I agree that there is some aspect of personal preference here. I think the branching should not be too difficult, it would be the same as what we already had in [PretrainedMixin](https://github.com/facebookresearch/multimodal/blob/8e99bf08a545acd99498bb69e768510db7069077/torchmultimodal/utils/common.py#L149). 

I also think the answer here depends on how the potentially separated weights would be used. For instance, in Omnivore the weights for encoder and heads are separated, because the user may want to plug in each one of these separately with other components, and so modularity becomes important. But for EMA it is not about plugging in as much as it is about train vs eval (in train we will want both model and EMA weights, in eval we only want EMA weights). So imo the value of keeping separate weights for model and EMA is not very high and it is better to just filter the state dict if calling from an eval loop.

torchmultimodal/models/mdetr/model.py
ebsmothers: I created a bootcamp task for this

PR #233, opened 2022-08-01
yaml config for VideoCLIP eval script

sophiazhi:
Summary:
Add VideoCLIP eval script and pass arguments through yaml config with `omegaconf`.
- `torchmultimodal/transforms/video_transform.py`: switch mean and std types to tuple to satisfy hydra requirements (immutability).
- `test/transforms/video_transform.py`: change mean and std inputs to tuple
- `examples/mugen/data/mugen_dataset.py`: change `get_[video/audio/text/seg_map]` defaults in `MUGENDatasetArgs` to `False` for better clarity when instantiating arguments. E.g. `MUGENDatasetArgs(get_game_frame=True, get_text_desc=True)` instead of `MUGENDatasetArgs(get_audio=False, get_seg_map=False)`.
- `examples/mugen/retrieval/definitions.py`: dataclasses for hydra type checking with yaml config
- `examples/mugen/retrieval/configs/eval.yaml`: yaml config for eval script
- `examples/mugen/retrieval/eval.py`: eval script
- `examples/mugen/requirements.txt`: add `omegaconf` and `hydra-core` requirements
- `examples/mugen/retrieval/model.py`: change target matrix device from `get_current_device()` to `similarity_matrix.device`, to support the case when gpu is available but accelerator has been set to cpu

Test plan:
`python eval.py config=configs/eval.yaml`


examples/mugen/retrieval/eval.py
```diff
@@ -4,114 +4,50 @@
...
-        }
-    )
+def get_yaml_config():
+    cli_conf = OmegaConf.from_cli()
+    if "config" not in cli_conf:
+        raise ValueError(
+            "Please pass 'config' to specify configuration yaml file for running VideoCLIP evaluation"
+        )
+    yaml_conf = OmegaConf.load(cli_conf.config)
+    return OmegaConf.to_container(yaml_conf)
```
langong347, line 25: Why `to_container`? Looks like you can access `conf` object like a dictionary?

examples/mugen/retrieval/eval.py
```diff
@@ -4,114 +4,50 @@
...
+    videoclip_args = args["videoclip_args"]
+    evaluation_args = args["evaluation_args"]
+
+    dataset_args = MUGENDatasetArgs(get_audio=False, **dataset_args)
     datamodule = MUGENDataModule(
         dataset_args,
-        text_transform=BertTextTransform(),
-        video_transform=VideoTransform(),
-        **vars(datamodule_args),
+        text_transform=BertTextTransform(**datamodule_args["bert_text_transform"]),
```
langong347, line 39: Why do we need items with `{}` value in the yaml file? Looks like they are bloating your constructor APIs here.

examples/mugen/retrieval/eval.py
```diff
@@ -4,114 +4,50 @@
...
+        batch_size=datamodule_args["batch_size"],
+        num_workers=datamodule_args["num_workers"],
+        shuffle=datamodule_args["shuffle"],
     )
 
-    model = VideoCLIPLightningModule(
-        **vars(lightningmodule_args), **vars(videoclip_args)
-    )
-    model = model.load_from_checkpoint(evaluation_args.checkpoint_path)
+    model = VideoCLIPLightningModule(**lightningmodule_args, **videoclip_args)
```
langong347, line 46: This can be error prone if the order of your expanded kwargs is different from that of your constructor signature.
Would probably map each argument explicitly here. See for example: https://github.com/facebookresearch/multimodal/blob/855c9ed7e719fab5a39b5e564073129e0e79af7d/examples/flava/finetune.py#L47-L56

examples/mugen/retrieval/configs/eval.yaml
```diff
@@ -0,0 +1,39 @@
+dataset_args:
```
langong347, line 1: Explore using structure config to manage the arguments: https://omegaconf.readthedocs.io/en/2.1_branch/structured_config.html#id1

At a first look, this will give you the benefits of:
- hierarchical data structures
- type safety
- handling of optional arguments

To get a sense of its usage:
- https://github.com/facebookresearch/multimodal/blob/2352bc5b63d30e7324a8547d2ee58eac4d48b163/examples/flava/definitions.py
- https://github.com/facebookresearch/multimodal/blob/main/examples/flava/configs/pretraining/debug.yaml

examples/mugen/retrieval/eval.py
```diff
@@ -4,114 +4,50 @@
...
-
-def get_args_from_group(args, group_name):
-    """Utility for grouping arguments based on the structure in ``arg_structure``"""
-    return argparse.Namespace(
-        **{
-            arg_name: getattr(args, arg_name)
-            for arg_name in arg_structure[group_name].keys()
-        }
-    )
+def get_yaml_config():
```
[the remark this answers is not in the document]
sophiazhi: this way, the user can pass in their own config filepath via cmd line if they don't want to use the one i made. i followed [flava](https://github.com/facebookresearch/multimodal/blob/main/examples/flava/utils.py#L30)'s example and [albef](https://github.com/facebookresearch/multimodal/blob/16cd6f9d750258adb417ac263263103b7c660c4f/examples/albef/finetune.py#L161) is doing something similar

examples/mugen/retrieval/eval.py
sophiazhi: I used `to_container` because when i unpacked the original config object it returned extra keys like `_metadata`, but I just tried it again and it gave me the expected keys so I'll switch it back

examples/mugen/retrieval/configs/eval.yaml
sophiazhi: the flava example follows the [hydra specs](https://hydra.cc/docs/advanced/instantiate_objects/overview/), so i updated the PR according to that

examples/mugen/retrieval/eval.py
```diff
@@ -22,33 +21,32 @@ def get_yaml_config():
...
-
-    trainer = Trainer(
-        accelerator=evaluation_args["accelerator"], devices=evaluation_args["devices"]
+    model = VideoCLIPLightningModule.load_from_checkpoint(
+        args.checkpoint_path,
+        **vars(args.lightningmodule_args),
+        **vars(args.videoclip_args)
     )
+
+    trainer = Trainer(accelerator="auto", devices=1)
```
langong347, line 49: Why using `auto` for inference? My understanding is that typically training is done on GPU but inference on CPU? I would also probably set `device` to be a parameter instead of hardcoding it --- unless there's a strong reason for `1`?

sophiazhi: I thought having the option of using gpu for evaluation would be convenient with the larger test set, but i'm not sure about the convention. I hardcoded `devices=1` because of the recommendation in the lightning docs (https://pytorch-lightning.readthedocs.io/en/stable/common/lightning_module.html#testing)

langong347: I am not 100% sure about how these convention (which makes sense to me) compare to our "typical" ways of doing things. I'd recommend you either write a post or seek others' feedback on this. 
Approved to unblock though.

langong347 (approved):

examples/mugen/retrieval/eval.py
sophiazhi: MUGEN authors used 1 gpu but it sounds like cpu is also a good option when batch size is small. So I'm thinking of making accelerator configurable, but leaving the default as "auto"

PR #234, opened 2022-08-01
Add clip vit encoder

ankitade:
Summary: CLIP vit encoder is slightly different from the general VIT with additional layer norm before first layer and quick gelu as activation function

Differential Revision: D38299221

### DIFF
--- a/torchmultimodal/modules/encoders/clip_vit_encoder.py
+++ b/torchmultimodal/modules/encoders/clip_vit_encoder.py
@@ -1,0 +1,105 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+import torch
+from torch import nn, Tensor
+
+# Taken from original clip implementation https://github.com/openai/CLIP/blob/main/clip/model.py#L167
+# TODO: unify with the implementation in text encoder
+def quick_gelu(x: Tensor) -> Tensor:
+    return x * torch.sigmoid(1.702 * x)
+
+
+class CLIPViTEncoder(nn.Module):
+    """
+    Vision transformer encoder for CLIP.
+
+    Args:
+        embedding_dim (int): Embedding dimension for text and image encoders projections.
+        patch_size (int): The dimension of each patch
+        image_size(int): The size (width==height) of input image
+        width (int): Dimensionality of the encoder layers and the pooler layer
+        heads (int): Number of attention heads for each attention layer in the Transformer encoder
+        layers (int): Number of hidden layers in the Transformer encoder
+
+    Inputs:
+        x (Tensor): image tensor with dimensions B x C(3) x image_size x image_size
+    """
+
+    def __init__(
+        self,
+        embedding_dim: int,
+        patch_size: int,
+        image_size: int,
+        width: int,
+        heads: int,
+        layers: int,
+    ):
+        super().__init__()
+        self.conv = nn.Conv2d(
+            in_channels=3,
+            out_channels=width,
+            kernel_size=patch_size,
+            stride=patch_size,
+            bias=False,
+        )
+        self.image_size = image_size
+
+        scale = width**-0.5
+        self.cls_token_embedding = nn.Parameter(scale * torch.randn(width))
+        self.positional_embedding = nn.Parameter(
+            scale * torch.randn((image_size // patch_size) ** 2 + 1, width)
+        )
+        self.ln_pre = nn.LayerNorm(width)
+        encoder_layer = nn.TransformerEncoderLayer(
+            d_model=width,
+            nhead=heads,
+            dropout=0.0,
+            activation=quick_gelu,
+            norm_first=True,
+            dim_feedforward=4 * width,
+            batch_first=True,
+        )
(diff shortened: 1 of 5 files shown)


torchmultimodal/modules/encoders/clip_vit_encoder.py
```diff
@@ -0,0 +1,105 @@
...
+    Args:
+        embedding_dim (int): Embedding dimension for text and image encoders projections.
+        patch_size (int): The dimension of each patch
+        image_size(int): The size (width==height) of input image
+        width (int): Dimensionality of the encoder layers and the pooler layer
+        heads (int): Number of attention heads for each attention layer in the Transformer encoder
+        layers (int): Number of hidden layers in the Transformer encoder
+
+    Inputs:
+        x (Tensor): image tensor with dimensions B x C(3) x image_size x image_size
```
RdoubleA, line 29: would it be worth having our own VisionTransformer? This is the third time we've had to customize the torchvision one (ALBEF changes one layer, FLAVA has another one from scratch)

torchmultimodal/modules/encoders/clip_vit_encoder.py
```diff
@@ -0,0 +1,105 @@
...
+        )
+        self.image_size = image_size
+
+        scale = width**-0.5
+        self.cls_token_embedding = nn.Parameter(scale * torch.randn(width))
+        self.positional_embedding = nn.Parameter(
+            scale * torch.randn((image_size // patch_size) ** 2 + 1, width)
+        )
+        self.ln_pre = nn.LayerNorm(width)
+        encoder_layer = nn.TransformerEncoderLayer(
```
RdoubleA, line 57: add TODO to replace this with unified one?

torchmultimodal/modules/encoders/clip_vit_encoder.py
```diff
@@ -0,0 +1,105 @@
...
+import torch
+from torch import nn, Tensor
+
+# Taken from original clip implementation https://github.com/openai/CLIP/blob/main/clip/model.py#L167
+# TODO: unify with the implementation in text encoder
+def quick_gelu(x):
+    return x * torch.sigmoid(1.702 * x)
+
+
+class CLIPViTEncoder(nn.Module):
```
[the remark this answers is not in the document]
ankitade: i plan to move in a separate diff

torchmultimodal/modules/encoders/clip_vit_encoder.py
ankitade: the clip vision transformer is a customized one with extra ln. I thought you were plannig to unify flava and albef one anyways, though i dont think this one needs to use it

torchmultimodal/modules/encoders/clip_vit_encoder.py
```diff
@@ -0,0 +1,105 @@
...
+            bias=False,
+        )
+        self.image_size = image_size
+
+        scale = width**-0.5
+        self.cls_token_embedding = nn.Parameter(scale * torch.randn(width))
+        self.positional_embedding = nn.Parameter(
+            scale * torch.randn((image_size // patch_size) ** 2 + 1, width)
+        )
+        self.ln_pre = nn.LayerNorm(width)
```
ebsmothers, line 56: I notice in MMF CLIP they do an fp32 layernorm here. Do we wanna do something similar? (Not sure if ViSE is using mixed precision or not)

torchmultimodal/modules/encoders/clip_vit_encoder.py
```diff
@@ -0,0 +1,105 @@
...
+
+        # B x C x patch_size x patch_size => B x C x patch_size ** 2
+        x = torch.flatten(x, start_dim=2)
+
+        # B x C x patch_size ** 2 => B x patch_size ** 2 x C
+        x = x.permute(0, 2, 1)
+        x = torch.cat(
+            [
+                self.cls_token_embedding.unsqueeze(0).expand(x.shape[0], -1, -1),
+                x,
```
ebsmothers, line 94: Probably a noob question but I'm confused by this part. Isn't the first element passed to the cat operation size `(B, 1, width)`? But then this doesn't match x's last dimension of C, does it?

torchmultimodal/modules/encoders/clip_vit_encoder.py
```diff
@@ -0,0 +1,105 @@
...
+            ],
+            dim=1,
+        )
+        x = x + self.positional_embedding
+        x = self.ln_pre(x)
+
+        x = self.encoder(x)
+
+        x = self.ln_post(x[:, 0, :])
```
ebsmothers, line 103: nit: add a comment explaining that we're only keeping the cls token embedding

ebsmothers (approved):
Looks good overall, just a few minor comments

torchmultimodal/modules/encoders/clip_vit_encoder.py
ankitade: yeah i saw that, was avoiding it till it was needed. but nvm, changed it now.

torchmultimodal/modules/encoders/clip_vit_encoder.py
ankitade: yeah the comments were confusing. C was referring to the output channel (not input) of the conv layer which is same as the width. changed it now. thanks for catching.

PR #235, opened 2022-08-01
[MUGEN] Fix VQVAE bugs when loading checkpoint

RdoubleA:
* #251
* #237
* #236
* __->__ #235

## Summary
Fixes some low-level bugs when attempting a forward pass after loading MUGEN's checkpoint.
- Output padding for `SamePadConv3dTranspose` layers were not getting calculated correctly because `self.pad_input` should be reversed. `self.pad_input` is used in `F.pad` which expects padding to be in reverse order, but this is not how it should be used in `calculate_tranpose_padding`.
- Embedding gets re-initialized after loading checkpoint because of a flag that's only set after first forward. Overload `_load_from_state_dict` from `nn.Module` to automatically switch the flag when a state dict is loaded. Added an assertion to the checkpoint test to make sure this is the case in #237.

## Test plan
`pytest test -vv`

Differential Revision: [D38524378](https://our.internmc.facebook.com/intern/diff/D38524378)


torchmultimodal/modules/layers/codebook.py
```diff
@@ -195,7 +191,7 @@ def _quantize(self, encoded_flat: Tensor) -> Tuple[Tensor, Tensor]:
 
     def forward(self, z: Tensor) -> CodebookOutput:
         # First check if embedding is initialized correctly
-        if not self._is_embedding_init and self.training:
+        if torch.sum(self.embedding).item() == 0 and self.training:
```
langong347, line 194: If you just remove `else`, we will go straight into `_preprocess` when loading the checkpoint, right?
```
if not self._is_embedding_init and self.training:
    encoded_flat, permuted_shape = self._init_embedding_and_preprocess(z)

# Reshape and flatten encoder output for quantization
encoded_flat, permuted_shape = self._preprocess(z)
```

RdoubleA: nope, loading the checkpoint will not switch the `_is_embedding_init` flag and it will still call `._init_embedding_and_preprocess`

torchmultimodal/modules/layers/codebook.py
```diff
@@ -213,3 +207,30 @@ def extra_repr(self) -> str:
         return "num_embeddings={}, embedding_dim={}".format(
             self.num_embeddings, self.embedding_dim
         )
+
+    def _load_from_state_dict(
```
langong347, line 211: Would probably adding a unit test here with fake inputs.
Reason being that it will give us early signal about not just the overloading logic but also the validity of the pytorch internal API.

RdoubleA: how about a unit test where I have a parent module that has `Codebook` as a child module, and I load in a fake state dict to see if this is still called and the flag is set?

langong347: Sounds good. Parent module -- you mean an instance of VQVAE?

langong347 (approved):

RdoubleA:
@RdoubleA has imported this pull request.  If you are a Facebook employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D38524378).

PR #247, opened 2022-08-05
VideoCLIP training script

sophiazhi:
Summary:
Training script for VideoCLIP on MUGEN dataset.

- `definitions.py`: add `TrainingArgs`, which includes lightning Trainer args like `max_epochs`
- `train.py`: training script
- `configs/train.yaml`: training config 
- `configs/eval.yaml`: fix default resolution value  ([default dataset args](https://github.com/mugen-org/MUGEN_baseline/blob/main/lib/data/mugen_data.py#L52), [eval command](https://github.com/mugen-org/MUGEN_baseline/blob/b45da493f3ddbcced52d44eee02d443f90842474/retrieval/experiments/eval_all.sh))
- `model.py`: remove recall calculation at end of training epoch, add split markers to recall logging (e.g. test/recall and validation/recall)

Test plan:
`python train.py config=configs/train.yaml`
[tensorboard](https://colab.research.google.com/drive/1RfZBYqpxmGiNCna-70_xm2ce85eEjB9I?usp=sharing) of logged results

### DIFF
--- a/examples/mugen/retrieval/model.py
+++ b/examples/mugen/retrieval/model.py
@@ -72,7 +72,7 @@ class VideoCLIPLightningModule(LightningModule):
         }
         return embeddings
 
-    def _compute_recall(self, text_embedding, video_embedding):
+    def _compute_recall(self, split, text_embedding, video_embedding):
         similarity_matrix = text_embedding @ video_embedding.T
         num_samples = similarity_matrix.shape[0]
         target_matrix = torch.eye(
@@ -82,11 +82,11 @@ class VideoCLIPLightningModule(LightningModule):
         for k in self.recall_ks:
             v2t_recall = self.metrics[f"v2t_recall_{k}"]
             v2t_recall(preds=similarity_matrix.T, target=target_matrix)
-            self.log(f"Recall@{k} (video query, text retrieval)", v2t_recall)
+            self.log(f"{split}/Recall@{k} (video query, text retrieval)", v2t_recall)
 
             t2v_recall = self.metrics[f"t2v_recall_{k}"]
             t2v_recall(preds=similarity_matrix, target=target_matrix)
-            self.log(f"Recall@{k} (text query, video retrieval)", t2v_recall)
+            self.log(f"{split}/Recall@{k} (text query, video retrieval)", t2v_recall)
 
     def configure_optimizers(self):
         params = self.parameters()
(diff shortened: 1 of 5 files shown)


examples/mugen/retrieval/model.py
```diff
@@ -106,15 +106,6 @@ def training_step(self, batch, batch_idx):
         )
         return {"loss": loss, "model_output": model_output}
 
-    def training_epoch_end(self, outputs):
```
langong347, line 109: Just to confirm: this is why previously recall was computed at every epoch?

examples/mugen/retrieval/train.py
```diff
@@ -47,7 +47,8 @@ def train():
 
     trainer = Trainer(
         accelerator=args.accelerator,
-        devices=1,
+        devices=args.devices,
+        strategy="ddp_find_unused_parameters_false",
```
langong347, line 51: How do we know there's no unused parameter?

langong347:
>fix default resolution value

The hyperlink there led to command line `python retrieval/eval.py --get_game_frame --get_audio --get_text_desc --use_manual_annotation`. Should we link to the dataset args instead?

https://github.com/mugen-org/MUGEN_baseline/blob/main/lib/data/mugen_data.py#L52

langong347 (approved):

examples/mugen/retrieval/model.py
sophiazhi: correct, that function is why recall was computed for the entire training set at every epoch (recall is still computed at every epoch for the validation set in `validation_epoch_end`)

examples/mugen/retrieval/train.py
sophiazhi: when i didn't have this argument, the script printed a warning saying no unused params were found and having `find_unused_parameters=True` makes performance worse

examples/mugen/retrieval/train.py
```diff
@@ -0,0 +1,67 @@
...
+    dataset_args: MUGENDatasetArgs = args.dataset_args
+    datamodule = MUGENDataModule(
+        dataset_args,
+        text_transform=BertTextTransform(
+            **vars(args.datamodule_args.bert_text_transform)
+        ),
+        video_transform=VideoTransform(**vars(args.datamodule_args.video_transform)),
+        batch_size=args.datamodule_args.batch_size,
+        num_workers=args.datamodule_args.num_workers,
+        shuffle=args.datamodule_args.shuffle,
```
langong347, line 41: Does this evaluate to `False` as the default in the yaml? In MUGEN `shuffle` is `True` during training.

langong347: https://github.com/mugen-org/MUGEN_baseline/blob/b45da493f3ddbcced52d44eee02d443f90842474/retrieval/train.py#L21

PR #250, opened 2022-08-05
Migrate clip resnet encoder test to pytest

sophiazhi:
Summary:
Migrate CLIP resnet encoder tests from `unittest` to `pytest`

Test plan:
`python -m pytest test/modules/encoders/test_clip_resnet_encoder.py` on cpu and gpu


test/modules/encoders/test_clip_resnet_encoder.py
```diff
@@ -4,29 +4,34 @@
...
+from test.test_utils import assert_expected
 from torchmultimodal.modules.encoders.clip_resnet_encoder import ResNetForCLIP
 from torchmultimodal.utils.common import get_current_device
 
 
-class TestCLIPModule(unittest.TestCase):
-    def setUp(self):
+class TestResnetEncoder:
+    @pytest.fixture(autouse=True)
+    def set_seed(self):
```
langong347, line 17: Although there's just one test class, having the seed fixture at the script level with `autouse` will ensure that each method (regardless of which TestClass it belongs to) will get a fresh seed. 

device could be universal fixture --- would recommend putting it at the script level as well.

Use `test.test_utils.set_rng_seed` here.

langong347 (approved):
Address review comments. Otherwise LGTM. Thanks for migrating this script.

PR #255, opened 2022-08-09
[Test] Add nested NamedTuple path to assert_expected_wrapper

langong347:
Stack from [ghstack](https://github.com/ezyang/ghstack):
* #260
* #259
* #258
* #257
* #256
* __->__ #255
* #254

Summary:
- `assert_expected_wrapper` is a unit test utility that validates actual and expected values for complex data structure
- currently it only validates sum and shape of the tensors but can potentially support validating tensors should we finalize our numerical testing design
- Add codepath for nested `NamedTuple` model outputs

Test Plan:
CI

Differential Revision: [D38621044](https://our.internmc.facebook.com/intern/diff/D38621044)

PR #254, opened 2022-08-09
[Lint] Correct type annotation in util torch_load_module

langong347:
Stack from [ghstack](https://github.com/ezyang/ghstack):
* #260
* #259
* #258
* #257
* #256
* #255
* __->__ #254

Summary:
Correct type annotation from `torch.nn.Module` to `nn.Module` in the utility function

Test Plan:

CI

Differential Revision: [D38621046](https://our.internmc.facebook.com/intern/diff/D38621046)

PR #256, opened 2022-08-09
[MUGEN] Add right shift op to MultimodalTransformerDecoder

langong347:
* #260
* #259
* #258
* #257
* __->__ #256
* #255
* #254

Summary:
- GPT model prepends start of sentence token after position embedding
- The token should only activate during training and when decoding the first modality token
- Adding control statement to manage the behavior with `right_shift` flag in the `forward` signature (to be invoked by the generation utility)

Test Plan:

```
$ python -m pytest test/models/test_gpt.py -k "right_shift" -vv
platform darwin -- Python 3.8.13, pytest-7.1.2, pluggy-1.0.0 -- /Users/langong/local/miniconda3/envs/t2v/bin/python
cachedir: .pytest_cache
rootdir: /Users/langong/gpt_attention, configfile: pyproject.toml
plugins: mock-3.8.2, cov-3.0.0
collected 18 items / 15 deselected / 3 selected

test/models/test_gpt.py::TestMultimodalTransformerDecoder::test_forward_eval_right_shift_on PASSED               [ 33%]
test/models/test_gpt.py::TestMultimodalTransformerDecoder::test_forward_eval_right_shift_off PASSED              [ 66%]
test/models/test_gpt.py::test_right_shift PASSED                                                                 [100%]

=========================================== 3 passed, 15 deselected in 1.34s ==========
```

Differential Revision: [D38621045](https://our.internmc.facebook.com/intern/diff/D38621045)


test/models/test_gpt.py
```diff
@@ -97,94 +127,152 @@ def decoder(decoder_layer, num_layers):
...
+    def test_forward_in_modality(self, mm_decoder, in_modality):
+        actual = mm_decoder(
+            in_modality=in_modality, in_pos_ids=self._pos_ids(in_modality)
+        )
         expected = {
             "last_hidden_states": (
                 torch.Size([1, 3, 4]),
-                -0.4598,
-            ),  # (bs, in_seq_len, emb_dim)
+                -0.5538,
```
ebsmothers, line 155: Why do the expected values change here? As far as I can tell the inputs are the same, so is it just the inclusion of RightShift?

torchmultimodal/models/gpt.py
```diff
@@ -123,6 +137,9 @@ def forward(
             x_out = self.out_token_emb(out_modality) + self.out_pos_emb(out_pos_ids)
             x = torch.cat((x_in, x_out), dim=1)
 
+        if self.training or right_shift:
```
ebsmothers, line 140: Noob question: when would someone want right_shift turned on if not in training mode?

langong347: Since we right shift during training mode, we will need to right shift the first token during generation to be consistent. For the latter because we are only interested in the first token, we will need to tell the generation utility (which takes the GPT model as input) to trigger it at `decode_step=0` and switch it off for the subsequent steps.

Context:
- MUGEN handles this by piping in `decode_step` throughout their transformer stack which not only makes the APIs look clunky but also introduces unnecessary branching codepaths inside, for example, the original `RightShift` forward pass was merely a passthrough when `decode_step>0` [here](https://github.com/mugen-org/MUGEN_baseline/blob/b45da493f3ddbcced52d44eee02d443f90842474/lib/models/gpt/attention.py#L318).

torchmultimodal/models/gpt.py
```diff
@@ -64,6 +73,8 @@ class MultimodalTransformerDecoder(nn.Module):
         use_cache (bool, optional): If ``True``, caches past key/value tensors for faster decoding. If ``False``,
             recomputes key and value for each decoding step. Defaults to ``False``.
         causal (bool. optional): If ``True``, use causal attention. Defaults to ``False``.
+        right_shift (bool): If ``True``, shifts the embedding vectors to the right and prepends it with start of
+            sentence token. Defaults to ``False``. This option is disregarded during training mode
```
RdoubleA, line 77: nit: maybe add a quick sentence saying it should be false for generation

torchmultimodal/models/gpt.py
```diff
@@ -123,6 +137,9 @@ def forward(
             x_out = self.out_token_emb(out_modality) + self.out_pos_emb(out_pos_ids)
             x = torch.cat((x_in, x_out), dim=1)
 
+        if self.training or right_shift:
+            x = self.right_shift(x)
```
RdoubleA, line 141: nit: maybe add a comment highlighting that the shape of x is now seq_len+1

RdoubleA (approved):

langong347:
@langong347 has imported this pull request.  If you are a Facebook employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D38621045).

PR #262, opened 2022-08-11
[Test] Enable numerical test and NamedTuple (for expected) in assert_expected_wrapper

langong347:
Stack from [ghstack](https://github.com/ezyang/ghstack):
* #264
* #257
* #263
* #259
* __->__ #262
* #258
* #256
* #255
* #254

Summary:
- Add support for numerical assertion 
- Add additional criteria for differentiating `tuple` from `namedtuple`
- Add `rtol`, `atol` configurability
- Can take both `Dict` and `NamedTuple` for actual and expected

Test Plan:
```
>>>from collections import namedtuple

>>> Out = namedtuple("Out", "x y")
>>> InnerOut = namedtuple("InnerOut", "z")

>>> actual = Out(x=InnerOut(z=tensor([1])), y=tensor([2]))
>>> expected = Out(x=InnerOut(z=tensor([1])), y=tensor([2]))

>>> assert_expected_wrapper(actual, expected)
```

```
$ python -m pytest test/models/test_gpt.py  -vv
================================================= test session starts ==================================================
platform darwin -- Python 3.8.13, pytest-7.1.2, pluggy-1.0.0 -- /Users/langong/local/miniconda3/envs/t2v/bin/python
cachedir: .pytest_cache
rootdir: /Users/langong/gpt_attention, configfile: pyproject.toml
plugins: mock-3.8.2, cov-3.0.0
collected 18 items

test/models/test_gpt.py::TestMultimodalGPT::test_bad_tokenizers PASSED                                           [  5%]
test/models/test_gpt.py::TestMultimodalGPT::test_fwd_for_generation PASSED                                       [ 11%]
test/models/test_gpt.py::TestMultimodalGPT::test_forward PASSED                                                  [ 16%]
test/models/test_gpt.py::TestMultimodalTransformerDecoder::test_bad_input PASSED                                 [ 22%]
test/models/test_gpt.py::TestMultimodalTransformerDecoder::test_forward_in_modality PASSED                       [ 27%]
test/models/test_gpt.py::TestMultimodalTransformerDecoder::test_forward_out_modality PASSED                      [ 33%]
test/models/test_gpt.py::TestMultimodalTransformerDecoder::test_forward_two_modality PASSED                      [ 38%]
test/models/test_gpt.py::TestMultimodalTransformerDecoder::test_forward_eval_right_shift_on PASSED               [ 44%]
test/models/test_gpt.py::TestMultimodalTransformerDecoder::test_forward_eval_right_shift_off PASSED              [ 50%]
[...]

PR #259, opened 2022-08-10
[BE] Migrate codebook unit test to pytest

langong347:
Stack from [ghstack](https://github.com/ezyang/ghstack):
* #264
* #257
* #263
* __->__ #259
* #262
* #258
* #256
* #255
* #254

Summary:
- `unittest` to `pytest`
- Use `assert_expected_wrapper` for `CodebookOutput` value assertion

Test Plan:
```
$ python -m pytest --cov=torchmultimodal/modules/layers/ test/modules/layers/test_codebook.py -vv
================================================= test session starts ==================================================
platform darwin -- Python 3.8.13, pytest-7.1.2, pluggy-1.0.0 -- /Users/langong/local/miniconda3/envs/t2v/bin/python
cachedir: .pytest_cache
rootdir: /Users/langong/gpt_attention, configfile: pyproject.toml
plugins: mock-3.8.2, cov-3.0.0
collected 11 items

test/modules/layers/test_codebook.py::TestCodebook::test_quantized_output PASSED                                 [  9%]
test/modules/layers/test_codebook.py::TestCodebook::test_preprocess PASSED                                       [ 18%]
test/modules/layers/test_codebook.py::TestCodebook::test_preprocess_channel_dim_assertion PASSED                 [ 27%]
test/modules/layers/test_codebook.py::TestCodebook::test_postprocess PASSED                                      [ 36%]
test/modules/layers/test_codebook.py::TestCodebook::test_init_embedding PASSED                                   [ 45%]
test/modules/layers/test_codebook.py::TestCodebook::test_ema_update_embedding PASSED                             [ 54%]
test/modules/layers/test_codebook.py::TestCodebook::test_register_buffer_tensors PASSED                          [ 63%]
test/modules/layers/test_codebook.py::TestCodebook::test_init_embedding_smaller_encoded PASSED                   [ 72%]
test/modules/layers/test_codebook.py::TestCodebook::test_codebook_restart PASSED                                 [ 81%]
test/modules/layers/test_codebook.py::TestCodebook::test_load_state_dict PASSED                                  [ 90%]
test/modules/layers/test_codebook.py::TestCodebook::test_lookup PASSED                                           [100%]

---------- coverage: platform darwin, python 3.8.13-final-0 ----------
Name                                                   Stmts   Miss  Cover
--------------------------------------------------------------------------
torchmultimodal/modules/layers/codebook.py                87      1    99%
--------------------------------------------------------------------------
TOTAL                                                    489    403    18%


[...]

PR #258, opened 2022-08-09
[MUGEN] Replace shape with latent_shape in position embedding

langong347:
Stack from [ghstack](https://github.com/ezyang/ghstack):
* #264
* #257
* #263
* #259
* #262
* __->__ #258
* #256
* #255
* #254

Summary:
- Make it clear that position embedding is added to the latent encodings not the input data
- Change `shape` to `latent_shape`

Test Plan:

CI

Differential Revision: [D38642050](https://our.internmc.facebook.com/intern/diff/D38642050)

PR #257, opened 2022-08-09
[MUGEN] Add MultimodalGPT Module

langong347:
Stack from [ghstack](https://github.com/ezyang/ghstack):
* __->__ #257
* #264

Summary:
- Defines the model architecture for the full multimodal GPT as the basis for the builder
- Defines the API for integration with generation utility
- Added `latent_shape` to reshape the token ids for decoding back to the real data
- Pulled token embedding layers out of `MultimodalTransformerDecoder` and put in `MultimodalGPT`.

Test Plan:

```
$ python -m pytest --cov=torchmultimodal/models/ test/models/test_gpt.py -vv
================================================= test session starts ==================================================
platform darwin -- Python 3.8.13, pytest-7.1.2, pluggy-1.0.0 -- /Users/langong/local/miniconda3/envs/t2v/bin/python
cachedir: .pytest_cache
rootdir: /Users/langong/gpt_attention, configfile: pyproject.toml
plugins: mock-3.8.2, cov-3.0.0
collected 20 items

test/models/test_gpt.py::TestMultimodalGPT::test_tokenizers_missing_methods PASSED       [  3%]
test/models/test_gpt.py::TestMultimodalGPT::test_encode_invalid_modality PASSED          [  7%]
test/models/test_gpt.py::TestMultimodalGPT::test_decode_tokens_wrong_shape PASSED        [ 11%]
test/models/test_gpt.py::TestMultimodalGPT::test_decode_tokens_reshape PASSED            [ 15%]
test/models/test_gpt.py::TestMultimodalGPT::test_lookup_invalid_modality PASSED          [ 19%]
test/models/test_gpt.py::TestMultimodalGPT::test_lookup_in_modality PASSED               [ 23%]
test/models/test_gpt.py::TestMultimodalGPT::test_lookup_out_modality PASSED              [ 26%]
test/models/test_gpt.py::TestMultimodalGPT::test_fwd_bad_input PASSED                    [ 30%]
test/models/test_gpt.py::TestMultimodalGPT::test_fwd_for_generation PASSED               [ 34%]
test/models/test_gpt.py::TestMultimodalGPT::test_forward PASSED                          [ 38%]
test/models/test_gpt.py::TestMultimodalGPT::test_forward_logits_mask PASSED              [ 42%]
test/models/test_gpt.py::TestMultimodalTransformerDecoder::test_bad_input PASSED         [ 46%]
[...]

PR #263, opened 2022-08-11
[BE] VQVAE unit test: retire class scope for fixtures

langong347:
* #264
* #257
* __->__ #263
* #259
* #262
* #258
* #256
* #255
* #254

Summary:
- Setting pytest fixtures to be of "class" scope is not necessary for small input data
- "class" scope will also make the test results dependent on the order of tests within the same "TestClass"

Test Plan:
CI

Differential Revision: [D38642046](https://our.internmc.facebook.com/intern/diff/D38642046)

### DIFF
--- a/test/models/test_vqvae.py
+++ b/test/models/test_vqvae.py
@@ -14,18 +14,47 @@ from torchmultimodal.models.vqvae import VQVAE
 from torchmultimodal.modules.layers.codebook import CodebookOutput
 
 
-@pytest.fixture(scope="class")
+@pytest.fixture(autouse=True)
 def random():
     set_rng_seed(4)
 
 
+@pytest.fixture
+def num_embeddings():
+    return 4
+
+
+@pytest.fixture
+def embedding_dim():
+    return 2
+
+
+@pytest.fixture
+def encoder():
+    enc = nn.Linear(2, 2, bias=False)
+    enc.weight = nn.Parameter(torch.tensor([[1.0, 2.0], [3.0, 4.0]]))
+    return enc
+
+
+@pytest.fixture
+def decoder():
+    dec = nn.Linear(2, 2, bias=False)
+    dec.weight = nn.Parameter(torch.tensor([[1.0, 2.0], [3.0, 4.0]]))
+    return dec
+
+
+@pytest.fixture
+def vqvae(encoder, decoder, num_embeddings, embedding_dim):
+    return VQVAE(encoder, decoder, num_embeddings, embedding_dim)
+
+
 class TestVQVAE:
-    @pytest.fixture(scope="class")
+    @pytest.fixture
     def decoder_input(self):
         d = torch.tensor([[[[3.0, 7.0], [6.0, 14.0]], [[9.0, 21.0], [12.0, 28.0]]]])
         return d
 
-    @pytest.fixture(scope="class")
+    @pytest.fixture
     def test_data(self, decoder_input):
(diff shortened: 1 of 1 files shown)


langong347:
@langong347 has imported this pull request.  If you are a Facebook employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D38642046).

test/models/test_vqvae.py
```diff
@@ -14,18 +14,47 @@
...
+
+@pytest.fixture
+def decoder():
+    dec = nn.Linear(2, 2, bias=False)
+    dec.weight = nn.Parameter(torch.tensor([[1.0, 2.0], [3.0, 4.0]]))
+    return dec
+
+
+@pytest.fixture
+def vqvae(encoder, decoder, num_embeddings, embedding_dim):
```
RdoubleA, line 47: nit: since this is vqvae specific, should we group it under the `TestVQVAE` class?

test/models/test_vqvae.py
```diff
@@ -39,26 +68,10 @@ def test_data(self, decoder_input):
             quantized_flat=torch.tensor(
                 [[3.0, 9.0], [7.0, 21.0], [6.0, 12.0], [14.0, 28.0]]
             ),
-            codebook_indices=torch.tensor([3, 2, 1, 0]),
+            codebook_indices=torch.tensor([1, 3, 0, 2]),
```
RdoubleA, line 71: just confirming - this changed because random_seed is now autouse, and the codebook indices are the only thing depending on randomization?

RdoubleA (approved):

langong347:
@langong347 has imported this pull request.  If you are a Facebook employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D38642046).

test/models/test_vqvae.py
langong347: Right. In the previous version only `vqvae` gets the seed but `encoder` and `decoder` (containing `nn.Linear`) have changed the random state as well. Making seed `autouse` ensures that each layer inside `vqvae` gets a fresh random state.

To verify, I reverted this PR and added `seed` to `encoder` and `decoder` which replicated the change in this PR:

```
E       AssertionError: actual: tensor([1, 3, 0, 2]), expected: tensor([3, 2, 1, 0])

FAILED test/models/test_vqvae.py::TestVQVAE::test_encode - AssertionError: actual: tensor([1, 3, 0, 2]), expected: te...
FAILED test/models/test_vqvae.py::TestVQVAE::test_forward - AssertionError: actual: tensor([1, 3, 0, 2]), expected: t..
```
```
    @pytest.fixture(scope="class")
    def encoder(self, random):
        enc = nn.Linear(2, 2, bias=False)
        enc.weight = nn.Parameter(torch.tensor([[1.0, 2.0], [3.0, 4.0]]))
        return enc

    @pytest.fixture(scope="class")
    def decoder(self, random):
        dec = nn.Linear(2, 2, bias=False)
        dec.weight = nn.Parameter(torch.tensor([[1.0, 2.0], [3.0, 4.0]]))
        return dec

    @pytest.fixture(scope="class")
    def vqvae(self, encoder, decoder, random):
        return VQVAE(encoder, decoder, 4, 2)
```

langong347:
@langong347 has imported this pull request.  If you are a Facebook employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D38642046).

PR #253, opened 2022-08-09
[Unification][FLAVA] Refactor FLAVA with unified text encoder

RdoubleA:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* #269
* #268
* #267
* #270
* __->__ #253
* #249

## Summary
- Remove FLAVA's text embeddings and text transformer classes
- Replace FLAVA's text transformer with general `TextEncoder`
- Update FLAVA checkpoint
- Update FLAVA tests with general `TransformerOutput` classes and `assert_expected_namedtuple`
- Update `TextEncoder` API with arguments needed to return attention weights, hidden states for FLAVA

## Test plan
`pytest test/models/flava -vv`

Differential Revision: [D38876763](https://our.internmc.facebook.com/intern/diff/D38876763)

PR #249, opened 2022-08-05
[Unification] Unified text encoder

RdoubleA:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* #269
* #268
* #267
* #270
* #253
* __->__ #249

## Summary
- Create general text transformer (`BERTTextEmbeddings` + `TransformerEncoder`)
- Replace ALBEF text encoder
- Replace MDETR text encoder
- Move ALBEF text encoder test to generic text encoder test since it's identical

## Test plan
`pytest test/models/mdetr -vv`
`pytest test/models/albef -vv`
`pytest test/modules/encoders/test_text_encoder.py -vv`

Differential Revision: [D38876739](https://our.internmc.facebook.com/intern/diff/D38876739)

PR #267, opened 2022-08-15
[Unification] Generalized vision transformer

RdoubleA:
* #269
* #268
* __->__ #267
* #270
* #253
* #249

## Summary
- Repurpose FLAVA's `ImageEmbeddings`, `PatchEmbeddings` and `ImageTransformer` as generic classes for unified image encoder
- Add tests for embeddings and image encoder

## Test plan
`pytest test/models/flava -vv`
`pytest test/modules/encoders/test_image_encoder.py -vv`
`pytest test/modules/layers/test_image_embedding.py -vv`

Verify that `torchvision.models.VisionTransformer` has the same output as our `VisionTransformer` with the same weights loaded. Ran on a local notebook and got identical outputs. (The FLAVA head has an additional tanh that I had to apply on top on torchvision's vision transformer)

image

### DIFF
--- a/torchmultimodal/modules/encoders/image_encoder.py
+++ b/torchmultimodal/modules/encoders/image_encoder.py
@@ -1,0 +1,88 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+from typing import Callable, Optional
+
+from torch import nn, Tensor
+from torchmultimodal.modules.layers.transformer import TransformerOutput
+
+
+class VisionTransformer(nn.Module):
+    """
+    General image transformer encoder with embeddings. Similar to ``VisionTransformer`` in torchvision,
+    but more composable. Can be constructed with any user-provided embeddings, encoder, and task head.
+
+    Attributes:
+        embeddings (nn.Module): Module that projects image pixels into embeddings.
+            ``forward()`` should follow interface:
+                images: Optional[Tensor], input data
+                image_patches_mask: Optional[Tensor], mask for patch embeddings
+        encoder (nn.Module): Module for transformer encoder. ``forward()`` should follow interface:
+            Inputs:
+                hidden_states: Tensor, input for encoder
+                attention_mask: Optional[Tensor], shape ``(b, num_heads, query_seq_len, key_seq_len)``
+                return_attn_weights: bool. See ``TransformerEncoder``.
+                return_hidden_states: bool. See ``TransformerEncoder``.
+            Returns:
+                ``TransformerOutput``
+        layernorm (nn.Module, optional): Module for layernorm to be applied after encoder, if provided.
+        pooler (nn.Module, optional): Module for head to be applied after layernorm, if provided.
+
+    Args:
+        images (Tensor): Tensor of input images of shape ``(b, c, h, w)``.
+        image_patches_mask (Tensor, optional): Tensor indicating which patches to replace with mask tokens,
+            shape ``(b, seq_len)``, where seq_len = (image_size // patch_size) ** 2
+        attention_mask (Tensor, optional): Tensor indicating which tokens to attend to, shape ``(b, seq_len + 1)``.
+            Concatenating class_token adds 1 to seq_len.
+    """
+
+    def __init__(
+        self,
+        embeddings: nn.Module,
+        encoder: nn.Module,
+        layernorm: Optional[nn.Module] = None,
+        pooler: Optional[nn.Module] = None,
+        weight_init_fn: Optional[Callable] = None,
+    ) -> None:
+        super().__init__()
+
+        self.embeddings = embeddings
+        self.encoder = encoder
+        self.layernorm = layernorm
+        self.pooler = pooler
+
+        if weight_init_fn:
+            self.apply(weight_init_fn)
+
+    def forward(
+        self,
+        images: Tensor,
+        image_patches_mask: Optional[Tensor] = None,
+        attention_mask: Optional[Tensor] = None,
+    ) -> TransformerOutput:
+
+        embedding_output = self.embeddings(
+            images, image_patches_mask=image_patches_mask
+        )
+
+        encoder_output = self.encoder(
+            embedding_output,
+            attention_mask=attention_mask,
+            return_attn_weights=True,
+            return_hidden_states=True,
+        )
+        sequence_output = encoder_output.last_hidden_state
+        sequence_output = self.layernorm(sequence_output)
+        pooled_output = (
+            self.pooler(sequence_output) if self.pooler is not None else None
+        )
+
+        return TransformerOutput(
+            last_hidden_state=sequence_output,
+            pooler_output=pooled_output,
+            hidden_states=encoder_output.hidden_states,
+            attentions=encoder_output.attentions,
+        )
--- a/torchmultimodal/modules/layers/image_embedding.py
+++ b/torchmultimodal/modules/layers/image_embedding.py
@@ -1,0 +1,163 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+import math
+import warnings
+from typing import Optional
+
+import torch
+from torch import nn, Tensor
+
+
+# Based on timm implementation, which can be found here:
+# https://github.com/rwightman/pytorch-image-models/blob/master/timm/models/vision_transformer.py
+class PatchEmbeddings(nn.Module):
+    """
+    Image to Patch Embedding.
+    """
+
+    def __init__(
+        self,
+        image_size: int = 224,
+        patch_size: int = 16,
+        num_channels: int = 3,
+        embed_dim: int = 768,
+    ) -> None:
+        super().__init__()
+        image_size = (image_size, image_size)
+        patch_size = (patch_size, patch_size)
+        num_patches = (image_size[1] // patch_size[1]) * (
+            image_size[0] // patch_size[0]
+        )
+        self.image_size = image_size
+        self.patch_size = patch_size
+        self.num_patches = num_patches
+
+        self.projection = nn.Conv2d(
+            num_channels, embed_dim, kernel_size=self.patch_size, stride=self.patch_size
+        )
+
+    def forward(
+        self, pixel_values: Tensor, interpolate_pos_encoding: bool = False
+    ) -> Tensor:
+        _, _, height, width = pixel_values.shape
+        if not interpolate_pos_encoding:
+            if height != self.image_size[0] or width != self.image_size[1]:
+                raise ValueError(
+                    f"Input image size ({height}*{width}) doesn't match model ({self.image_size[0]}*{self.image_size[1]})."
+                )
+        x = self.projection(pixel_values).flatten(2).transpose(1, 2)
+        return x
+
+
+class ImageEmbeddings(nn.Module):
+    """
+    Construct the CLS token, position and patch embeddings.
+    """
+
+    def __init__(
+        self,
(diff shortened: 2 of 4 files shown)


ebsmothers (commented):
One high-level comment about this change. While I agree that the current TorchVision implementation is not the most composable, the current ALBEF code shows that it is pretty easy to extract the intermediate layers (i.e. before the pooler). I am less familiar with all the details of FLAVA's ViT, but wonder if a similar approach could also be applied there. 

Basically I am wondering why we want to reimplement the entire VisionTransformer class ourselves for the sake of modularity when there is already a pretty viable solution in an upstream library. (Btw I am not claiming there is a clear answer one way or the other here, as we can justify either of these approaches with a different one of our design principles.)

RdoubleA:
> One high-level comment about this change. While I agree that the current TorchVision implementation is not the most composable, the current ALBEF code shows that it is pretty easy to extract the intermediate layers (i.e. before the pooler). I am less familiar with all the details of FLAVA's ViT, but wonder if a similar approach could also be applied there. 
> 
> 
> 
> Basically I am wondering why we want to reimplement the entire VisionTransformer class ourselves for the sake of modularity when there is already a pretty viable solution in an upstream library. (Btw I am not claiming there is a clear answer one way or the other here, as we can justify either of these approaches with a different one of our design principles.)

This is a fair point, I am on the fence with this. I do know that there will be a CLIP vision transformer where torchvision's may be insufficient (@ankitade would know more), so thinking long term it may be good to just have our own implementation. And based on feedback on prior changes, it seemed like we want to favor consistency within our own library first.

As for if it's possible to use torchvision's ViT in FLAVA, the main issue is that the entire stack is already unified with our own implementations (encoder, layer, MHA, self attention) so changing to torchvision's would undo all that. That may be okay depending on if we want to prioritize using our own components or not.

This may be worth a post to get more feedback from the team.

torchmultimodal/modules/encoders/image_encoder.py
```diff
@@ -0,0 +1,88 @@
...
+    """
+    General image transformer encoder with embeddings. Similar to ``VisionTransformer`` in torchvision,
+    but more composable. Can be constructed with any user-provided embeddings, encoder, and task head.
+
+    Attributes:
+        embeddings (nn.Module): Module that projects image pixels into embeddings.
+            ``forward()`` should follow interface:
+                images: Optional[Tensor], input data
+                image_patches_mask: Optional[Tensor], mask for patch embeddings
+        encoder (nn.Module): Module for transformer encoder. ``forward()`` should follow interface:
```
langong347, line 23: The same comment as in #249 to leverage cross reference to avoid hard-documenting the interface.

torchmultimodal/modules/encoders/image_encoder.py
```diff
@@ -0,0 +1,88 @@
...
+            Concatenating class_token adds 1 to seq_len.
+    """
+
+    def __init__(
+        self,
+        embeddings: nn.Module,
+        encoder: nn.Module,
+        layernorm: Optional[nn.Module] = None,
+        pooler: Optional[nn.Module] = None,
+        weight_init_fn: Optional[Callable] = None,
```
langong347, line 48: The same comment as in #249 to provide more details/docstring about this fn.

torchmultimodal/modules/encoders/image_encoder.py
```diff
@@ -0,0 +1,88 @@
...
+        image_patches_mask (Tensor, optional): Tensor indicating which patches to replace with mask tokens,
+            shape ``(b, seq_len)``, where seq_len = (image_size // patch_size) ** 2
+        attention_mask (Tensor, optional): Tensor indicating which tokens to attend to, shape ``(b, seq_len + 1)``.
+            Concatenating class_token adds 1 to seq_len.
+    """
+
+    def __init__(
+        self,
+        embeddings: nn.Module,
+        encoder: nn.Module,
```
langong347, line 45: `transformer_encoder`? but either way I think is fine as long as we are being clear in the docstring as to what's allowed.

torchmultimodal/modules/encoders/image_encoder.py
```diff
@@ -0,0 +1,88 @@
...
+            images, image_patches_mask=image_patches_mask
+        )
+
+        encoder_output = self.encoder(
+            embedding_output,
+            attention_mask=attention_mask,
+            return_attn_weights=True,
+            return_hidden_states=True,
+        )
+        sequence_output = encoder_output.last_hidden_state
```
langong347, line 77: The comment as in #249 use `last_hidden_states` as the variable name here?

langong347:
>And based on feedback on prior changes, it seemed like we want to favor consistency within our own library first.

A different perspective here: it really depends on the model. In some cases having our own version is the better choice; other cases (for more stable model) upstreaming to Torch* libraries will benefit the ecosystem to reduce fragmentation. One example is `S3D`:  we initially implemented our own version but are now upstreaming to TorchVision since it's just a variant of the base model (inception 3D) they already have.

langong347:
>it's mainly just moving the vision encoder stuff out of flava folder to the general core

Maybe we can just put `PatchEmbedding`, `ImageEmbedding` and `VisionTransformer` from flava into  `torchmultimodal/modules/encoders/flava_vision_encoder.py` (transformer can be considered as a type of encoder):
- the three components have linear dependency
- there's no evidence yet about stability/extensibility of those components

For example, omnivore puts its patch-embedding and transformer encoder together in one module [here](https://github.com/facebookresearch/multimodal/blob/main/torchmultimodal/modules/encoders/swin_transformer_3d_encoder.py).

torchmultimodal/modules/encoders/image_encoder.py
```diff
@@ -0,0 +1,81 @@
...
+        )
+
+        encoder_output = self.encoder(
+            embedding_output,
+            attention_mask=attention_mask,
+            return_attn_weights=True,
+            return_hidden_states=True,
+        )
+        last_hidden_state = encoder_output.last_hidden_state
+        last_hidden_state = self.layernorm(last_hidden_state)
```
langong347, line 71: Check if `layernorm is None`?

torchmultimodal/modules/encoders/image_encoder.py
```diff
@@ -0,0 +1,81 @@
...
+    but more composable. Can be constructed with any user-provided embeddings, encoder, and task head.
+
+    Attributes:
+        embeddings (nn.Module): Module that projects image pixels into embeddings.
+            See :py:class: ImageEmbeddings for interface.
+        encoder (nn.Module): Module for transformer encoder. See :py:class: TransformerEncoder for interface.
+        layernorm (nn.Module, optional): Module for layernorm to be applied after encoder. Defaults to ``None``.
+        pooler (nn.Module, optional): Module for pooler to be applied after layernorm. Defaults to ``None``.
+        weight_init_fn (Callable, optional): function for custom weight initialization of both the transformer
+            encoder and embeddings. See :py:func: init_transformer_weights as an example. Defaults to ``None``.
```
langong347, line 25: The same comment about `py:class` as in https://github.com/facebookresearch/multimodal/pull/249#discussion_r949665113

torchmultimodal/modules/encoders/image_encoder.py
```diff
@@ -0,0 +1,81 @@
...
+
+    Args:
+        images (Tensor): Tensor of input images of shape ``(b, c, h, w)``.
+        image_patches_mask (Tensor, optional): Tensor indicating which patches to replace with mask tokens,
+            shape ``(b, seq_len)``, where seq_len = (image_size // patch_size) ** 2
```
langong347, line 30: What do you mean by `(image_size // patch_size) ** 2`?
Are `image_size` and `patch_size` tuples or numbers?
Consider using math notations if needed? Something like:
```
:math:`seq_len = ....`
```

torchmultimodal/modules/layers/image_embedding.py
```diff
@@ -0,0 +1,163 @@
...
+        )
+
+        _, seq_len, _ = embeddings.size()
+        if image_patches_mask is not None:
+            if self.mask_token is not None:
+                mask_tokens = self.mask_token.expand(batch_size, seq_len, -1)
+                # replace the masked visual tokens by mask_tokens
+                w = image_patches_mask.unsqueeze(-1).type_as(mask_tokens)
+                embeddings = embeddings * (1 - w) + mask_tokens * w
+            else:
```
langong347, line 145: Can we add a test for the `else` path?

torchmultimodal/modules/layers/image_embedding.py
```diff
@@ -0,0 +1,163 @@
...
+
+        self.projection = nn.Conv2d(
+            num_channels, embed_dim, kernel_size=self.patch_size, stride=self.patch_size
+        )
+
+    def forward(
+        self, pixel_values: Tensor, interpolate_pos_encoding: bool = False
+    ) -> Tensor:
+        _, _, height, width = pixel_values.shape
+        if not interpolate_pos_encoding:
```
langong347, line 47: Add a test for this codepath?

torchmultimodal/modules/encoders/image_encoder.py
```diff
@@ -0,0 +1,81 @@
...
+from typing import Callable, Optional
+
+from torch import nn, Tensor
+from torchmultimodal.modules.layers.transformer import TransformerOutput
+
+
+class VisionTransformer(nn.Module):
+    """
+    General image transformer encoder with embeddings. Similar to ``VisionTransformer`` in torchvision,
+    but more composable. Can be constructed with any user-provided embeddings, encoder, and task head.
```
ebsmothers, line 16: nit: and pooler

torchmultimodal/modules/encoders/image_encoder.py
```diff
@@ -0,0 +1,81 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
```
ebsmothers, line 1: rename to vision_transformer.py?

torchmultimodal/modules/encoders/image_encoder.py
ebsmothers, line 16: another nit: I would take out the "more composable" part. Imo it's not necessary to explicitly say this

torchmultimodal/modules/layers/image_embedding.py
```diff
@@ -0,0 +1,163 @@
...
+            if height != self.image_size[0] or width != self.image_size[1]:
+                raise ValueError(
+                    f"Input image size ({height}*{width}) doesn't match model ({self.image_size[0]}*{self.image_size[1]})."
```
ebsmothers, line 50: nit: either write shape as a tuple or use `x` instead of `*`

torchmultimodal/modules/layers/image_embedding.py
```diff
@@ -0,0 +1,163 @@
...
+            torch.zeros(1, num_patches + 1, hidden_size)
+        )
+        self.dropout = nn.Dropout(hidden_dropout_prob)
+
+        if use_image_masking:
+            self.mask_token = nn.Parameter(torch.zeros(1, 1, hidden_size))
+        else:
+            self.mask_token = None
+
+    def interpolate_pos_encoding(
```
ebsmothers, line 90: What's the use case for this? I see it's added as an option for FLAVA but don't see it ever being used. cc @ankitade

torchmultimodal/modules/encoders/image_encoder.py
langong347: Do you mean it's implied by the API that we don't have to explicitly say this? 
On the other hand, with docstring the wording should be descriptive while "more XXX" can feel a bit opinionated ? :)

torchmultimodal/modules/encoders/image_encoder.py
langong347: `image_encoder.py` could be just a script for all types of image related encoders where `VisionTransformer` is just one of them? Rafi also has `text_encoder.py` with `BERTTextXXX`.

PR #279, opened 2022-08-22
[BE] Migrate some tests to pytest

ebsmothers:
Summary: Migrated FLAVA, two tower, late fusion, and CNNLSTM tests from unittest to pytest.
Kept initialization order the same so we don't have to change any expected values in this PR (most relevant for FLAVA tests).

Test plan:
```
python -m pytest -v
```

### DIFF
--- a/test/models/flava/test_flava.py
+++ b/test/models/flava/test_flava.py
@@ -4,10 +4,10 @@
 # This source code is licensed under the BSD-style license found in the
 # LICENSE file in the root directory of this source tree.
 
-import unittest
+import pytest
 
 import torch
-from test.test_utils import assert_expected_namedtuple
+from test.test_utils import assert_expected, assert_expected_namedtuple, set_rng_seed
 from torch import nn
 from torchmultimodal.models.flava.model import (
     flava_image_encoder,
@@ -22,34 +22,21 @@ from torchmultimodal.modules.layers.transformer import TransformerOutput
 NUM_CLASSES = 2
 
 
-class TestFLAVA(unittest.TestCase):
-    def setUp(self):
-        torch.manual_seed(1234)
+@pytest.fixture(autouse=True)
+def random():
+    set_rng_seed(1234)
 
-    @torch.no_grad()
-    def test_forward_classification(self):
+
+class TestFLAVA:
+    @pytest.fixture
+    def classification_inputs(self):
         text = torch.randint(0, 30500, (2, 77), dtype=torch.long)
         image = torch.rand((2, 3, 224, 224))
-
         labels = torch.randint(0, 2, (2,), dtype=torch.long)
+        return text, image, labels
 
-        flava = flava_model_for_classification(NUM_CLASSES, pretrained_model_key=None)
-        flava.eval()
-
-        # Test multimodal scenario
-        output = flava(image, text, "mm", labels)
-        self.assertAlmostEqual(output.loss.item(), 0.7180, places=4)
-
-        # Test unimodal image scenario
-        output = flava(image, text, "image", labels)
-        self.assertAlmostEqual(output.loss.item(), 0.7020, places=4)
-
-        # Test unimodal text scenario
-        output = flava(image, text, "text", labels)
-        self.assertAlmostEqual(output.loss.item(), 0.6663, places=4)
-
-    @torch.no_grad()
-    def test_forward_pretraining(self):
+    @pytest.fixture
+    def pretraining_inputs(self):
         text = torch.randint(0, 30500, (2, 77), dtype=torch.long)
         image = torch.rand((2, 3, 224, 224))
         image_for_codebook = torch.rand(2, 3, 112, 112)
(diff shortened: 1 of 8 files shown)


test/models/flava/test_flava.py
```diff
@@ -155,50 +188,84 @@ def setUp(self):
...
             multimodal=TransformerOutput(),
-            text=self.text_encoder(
+            text=text_encoder(
                 text, return_attn_weights=True, return_hidden_states=True
             ),
-            image=self.image_encoder(image),
-            image_masked=self.image_encoder(image),
+            image=image_encoder(image),
+            image_masked=image_encoder(image),
             projected_image_embeddings=actual.image.last_hidden_state[:, 0, :],
```
langong347, line 252: `expected` depends on `actual`? !

```
projected_image_embeddings=actual.image.last_hidden_state[:, 0, :],
projected_text_embeddings=actual.text.last_hidden_state[:, 0, :],
```

Generally, it'd be good not to use components we test to calculate expected even if the values cannot be computed by hand to ensure isolation. For example, `text`, `image` and `image_masked` fields are computed from the low level components which are part of flava model. 

I'd suggest at least remove the explicit dependency on `actual` for the last two fields (and check if there're similar cases elsewhere). Your call for those that are derived from low level components.

test/models/flava/test_flava.py
```diff
@@ -209,43 +276,44 @@ def test_forward_masked_image_and_text(self):
...
             multimodal=TransformerOutput(),
-            text=self.text_encoder(
+            text=text_encoder(
                 text, return_attn_weights=True, return_hidden_states=True
             ),
-            image=self.image_encoder(image),
-            image_masked=self.image_encoder(image, masked_image),
+            image=image_encoder(image),
+            image_masked=image_encoder(image, masked_image),
             projected_image_embeddings=actual.image.last_hidden_state[:, 0, :],
```
langong347, line 284: remove `expected` dep on `actual`

langong347 (commented):
Thanks for migrating those tests!

test/models/flava/test_flava.py
```diff
@@ -72,18 +99,19 @@ def test_forward_pretraining(self):
...
             mlm_labels=mlm_labels,
         )
-        self.assertIsNone(output.mlm_output)
-        self.assertIsNone(output.mim_output)
-        self.assertIsNotNone(output.global_contrastive_output)
-        self.assertIsNotNone(output.mmm_text_output)
-        self.assertIsNotNone(output.mmm_image_output)
-        self.assertIsNotNone(output.itm_output)
-        self.assertAlmostEqual(
+        assert output.mlm_output is None
```
langong347, line 102:
```
expected = {
    "mlm_output": None,
    "mim_output": None,
    "global_constrastive_output": (size, sum),
    "losses": (size, 21.5150),
}
assert_expected_namedtuple(actual, expected, rtol=1e-5, atol=1e-4)
```

test/models/flava/test_flava.py
ebsmothers: Yeah this is a good point. Let me derive these separately without relying on `actual`

test/models/flava/test_flava.py
```diff
@@ -22,44 +22,71 @@
...
+            text,
+            image,
+            image_for_codebook,
+            image_patches_mask,
+            text_masked,
+            mlm_labels,
+            itm_labels,
+        )
+
+    @torch.no_grad()
```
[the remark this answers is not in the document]
ebsmothers: In this case it may actually be useful to keep it; the pretraining test is using the full FLAVA model architecture and is one of our slower tests. I will remove the `no_grad` decorator and test locally as a sanity check though

ebsmothers:
> By order of initialization do you mean the order of the fixtures for each model layer? That shouldn't matter anymore if:

@langong347 I mean that for some existing tests multiple components are initialized with random weights without resetting the seed between (e.g. [here](https://github.com/facebookresearch/multimodal/blob/main/test/models/flava/test_flava.py#L31-L34)). To avoid remapping the expected values in this PR I initialize any such components in the same fixture to maintain the same seed status. Lmk if you have an alternative suggestion here though

test/models/flava/test_flava.py
```diff
@@ -155,50 +188,84 @@ def setUp(self):
...
+        self,
+        image_encoder,
+        text_encoder,
+        mm_encoder,
+        image_to_mm_projection,
+        text_to_mm_projection,
+        text_projection,
+        image_projection,
+    ):
+        return FLAVAModel(
```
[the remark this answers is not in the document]
ebsmothers: I think the behavior is the same in both cases. FLAVA sets dropout prob to 0.0 and uses layer norm, which is unaffected by eval mode. So we could add it but I don't think it would actually change anything

test/models/flava/test_flava.py
ebsmothers: Ultimately decided to leave this as is, mainly because the test is currently not explicitly checking sizes and values of all these components and the final loss value check should be sufficient here

test/models/flava/test_flava.py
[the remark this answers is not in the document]
langong347: I think as a good practice we should add it because the model may at some point introduce randomness. What do you think?

langong347:
> I mean that for some existing tests multiple components are initialized with random weights without resetting the seed between (e.g. [here](https://github.com/facebookresearch/multimodal/blob/main/test/models/flava/test_flava.py#L31-L34)). To avoid remapping the expected values in this PR I initialize any such components in the same fixture to maintain the same seed status. Lmk if you have an alternative suggestion here though

Since the test data are defined in the test module, their order of initialization we have full visibilty.
My comment was more around the initialization of model layers where a fresh random seed in between the fixtures can be helpful to shield impact from trivial refactoring (e.g., reordering of model layers in the model class).

test/models/flava/test_flava.py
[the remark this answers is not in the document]
ebsmothers: I will just add it so it doesn't lead to any confusion in the future

langong347 (approved):

PR #282, opened 2022-08-23
[MUGEN] Add Auto Regressive Generation Utility

langong347:
* __->__ #282
* #292

Summary:
- Refactored the [generation](https://github.com/mugen-org/MUGEN_baseline/blob/main/lib/models/gpt/gpt.py#L200) logic from MUGEN `MultimodalGPT` lightning module
- The AR decoding utility is re-designed as a layer of abstraction on top of the GPT model for improved modularity
- This utility can potentially be generalized for other generation models if they satisfy certain API requirements

Test Plan:

```
$ python -m pytest --cov=torchmultimodal/utils/ test/utils/test_generate.py -vv

test/utils/test_generate.py::TestLogitsMask::test_normal PASSED                                                  [  6%]
test/utils/test_generate.py::TestLogitsMask::test_zero_dims PASSED                                               [ 13%]
test/utils/test_generate.py::TestLogitsMask::test_in_seq_only PASSED                                             [ 20%]
test/utils/test_generate.py::TestLogitsMask::test_out_seq_only PASSED                                            [ 26%]
test/utils/test_generate.py::TestGenerationUtil::test_model_eval_warning PASSED                                  [ 33%]
test/utils/test_generate.py::TestGenerationUtil::test_sample PASSED                                              [ 40%]
test/utils/test_generate.py::TestGenerationUtil::test_filter_logits PASSED                                       [ 46%]
test/utils/test_generate.py::TestLogitsFilterTopK::test_min_tokens_to_keep PASSED                                [ 53%]
test/utils/test_generate.py::TestLogitsFilterTopK::test_top_k_invalid PASSED                                     [ 60%]
test/utils/test_generate.py::TestLogitsFilterTopK::test_default PASSED                                           [ 66%]
test/utils/test_generate.py::TestLogitsFilterTopK::test_top_k PASSED                                             [ 73%]
test/utils/test_generate.py::TestLogitsFilterTopP::test_min_tokens_to_keep PASSED                                [ 80%]
test/utils/test_generate.py::TestLogitsFilterTopP::test_top_p_invalid PASSED                                     [ 86%]
test/utils/test_generate.py::TestLogitsFilterTopP::test_default PASSED                                           [ 93%]
test/utils/test_generate.py::TestLogitsFilterTopP::test_top_p PASSED                                             [100%]

---------- coverage: platform darwin, python 3.8.13-final-0 ----------
Name                                 Stmts   Miss  Cover
--------------------------------------------------------
torchmultimodal/utils/generate.py       90      0   100%

========================= 15 passed in 3.17s ==========================
```

Differential Revision: [D38932535](https://our.internmc.facebook.com/intern/diff/D38932535)


langong347:
@langong347 has imported this pull request.  If you are a Facebook employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D38932535).

langong347:
@langong347 has imported this pull request.  If you are a Facebook employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D38932535).

torchmultimodal/utils/generate.py
```diff
@@ -0,0 +1,212 @@
...
+from typing import Any, Optional
+
+import torch
+from torch import nn, Tensor
+from torch.nn import functional as F
+
+from torchmultimodal.utils.attention import get_causal_attention_mask
+
+
+class Generation:
```
ebsmothers, line 17: High-level q: why make this a class and not a function/script?

torchmultimodal/utils/generate.py
```diff
@@ -0,0 +1,212 @@
...
+    total_seq_len = in_seq_len + out_seq_len
+    num_tokens = num_in_tokens + num_out_tokens
+    logits_mask = torch.ones(total_seq_len, num_tokens)
+    logits_mask[in_seq_len:, :num_in_tokens] = 0
+    logits_mask[:in_seq_len, num_in_tokens:] = 0
+
+    return logits_mask  # (total_seq_len, total_num_tokens)
+
+
+def top_k_top_p_filtering(
```
ebsmothers, line 159: This utility is a bit confusing to me... so users will want to perform both top-k and top-p filtering for the same set of predictions? Aren't these two non-commutative operations?

torchmultimodal/utils/generate.py
```diff
@@ -0,0 +1,212 @@
...
+
+        # Feed the input modality tokens `(b, in_seq_len)` through the model's transformer to learn
+        # the intermediate context vectors (i.e., key/value).
+        # The sequence is shifted to the right by one unit so that the predicted token at each location
+        # along the sequence is based off the previous token.
+        # Note that the first token is predicted from the learnt start-of-sentence ("sos") token which
+        # gets prepended to the sequence after the position embedding layer.
+        # Attention mask is required to avoid attending to future positions along the sequence.
+        # See :class:`~torchmultimodal.models.gpt.RightShift` for more implementation details.
+        self.model.fwd(  # type: ignore
```
ebsmothers, line 88: No return value?

torchmultimodal/utils/generate.py
```diff
@@ -0,0 +1,212 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
```
ebsmothers, line 1: Why put this in utils? It feels like more of an example script

torchmultimodal/utils/generate.py
langong347: I was on the fence about it. We can turn this back into a function but in general "Generation" can involve several techniques other than multinomial sampling (which is what we are doing now) such that `Class` will be more appropriate for sharing of common attributes / methods.

torchmultimodal/utils/generate.py
langong347: I think the more common scenarios would be either one of the two. But there's nothing wrong to pick top-k logits before picking amongst the remainder the top-p cumulative probability. what do you think?

torchmultimodal/utils/generate.py
langong347: Right this is intentional because for the input modality (or prompt sequence) we are only interested in learning the intermediate key/value representations.

So after calling `fwd` the codepath below will cache the key/value matrices `(batch, head, in_seq_len, d_model)`:

https://github.com/facebookresearch/multimodal/blob/8fc2483290fa03e6b0d8d4ec56a57e43de7406e4/torchmultimodal/modules/layers/attention.py#L218-L231

The cached values will be attended to element-wise during AR decoding for the output-modality.

Might be easier to visualize the internal process by looking at this unit test for GPT model here:

https://github.com/facebookresearch/multimodal/blob/8fc2483290fa03e6b0d8d4ec56a57e43de7406e4/test/models/test_gpt.py#L321-L362

torchmultimodal/utils/generate.py
langong347: The idea is that any GPT-like model with AR decoding feature can be plugged into this utility to generate data.
It is more of a layer of abstraction rather than a full blown script with project-specific details such as datamodule and etc. 

For example:
- Huggingface designs the similar component as a model mixin: https://github.com/huggingface/transformers/blob/v4.21.2/src/transformers/generation_utils.py#L379
- Fairseq does so by enforcing a model to inherit from `FairseqDecoder`: https://fairseq.readthedocs.io/en/latest/_modules/fairseq/models/fairseq_incremental_decoder.html

I am just taking the middle route (i.e., as a utility function) that is neither a mixin nor through inheritance.

torchmultimodal/utils/generate.py
ebsmothers: Yeah I think it's reasonable to use both. My comment is more around how we structure it. Since the order in which we apply top-k and top-p matters, maybe it's better to not combine the two into a single function like this, and instead just apply them separately in the main script.

torchmultimodal/utils/generate.py
ebsmothers: Makes sense. Fwd does return something here, right? Maybe just add `_ = self.model.fwd(...)` to make it explicit that we are not using the results.

torchmultimodal/utils/generate.py
ebsmothers: Thanks for the explanation. It would be nice to see a toy example of how this utility fits together with model class and other components like datamodule. But not really a concern for this PR, mainly just something that would help my own understanding

langong347: Yeah once I load the checkpoint will show that along with the dataset.
Right now the conceptual idea can be seen from the unit test where `model` is `video_gpt` and data is random inputs:

https://github.com/facebookresearch/multimodal/blob/a896bf7b952d08957ea279c05af66c2643dbdc9b/test/utils/test_generate.py#L65-L83

Note that the patching of multinomial is needed here because the model weights are random (i.e., unconditioned on any training data) so sampling can spit out tokens that are out side of the "vocabulary". Therefore, we force sampling to return always `0` token while still being able to test the generation loop.

langong347:
@langong347 has imported this pull request.  If you are a Facebook employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D38932535).

torchmultimodal/utils/generate.py
```diff
@@ -0,0 +1,277 @@
...
+            sorted_indices_to_remove[..., : self.min_tokens_to_keep] = 0
+        # Shift the indices to the right to keep also the first token above the threshold
+        sorted_indices_to_remove[..., 1:] = sorted_indices_to_remove[..., :-1].clone()
+        sorted_indices_to_remove[..., 0] = 0
+
+        # scatter sorted tensors to original indexing
+        indices_to_remove = sorted_indices_to_remove.scatter(
+            1, sorted_indices, sorted_indices_to_remove
+        )
+        logits[indices_to_remove] = self.filter_value
```
ebsmothers, line 275: This feels needlessly complicated, can we just apply top-k and top-p separately? Torch already has a builtin for top-k [here](https://pytorch.org/docs/stable/generated/torch.topk.html), and top-p should also just be a simple filter

langong347: By "separately" do you mean to split this into two classes one for `top_k` one for `top_p`?

While that's possible but in regarding your second point:

We are calling pytorch's `top_k` underneath but you would still need 1) sanity checking to ensure minimum number of tokens still exist after the filtering 2) assign `max_neg_value` to the unwanted tokens (so they are excluded from sampling without changing the size of the logit matrix).

https://github.com/facebookresearch/multimodal/blob/6ca2b5e2147ed3aaae996208c8b8e028416c34c8/torchmultimodal/utils/generate.py#L245-L250

langong347: >top-p should also just be a simple filter

I am not aware if there's an out-of-box top-p filter we can use here.

ebsmothers: Thanks for the additional information here. I am still envisioning a case where users can pass both top-k and top-p, but that each has its own function (rather than being combined into a single class). I think this is ultimately what HF does anyways: [ref](https://github.com/huggingface/transformers/blob/169b8cde471ec48e557f4c2d0c8e0d40abb13040/src/transformers/generation_logits_process.py#L212), but as you mentioned they do add a utility function to combine each strategy. Regarding your suggestion:

> We can create individual classes for different filtering strategies
We still need a topping layer to manage the combination of strategies based on user inputs

Why do we need a separate class for managing combinations? Seems it could just be handled directly in the generation class (which already has access to the `top_k` and `top_p` values). Then imo the behavior is clearer anyways, we can see directly from the main class the roles of `top_k` and `top_p`, whereas otherwise they are obscured by the `LogitsFilter` class.

langong347: I agree the selection logic fits more naturally to `GenerateUtil` given the latter already has the knobs (i.e. ,`top_k`, `top_p`). So:
 
 - Move the selection logic into `GenerateUtil` exposing the "knobs"
 - Convert `_filter_top_k` and `_filter_top_p` as two separate classes
 - `top_p` we will keep the current implementation as there's no out-of-box tool from pytorch core

let me know if we are in line now.

torchmultimodal/utils/generate.py
langong347: >Since the order in which we apply top-k and top-p matters

Thought I missed the point of this question: looks like it's a convention to apply top-k before top-p if both knobs are not `None`. I am not sure whether it's just out of convenience than strong mathematical reasons. My guess is that if we do `top_p` first we won't know for sure how many tokens will be left (as the filter reduce the samples by cumulative probability) for us to prescribe a meaningful `top_k`.

torchmultimodal/utils/generate.py
```diff
@@ -0,0 +1,314 @@
...
+
+    Raises:
+        ValueError: If 'top_k' is outside of valid numerical ranges.
+
+    """
+
+    def __init__(
+        self,
+        top_k: Optional[int] = None,
+        min_tokens_to_keep: int = 1,
```
ebsmothers, line 230: Just an FYI that as written this param is not exposed in the `GenerationUtil` class. Probably not a huge deal, but just wanna point it out in case you think people would want to configure this from the main class

ebsmothers (approved):
Thanks for addressing all the comments, looks good!

torchmultimodal/utils/generate.py
langong347: Thanks. I suspect `min_tokens_to_keep` is not the first order param to be exposed. For example, in Huggingface this param is controlled by `num_beams` (for beam search):

https://github.com/huggingface/transformers/blob/v4.21.2/src/transformers/generation_utils.py#L663

Until then we could probably keep it "hidden" from the user (just like `filter_value` being `-float("inf")`.)

langong347:
@langong347 has imported this pull request.  If you are a Meta employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D38932535).

PR #289, opened 2022-08-27
[MUGEN] Upstream S3D to TorchVision

langong347:
* #300
* __->__ #289

Summary:
- Remapped the weights of `VideoEncoder` from TorchVision's [KINETICS400_V1](https://github.com/pytorch/vision/blob/main/torchvision/models/video/s3d.py#L155) and uploaded to [S3](https://pytorch.s3.amazonaws.com/models/multimodal/mugen/S3D_kinetics400_v1.pth)
- Remove our local implementation of S3D and tests

Test Plan:
- The expected values have changed as a result of using a larger test input because `AvgPool` of `kernel_size = (2, 7, 7)` requires a minimum shape

```
$ python -m pytest examples/mugen/test/
platform darwin -- Python 3.8.13, pytest-7.1.2, pluggy-1.0.0
rootdir: /Users/langong/s3d, configfile: pyproject.toml
plugins: hydra-core-1.1.2, mock-3.8.2, cov-3.0.0
collected 35 items

examples/mugen/test/generation/test_text_video_gpt.py ..................                                         [ 51%]
examples/mugen/test/generation/test_video_vqvae.py ........                                                      [ 74%]
examples/mugen/test/retrieval/test_video_clip.py .........                                                       [100%]

======================== 35 passed in 57.10s ===========
```

Differential Revision: [D39078778](https://our.internmc.facebook.com/intern/diff/D39078778)


langong347:
@langong347 has imported this pull request.  If you are a Facebook employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D39078778).

langong347:
@langong347 has imported this pull request.  If you are a Facebook employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D39078778).

examples/mugen/retrieval/video_clip.py
```diff
@@ -75,26 +78,41 @@ class VideoEncoder(nn.Module):
...
-    def __init__(
-        self,
-    ):
+    def __init__(self, num_classes: int = 400) -> None:
         super().__init__()
-        self.model = S3D(400)
-        self.out_dim = self.model.fc.in_channels
-        self.model.fc = nn.Identity()
+        base = S3D(num_classes=num_classes)
+        self.return_node_name = "features"
```
YosuaMichael, line 96: I think we can have something more similar with the previous implementation:
```
# Getting input channels from the Conv3d layer
self.out_dim = self.model.classifier[1].in_channels

# Transform the classifier into identity
self.model.classifier = nn.Identity()
```

If we do this way, I think we can let the forward as before (no need for additional avgpool and mean)

YosuaMichael (commented):
Hi @langong347 , added a comment on a simpler way to use s3d model (more similar with previous implementation)

examples/mugen/retrieval/video_clip.py
```diff
@@ -8,16 +8,19 @@
...
 from torchmultimodal.utils.common import load_module_from_url
+from torchvision.models.feature_extraction import create_feature_extractor
+
+from torchvision.models.video import S3D
 from transformers import DistilBertConfig, DistilBertModel
 
 
 PRETRAINED_S3D_KINETICS400_URL = (
-    "https://pytorch.s3.amazonaws.com/models/multimodal/mugen/S3D_kinetics400.pt"
+    "https://pytorch.s3.amazonaws.com/models/multimodal/mugen/S3D_kinetics400_v1.pth"
```
YosuaMichael, line 23: I think it might be a good idea to get some hash from the file as part of the name (this is useful for versioning as well). In torchvision we get the hash in [here](https://github.com/pytorch/vision/blob/main/references/classification/utils.py#L387) (note that we only take the first 8 characters of the hash digest).

examples/mugen/retrieval/video_clip.py
```diff
@@ -75,26 +78,41 @@ class VideoEncoder(nn.Module):
...
     Inputs:
         x (Tensor): batch of videos with dimensions (batch, channel, time, height, width)
             Size of ``channel`` dimension must be ``3``.
 
     """
 
-    def __init__(
-        self,
-    ):
+    def __init__(self, num_classes: int = 400) -> None:
```
YosuaMichael, line 93: Since we only use as encoder, I think the `num_classes` here is not really used.
I think we can remove this param and just do: `base = S3D()` instead.

examples/mugen/retrieval/video_clip.py
langong347: Thanks for the suggestion. I was on the fence about whether I should:

1. Put assumptions on the layer structure of `classifier` (that the indexing is always `1`) and the existence of attribute `in_channels`

vs.

2. Treat TorchVision's S3D a blackbox with API `create_feature_extractors`, for example, if for some reason `in_channels` becomes invalid it won't affect my callsite as I've been deriving it by passing actual data through the extracted features:

https://github.com/facebookresearch/multimodal/blob/82beb8f726bc1e19c5489f9cf6e30b580e0608a7/examples/mugen/retrieval/video_clip.py#L98-L102

The downside of 2) is that I have to manually attach `Avg_Pool3d` and `torch.mean` in my `forward` pass. 
On the other hand, if the internals of TorchVision S3D remain stable over time, 1) will be the cleaner solution here.

What do you think?

YosuaMichael: Hi Lan, I think since S3D already become a public API in torchvision, it will be likely quite stable (we will be careful to do any change on it). And I think using the `create_feature_extractors` you already assume to use `model.features` in the model.

I what we need to consider here is where this `VideoEncoder` can be generic (using other encoder, not just S3D) or not. For this, I think its better to have `S3DEncoder` that specific to S3D, and make it standard there (like what we did here), then the `VideoEncoder` can accept any `standard` encoder.

langong347: >I think since S3D already become a public API in torchvision, it will be likely quite stable 
That makes sense to me.

>have S3DEncoder that specific to S3D, and make it standard there (like what we did here), then the VideoEncoder can accept any standard encoder.

What do you mean here by "standard" encoder? So for our `VideoEncoder` to accept any encoder, we will have to request the latter to expose the same attributes (e.g., `classifier`, `in_channels`) as those of TorchVision S3D.

YosuaMichael: Here is a snippet to illustrate my comment:
```
class VideoEncoder(nn.Module):
  def forward(self, x):
    if x.shape[1] != 3:
      ...
    return self.model(x)
    
class S3DEncoder(VideoEncoder):
  def __init__(self):
    ...
```

langong347: Thanks for the code sample! Since we haven't yet needed to generalize, I will keep it specific to using S3D (with the classifier layer replaced) for now.

examples/mugen/retrieval/video_clip.py
langong347: Thanks for the suggestion. I have re-generated the filename using hexdigest. Will probably include TorchVision's util in my next PR as well.

langong347:
@langong347 has imported this pull request.  If you are a Meta employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D39078778).

YosuaMichael (approved):
LGTM :)

PR #290, opened 2022-08-29
[CLIP]Fix gpu tests

ankitade:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* #291
* __->__ #290
* #288

Differential Revision: [D39106446](https://our.internmc.facebook.com/intern/diff/D39106446)

- Fix the gpu test. They were failing because spawn needs global or static functions afaic
- Move test and context manager to pytest

Test plan : 
on gpu 

pytest test/modules/losses/test_contrastive_loss_with_temperature.py -vv
======================================================================== test session starts =========================================================================
platform linux -- Python 3.8.13, pytest-7.1.2, pluggy-1.0.0 -- /data/home/deankita/miniconda/envs/tmm2/bin/python
cachedir: .pytest_cache
rootdir: /data/home/deankita/tmm2, configfile: pyproject.toml
plugins: hydra-core-1.1.2, mock-3.8.2, cov-3.0.0
collected 7 items                                                                                                                                                    

test/modules/losses/test_contrastive_loss_with_temperature.py::TestContrastiveLossWithTemperature::test_local_loss PASSED                                      [ 14%]
test/modules/losses/test_contrastive_loss_with_temperature.py::TestContrastiveLossWithTemperature::test_temperature_clamp_max PASSED                           [ 28%]
test/modules/losses/test_contrastive_loss_with_temperature.py::TestContrastiveLossWithTemperature::test_temperature_clamp_min PASSED                           [ 42%]
test/modules/losses/test_contrastive_loss_with_temperature.py::TestContrastiveLossWithTemperature::test_loss_with_ce_kwargs PASSED                             [ 57%]
test/modules/losses/test_contrastive_loss_with_temperature.py::TestContrastiveLossWithTemperature::test_temperature_clamp_invalid PASSED                       [ 71%]
test/modules/losses/test_contrastive_loss_with_temperature.py::TestContrastiveLossWithTemperature::test_single_gpu_loss PASSED                                 [ 85%]
test/modules/losses/test_contrastive_loss_with_temperature.py::TestContrastiveLossWithTemperature::test_multi_gpu_loss PASSED                                  [100%]

========================================================================= 7 passed in 17.50s =========================================================================

PR #288, opened 2022-08-26
[CLIP]Remove duplicate vit encoder

ankitade:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* __->__ #288

Differential Revision: [D39074106](https://our.internmc.facebook.com/intern/diff/D39074106)
The vit encoder is already part of clip/image_encoder. This one got committed by mistake

Test plan: CI

PR #291, opened 2022-08-29
[CLIP] Fix transform bug related to text truncation

ankitade:
* __->__ #291
* #290
* #288

Differential Revision: [D39106445](https://our.internmc.facebook.com/intern/diff/D39106445)

If text is longer than the context length, it needs to get truncated before adding the eos token else the eos token also gets removed

Test plan
Fixed the expected value in test and ran pytest. Also running E2E training on internal infra


ankitade:
@ankitade has imported this pull request.  If you are a Facebook employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D39106445).

PR #304, opened 2022-09-08
[MUGEN][BugFix] MultimodalGPT layer norm output

langong347:
Stack from [ghstack](https://github.com/ezyang/ghstack):
* #307
* #305
* __->__ #304
* #306
* #303

Summary:
Output of layer norm previously was not passed to the logit projection layer.

Test Plan:
CI

Differential Revision: [D39420774](https://our.internmc.facebook.com/intern/diff/D39420774)

PR #303, opened 2022-09-06
[MUGEN][BugFix] Offset generated tokens by num in_modality token states

langong347:
Stack from [ghstack](https://github.com/ezyang/ghstack):
* #307
* #305
* #304
* #306
* __->__ #303

Summary:
- We should subtract the input token numbers from the result to keep the latter within the range of output token numbers (or codebook size if using VQVAE)
- This is due to the fact that sampling is done in the concatenated token space `(batch, sequence, num_in_tokens + num_out_tokens)`

Test Plan:
```
$ python -m pytest test/utils/test_generate.py -vv

test/utils/test_generate.py::TestLogitsMask::test_normal PASSED                                                  [  6%]
test/utils/test_generate.py::TestLogitsMask::test_zero_dims PASSED                                               [ 13%]
test/utils/test_generate.py::TestLogitsMask::test_in_seq_only PASSED                                             [ 20%]
test/utils/test_generate.py::TestLogitsMask::test_out_seq_only PASSED                                            [ 26%]
test/utils/test_generate.py::TestGenerationUtil::test_model_eval_warning PASSED                                  [ 33%]
test/utils/test_generate.py::TestGenerationUtil::test_sample PASSED                                              [ 40%]
test/utils/test_generate.py::TestGenerationUtil::test_filter_logits PASSED                                       [ 46%]
test/utils/test_generate.py::TestLogitsFilterTopK::test_min_tokens_to_keep PASSED                                [ 53%]
test/utils/test_generate.py::TestLogitsFilterTopK::test_top_k_invalid PASSED                                     [ 60%]
test/utils/test_generate.py::TestLogitsFilterTopK::test_default PASSED                                           [ 66%]
test/utils/test_generate.py::TestLogitsFilterTopK::test_top_k PASSED                                             [ 73%]
test/utils/test_generate.py::TestLogitsFilterTopP::test_min_tokens_to_keep PASSED                                [ 80%]
test/utils/test_generate.py::TestLogitsFilterTopP::test_top_p_invalid PASSED                                     [ 86%]
test/utils/test_generate.py::TestLogitsFilterTopP::test_default PASSED                                           [ 93%]
test/utils/test_generate.py::TestLogitsFilterTopP::test_top_p PASSED                                             [100%]

================================================== 15 passed in 3.48s ==========
```

Differential Revision: [D39266383](https://our.internmc.facebook.com/intern/diff/D39266383)

PR #305, opened 2022-09-08
[MUGEN][Bugfix] Transformer residual layer

langong347:
* #307
* __->__ #305
* #304
* #306
* #303

Summary:
- Fixed a bug that overrode the input to the transformer layer before the latter was added to the attention output
- Updated unit tests

Test Plan:
CI

Differential Revision: [D39420772](https://our.internmc.facebook.com/intern/diff/D39420772)

### DIFF
--- a/torchmultimodal/models/gpt.py
+++ b/torchmultimodal/models/gpt.py
@@ -17,6 +17,18 @@ from torchmultimodal.utils.common import checkpoint_wrapper, get_clones
 
 
 class TransformerDecoderOutput(NamedTuple):
+    """Contains output from ``forward`` of :class:`MultimodalTransformerDecoder`, :class:`TransformerDecoder`.
+
+    Attributes:
+        last_hidden_states (Tensor): Output from the last layer of the transformer.
+        hidden_states (Tuple[Tensor, ...], optional): Outputs from all layers of the transformer.
+            Defaults to ``None``.
+        attention_weights (Tuple[Tensor, ...], optional): Attention probabilities from all layers of the
+            transformer. Defaults to ``None``.
+        past_key_values (Tuple[Dict[str, Tensor], ...]], optional): If ``use_cache`` is on, contains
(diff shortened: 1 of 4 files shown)


langong347:
@langong347 has imported this pull request.  If you are a Meta employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D39420772).

torchmultimodal/models/gpt.py
```diff
@@ -633,16 +633,17 @@ def forward(
...
 
         if use_cache:
             past_key_values = self.attention.cache
 
-        x = x + self.dropout_attn(x)
-        x = self._mlp_block(self.norm_mlp(x))
-        x = x + self.dropout_mlp(x)
+        x = x + self.dropout_attn(hidden)
+
+        hidden = self._mlp_block(self.norm_mlp(x))
```
ebsmothers, line 645: nit: I think the use of `hidden` here is kinda confusing. Maybe we can more closely follow naming in the [existing transformer layer](https://github.com/facebookresearch/multimodal/blob/main/torchmultimodal/modules/layers/transformer.py#L120-L168)?

ebsmothers (approved):

torchmultimodal/models/gpt.py
langong347: Here I am also exposing the option to return the attention probabilities so it's not as straightforward to chain the outputs like in what you quoted (where `return_attn_weights` is always `False`, [ref](https://github.com/facebookresearch/multimodal/blob/main/torchmultimodal/modules/layers/transformer.py#L95)).

I am renaming `hidden` to `attn_hidden_states` and `mlp_hidden_states` to be clear.

langong347:
@langong347 has imported this pull request.  If you are a Meta employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D39420772).

PR #312, opened 2022-09-13
10b changes

rohan-varma:
Summary:

Test plan:

Fixes #{issue number}

### DIFF
--- a/examples/flava/native/train.py
+++ b/examples/flava/native/train.py
@@ -11,6 +11,7 @@
 import time
 from functools import partial
 from typing import Any, Dict, Tuple, Union
+from torchdistx import deferred_init
 
 import datasets
 import torch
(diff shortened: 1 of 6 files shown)


examples/flava/native/train.py
```diff
@@ -156,33 +168,50 @@ def create_model(self) -> torch.nn.Module:
...
+                    TransformerEncoderLayer,
+                    ImageTransformer,
+                    FLAVATransformerWithoutEmbeddings,
+                    DalleVAEEncoder,
+                    DalleEncoderBlock,
+                    BERTTextEncoder,
+                    TransformerEncoder,
+            #        BERTTextEmbeddings,
+                    }
+            wrapper_cls = {
```
langong347, line 186: Is one of those `wrapper_cls` redundant?

examples/flava/native/train.py
```diff
@@ -156,33 +168,50 @@ def create_model(self) -> torch.nn.Module:
...
+                    ImageTransformer,
+                    BERTTextEncoder,
+                    DalleVAEEncoder,
+                    DalleEncoderBlock,
+            }
+            #wrapper_cls = {TransformerEncoderLayer}
+            print(f"mp config {mp}")
 
             if self.config.training.activation_checkpointing:
                 # note: activation checkpointing wrapper currently is faulty
```
langong347, line 197: N00b Q: since you put these comments here, would you mind explaining:
- As FSDP by itself distributes a large model across GPUs, is it still necessary to use activation checkpointing?
- When would you want to use `re-entrant` vs. `non re-entrant`?
- Any idea why memory reduction from checkpointing is less than expected?

Thanks!

langong347 (commented):
Left a few questions. Also, would be good to add a summary and a test plan so people like me who do not have much context here would understand. :)

rohan-varma:
@langong347 Sorry for late response, I didn't notice the comments here. Most of these changes have been merged by @edward-io. Some quick responses to your questions, feel free to ping me for more details: 

1. Yes, AC is necessary, as FSDP does not itself reduce memory consumption due to activations
2. Non-reentrant is a new implementation approach in PyTorch core, that supports several additional use cases, and the thought is that it would be more composable with FSDP. 
3. Somehow, non-reentrant version of checkpoint doesn't save as much as reentrant. We are not sure why, but have some follow up work to sync with the core team and understand what is going on here.

PR #328, opened 2022-09-22
[Omnivore] Speedup omnivore test

YosuaMichael:
Summary:
Change the fixture on the models for omnivore to use `autouse=False` in order to speedup the test since the models are not always used on all test.

Also set the input into explicitly to correct device.

### DIFF
--- a/test/models/test_omnivore.py
+++ b/test/models/test_omnivore.py
@@ -18,17 +18,17 @@ def device():
     return get_current_device()
 
 
-@pytest.fixture(autouse=False)
+@pytest.fixture()
 def omnivore_swin_t_model(device):
     return omnivore.omnivore_swin_t().to(device)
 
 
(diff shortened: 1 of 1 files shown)


langong347 (approved):
Thanks for the fix! 
- Nit: setting autouse explicitly to be `False` -- I suppose this is out of precaution rather than necessity?
- I retriggered the lint job as it failed due to some network connection error. Please verify when you are back.

YosuaMichael:
> Thanks for the fix!
> 
> * Nit: setting autouse explicitly to be `False` -- I suppose this is out of precaution rather than necessity?
> * I retriggered the lint job as it failed due to some network connection error. Please verify when you are back.

Thanks @langong347 , yeah I put `autouse=False` just to be safe and you are correct they are default value so I removed it now.

PR #143, opened 2022-07-10
[Unification] Refactor all architectures

RdoubleA:
Summary:
Move all models in `architectures/` to `models/` and refactor accordingly. Remove "Architecture" from any model name.

Test plan:
`pytest test -vv`

PR #332, opened 2022-09-26
[Packaging] Remove pyproject config that modifies pythonpath during testing

langong347:
* (to be filled)

Summary:

- Remove config in `pyproject.toml` that defaults pythonpath to the source code directory during test execution
- We should not maniuplate `pythonpath` implicitly anywhere in the codebase as it could lead to many pitfalls down the road.
- Specifically, it makes setting pytest `import-mode=append` futile and consequently the tests won't run against the installed package.

Test Plan:

CI

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,7 +1,2 @@
 [tool.usort]
 first_party_detection = false
-
-[tool.pytest.ini_options]
-pythonpath = [
-  "."
-]


pikapecan (approved):

ankitade:
do u remember why we added this in the first place?

ankitade (approved):

langong347:
@ankitade  It was added in #143 on an unrelated topic, probably to enforce the prepending of `multimodal/` to `sys.path`. This makes sense for development but becomes undesirable for testing against the installation. See [pytest configuration for pythonpath](https://docs.pytest.org/en/7.1.x/reference/reference.html#confval-pythonpath).

PR #339, opened 2022-09-28
[FLAVA]Fix activation ckpting api name

ankitade:
Summary:
1. with recent nightly, the name of the api for ac  changed, fixing it
2. Adding mm encoder also to the wrapping module. since its more intuitive for user to wrap all encoders separately and also doesnt seem to impact performance.

Test plan:
torchrun --master_port=1256 --nproc_per_node=8 -m flava.native.train config=flava/native/configs/pretrain_debug.yaml training.max_steps=1000 training.activation_checkpointing=True training.strategy=fsdp


edward-io (approved):
LGTM

PR #346, opened 2022-10-03
Add limit_all_gathers=True

rohan-varma:
Adding rate limiter support. Per @ankitade @edward-io, this can provide up to 2x speedup for 10B parameter model.


ankitade (approved):

ankitade:
can u import and land there

PR #361, opened 2022-10-26
Refactor contrastive loss to make tensor gathering a util

abhinavarora:
Summary:
This diff refactors Contrastive Loss to make tensor gathering a util so that it can be used with other losses.

This diff also fixes an existing bug in the previous logic of gather tensors, that prevents backpropagating to the local rank when `backprop_in_gather` is set to `False`

Differential Revision: D40598601

### DIFF
--- a/tests/utils/test_distributed.py
+++ b/tests/utils/test_distributed.py
@@ -1,0 +1,89 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+import pytest
+import torch
+import torch.multiprocessing as mp
+from tests.test_utils import (
+    assert_expected,
+    gpu_test,
+    init_distributed_on_file,
+    set_rng_seed,
+    with_temp_files,
+)
+from torch import Tensor
+from torchmultimodal.utils.distributed import gather_tensor
+
+BATCH_SIZE = 4
+EMBEDDING_DIM = 8
+
+
+class TestGatherTensor:
+    """
+    Test gather_tensor method with backprop_in_gather param
+    """
+
+    @pytest.fixture(autouse=True)
+    def setup(self):
+        set_rng_seed(1234)
+
+    @pytest.fixture
+    def input_tensor(self):
+        return torch.randn(BATCH_SIZE, EMBEDDING_DIM)
+
+    @staticmethod
+    def _worker(
+        gpu_id: int,
+        world_size: int,
+        sync_file: str,
+        input_tensor: Tensor,
+        backprop_in_gather: bool,
+    ):
+        init_distributed_on_file(
+            world_size=world_size, gpu_id=gpu_id, sync_file=sync_file
+        )
+        gpu_tensor = input_tensor.clone().requires_grad_().to(gpu_id) + gpu_id
+        expected_output = [input_tensor + i for i in range(world_size)]
+        gathered_output = gather_tensor(gpu_tensor, backprop_in_gather)
+        assert_expected(len(expected_output), len(gathered_output))
+        for i, tensor in enumerate(gathered_output):
+            assert_expected(tensor, expected_output[i], check_device=False)
+            if i == gpu_id or backprop_in_gather:
+                assert tensor.grad_fn is not None
+            else:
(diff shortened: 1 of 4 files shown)


tests/utils/test_distributed.py
```diff
@@ -0,0 +1,89 @@
...
+class TestGatherTensor:
+    """
+    Test gather_tensor method with backprop_in_gather param
+    """
+
+    @pytest.fixture(autouse=True)
+    def setup(self):
+        set_rng_seed(1234)
+
+    @pytest.fixture()
```
langong347, line 33: The parentheses are not required when there're no arguments: https://docs.pytest.org/en/6.2.x/fixture.html#back-to-fixtures

tests/utils/test_distributed.py
```diff
@@ -0,0 +1,89 @@
...
+    ):
+        init_distributed_on_file(
+            world_size=world_size, gpu_id=gpu_id, sync_file=sync_file
+        )
+        gpu_tensor = input_tensor.clone().requires_grad_().to(gpu_id) + gpu_id
+        expected_output = [input_tensor + i for i in range(world_size)]
+        gathered_output = gather_tensor(gpu_tensor, backprop_in_gather)
+        assert_expected(len(expected_output), len(gathered_output))
+        for i, tensor in enumerate(gathered_output):
+            assert_expected(tensor, expected_output[i], check_device=False)
```
langong347, line 53: So `expected_output` and `gathered_output` are not on the same device and that's why we need to turn off `check_device`. What if, instead, we deliberately move either one onto the device of the other before comparison? Separately, it's ok to expose `check_device` to `assert_expected` to allow more control but I am not sure if here the use case is necessary.

abhinavarora: I think that is a fair point. Personally, I do not like the current design of the `assert_expected` method and instead we should refactor it at some point. The `assert_expected` is a lightweight wrapper around `torch.testing.assert_close` that just specifies a `msg` argument. Hence in principal, we should expose all functionalities of `torch.testing.assert_close` to developers. Therefore, I thought that it is better to add `check_device` for now instead of unnecessarily moving tensors till we figure out a better design for `assert_expected`. The use case for `check_device` is just because we only want to check the content of the tensors regardless of the location. I don't think there is a reason to unnecessarily move tensors when `torch.testing.assert_close` provides the functionality to ignore the device. 

What do you think?

tests/utils/test_distributed.py
abhinavarora: Fixed in the latest revision. Thank you!

torchmultimodal/utils/distributed.py
```diff
@@ -0,0 +1,40 @@
...
+from torch.distributed.nn.functional import all_gather as all_gather_with_backprop
+
+
+def gather_tensor(tensor: Tensor, backprop_in_gather: bool = True) -> List[Tensor]:
+    """Gathers a tensor across all GPUs.
+
+    Args:
+        tensor (Tensor): Tensors that need to be gathered.
+        backprop_in_gather (bool): Whether to backpropagate the gradients from
+            all_gather to all workers (versus just the local worker).
```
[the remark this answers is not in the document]
abhinavarora: Fixed in the latest revision. Thank you!

langong347:
I think this PR is good to go. cc @ebsmothers  @ankitade any further thoughts?
@abhinavarora please rerun the CI as the upstream nightlies were broken yesterday.

ebsmothers (approved):

torchmultimodal/utils/distributed.py
```diff
@@ -0,0 +1,41 @@
...
+    if backprop_in_gather:
+        return all_gather_with_backprop(tensor)
+
+    # Otherwise just backprop to the current worker
+    # This means that the image gradients on a given worker will only
+    # consider the text samples from the same worker
+    else:
+        tensor_all_gpus = [torch.zeros_like(tensor) for _ in range(world_size)]
+        all_gather_no_backprop(tensor_all_gpus, tensor)
+        tensor_all_gpus[torch.distributed.get_rank()] = tensor
```
ankitade, line 40: there might be a case u dont want to do this in the general util, say gathering for metrics. but fine for now

ankitade (approved):
thanks for bug fix

PR #362, opened 2022-10-26
Upgrade HF Datasets Library for Flava Example

pbontrager:
Summary:
Updated Hugging Face datasets library version in examples/flava. The older version can no longer interface with the HF online API and therefore cannot pull requested datasets.

Test plan:
I created a new conda env and built the dependencies using the new requirement. From there I ran the flava train.py script where it successfully pulled the Imagenet dataset as it's supposed to do and successfully started training.

### DIFF
--- a/examples/flava/requirements.txt
+++ b/examples/flava/requirements.txt
@@ -1,6 +1,6 @@
 Pillow==9.0.1
 pytorch-lightning==1.6.0
-datasets==2.0.0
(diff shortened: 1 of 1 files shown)


ankitade (approved):

apsdehal (approved):
LGTM but looks like tests are breaking unless they were failing before as well.

PR #364, opened 2022-11-04
Change the expected values to match the exact value from the new exponential implementation of PyTorch

CaoE:
Summary:
https://github.com/pytorch/pytorch/pull/69967 adds a MKL implementation for exponential  in PyTorch. Even the new implementation fits the exponential distribution but it is not the same as the old one. 
Change the expected values to accommodate the new exponential implementation of PyTorch.


ebsmothers (changes requested):
@CaoE thanks for the PR. Can you just change the expected value to match the exact value from the new exponential implementation (instead of changing the tolerance, which is likely to allow more false positives)?

CaoE:
@ebsmothers Yes, we can just change the expected value, but will it fail the pre-ci of facebookresearch ? or other PRs for pytorchs can't pass the internal facebookresearch test cases before https://github.com/pytorch/pytorch/pull/69967 is merged.

ebsmothers:
@CaoE it can be done as two separate PRs: one to change the value and disable the test until pytorch/pytorch#69667 lands, then another to re-enable the test afterwards. See e.g. #349 and #359

CaoE:
@ebsmothers Thanks for your suggestions. Changed the expected values and disabled the tests.

ebsmothers (commented):
Thanks, this looks good. Can you run the pre-commit hooks to fix linter complaints? Besides that this should be good to go

CaoE:
Fix the linter complaints.

ebsmothers:
Thanks, this looks good. CI is still complaining but it should be due to some upstream breakage and not your changes. To be safe we can wait until tomorrow (once new nightlies are out) to re-run CI and merge after that

ebsmothers (approved):

PR #367, opened 2022-11-09
[MDETR] Fix phrase grounding example and add README

ebsmothers:
Summary:

A few changes to get phrase grounding eval for MDETR up and running again.

1) Move to relative imports inside MDETR example so that phrase_grounding.py script is runnable without any hacks to sys.path.
2) Set torch.use_deterministic_algorithms to warn_only mode (due to upstream changes in cumsum)
3) Fix issue with batch_size = 1 by manually reshaping mask inside text encoder

Finally add a README so users can run more easily.

Test plan:
See the repro in examples/mdetr/README.md. Gives the same results as the test plan in #231.

Also unit tests

### DIFF
--- a/torchmultimodal/models/mdetr/text_encoder.py
+++ b/torchmultimodal/models/mdetr/text_encoder.py
@@ -69,6 +69,8 @@ class ModifiedTransformerEncoder(nn.Module):
     ) -> TransformerOutput:
         encoded = embeddings
         mask = torch.squeeze(attention_mask.to(dtype=torch.bool))
+        if mask.dim() == 1:
+            mask = mask.unsqueeze(0)
         # Do this in a loop because otherwise it can cause OOM
         for layer in self.layers.layers:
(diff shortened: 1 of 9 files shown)


torchmultimodal/models/mdetr/text_encoder.py
```diff
@@ -69,6 +69,8 @@ def forward(
     ) -> TransformerOutput:
         encoded = embeddings
         mask = torch.squeeze(attention_mask.to(dtype=torch.bool))
+        if mask.dim() == 1:
```
ankitade, line 72: why is dim 1 for bsz 1?

ebsmothers: Because of the unsqueeze operation just above. It will remove all dimensions of size 1, so in this case batch size dim also gets removed, which we don't want.

ankitade (approved):

PR #379, opened 2022-11-23
[omnivore]Fix imports in the train script

ankitade:
Summary:
use omnivore directory as the root for imports

Test plan:
torchrun --nproc_per_node=8 --nnodes=1 -m omnivore.train     --batch-size=128 --workers=6 --model="omnivore_swin_t"     --cache-video-dataset     --extra-video-dataloader-workers=7     --kinetics-dataset-workers=12     --val-resize-size=224     --video-grad-accum-iter=32     --modalities image video rgbd     --val-data-sampling-factor 1 1 1     --test-only --pretrained --imagenet-data-path=/datasets01_ontap/imagenet_full_size/061417 --kinetics-data-path=/datasets01_ontap/kinetics/070618/400 --sunrgbd-data-path=/data/home/yosuamichael/datasets/SUN_RGBD

[2022-11-23 04:12:32,122] INFO - Test:  rgbd Acc@1 61.964 rgbd Acc@5 87.822
[2022-11-23 04:12:32,122] INFO - Test:  Video Acc@1 77.985 Video Acc@5 93.462
[2022-11-23 04:12:32,122] INFO - Test:  rgbd Acc@1 61.964 rgbd Acc@5 87.822

PR #382, opened 2022-11-24
[omnivore] Using Upstreamed SwinTransformer3d in TorchVision

YosuaMichael:
Summary:

- Use the upstreamed SwinTransformer3d implementation in TorchVision
- Fix the test for SwinTransformer3d (update the value because the random init changed)
- Remove component test for SwinTransformer3d since it is upstreamed from torchvision no need for this anymore
- Also fix the notebook import

Test plan:
- Run all cells in the notebook: https://github.com/facebookresearch/multimodal/blob/main/examples/omnivore/omnivore_inference_demo.ipynb
- See the CI tests run, or run manually:
    - `pytest models/test_omnivore.py`
    - `pytest modules/encoders/test_swin_transformer_3d_encoder.py`

### DIFF
--- a/examples/omnivore/omnivore_inference_demo.ipynb
+++ b/examples/omnivore/omnivore_inference_demo.ipynb
@@ -29,6 +29,7 @@
     "import torchvision.transforms as T\n",
     "\n",
     "import sys\n",
+    "# We add the folder: \"<torchmultimodal_repo>/examples\" to path to import the presets\n",
     "sys.path.append(os.path.dirname(os.path.abspath(\"\")))\n",
     "\n",
     "import torchmultimodal.models.omnivore as omnivore\n",
--- a/examples/omnivore/omnivore_inference_demo.ipynb
+++ b/examples/omnivore/omnivore_inference_demo.ipynb
@@ -28,8 +28,11 @@
     "from torchvision.io import read_video, _read_video_from_file\n",
     "import torchvision.transforms as T\n",
     "\n",
+    "import sys\n",
+    "sys.path.append(os.path.dirname(os.path.abspath(\"\")))\n",
+    "\n",
     "import torchmultimodal.models.omnivore as omnivore\n",
-    "from examples.omnivore.data import presets\n",
+    "from omnivore.data import presets\n",
     "\n",
     "from IPython.display import Video\n",
     "from matplotlib import pyplot as plt"
@@ -338,7 +341,7 @@
    "name": "python",
    "nbconvert_exporter": "python",
    "pygments_lexer": "ipython3",
-   "version": "3.9.10"
+   "version": "3.10.6"
   }
  },
  "nbformat": 4,
--- a/tests/models/test_omnivore.py
+++ b/tests/models/test_omnivore.py
@@ -41,19 +41,19 @@ def test_omnivore_swin_t_forward(omnivore_swin_t_model, device):
 
     assert_expected(image_score.size(), torch.Size((1, 1000)))
     assert_expected(
-        image_score.abs().sum(), torch.tensor(200.27574), rtol=1e-3, atol=1e-3
+        image_score.abs().sum(), torch.tensor(184.01417), rtol=1e-3, atol=1e-3
     )
 
     rgbd = torch.randn((1, 4, 1, 112, 112), device=device)
     rgbd_score = model(rgbd, input_type="rgbd")
     assert_expected(rgbd_score.size(), torch.Size((1, 19)))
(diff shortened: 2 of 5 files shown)


tests/modules/encoders/test_swin_transformer_3d_encoder.py
```diff
@@ -41,41 +39,7 @@ def test_swin_transformer_3d_encoder(self):
...
         scores = self.encoder(image)
         self.assertEqual(scores.size(), torch.Size([1, 768]))
-        self.assertAlmostEqual(scores.abs().sum().item(), 277.63833, 2)
+        self.assertAlmostEqual(scores.abs().sum().item(), 247.14674, 2)
 
     def test_swin_transformer_3d_scripting(self):
         torch.jit.script(self.encoder)
-
-
-class TestSwinTransformer3dComponents(unittest.TestCase):
```
YosuaMichael, line 50: Since we have upstreamed the model to TorchVision and import from there, there is no point in testing the components here.

examples/omnivore/omnivore_inference_demo.ipynb
```diff
@@ -28,8 +28,12 @@
     "from torchvision.io import read_video, _read_video_from_file\n",
     "import torchvision.transforms as T\n",
     "\n",
+    "import sys\n",
+    "# We add the folder: \"<torchmultimodal_repo>/examples\" to path to import the presets\n",
+    "sys.path.append(os.path.dirname(os.path.abspath(\"\")))\n",
+    "\n",
     "import torchmultimodal.models.omnivore as omnivore\n",
-    "from examples.omnivore.data import presets\n",
+    "from omnivore.data import presets\n",
```
YosuaMichael, line 36: Follow the import path in https://github.com/facebookresearch/multimodal/pull/379

To do this, we need to add the path to `<torchmultimodal_repo>/examples` to the path.

tests/models/test_omnivore.py
```diff
@@ -41,19 +41,19 @@ def test_omnivore_swin_t_forward(omnivore_swin_t_model, device):
 
     assert_expected(image_score.size(), torch.Size((1, 1000)))
     assert_expected(
-        image_score.abs().sum(), torch.tensor(200.27574), rtol=1e-3, atol=1e-3
+        image_score.abs().sum(), torch.tensor(184.01417), rtol=1e-3, atol=1e-3
```
YosuaMichael, line 44: Since the underlying SwinTransformer3d implementation changed (probably changed on some ordering or other thing but the logic stay the same), the random initiation of the weight also change, hence we need to update the expected value here.

YosuaMichael (commented):
Add few comment for further context:

torchmultimodal/modules/encoders/swin_transformer_3d_encoder.py
```diff
@@ -8,338 +8,25 @@
...
-from torch import nn, Tensor
-from torchvision.models.swin_transformer import PatchMerging, SwinTransformerBlock
+from torch import nn
+from torchvision.models.video.swin_transformer import (
+    PatchEmbed3d as TVPatchEmbed3d,
+    PatchMerging as TVPatchMerging,
+    SwinTransformer3d as TVSwinTransformer3d,
+)
 
+# We redefine this because they are imported on other script
```
ankitade, line 20: i dont understand this part, which script?

ankitade: yeah i am fine with importing from TV

ankitade (approved):
any chance u can move to composition vs inheritance for swin?

ankitade (changes requested):
tests failing

YosuaMichael:
@ankitade 

> any chance u can move to composition vs inheritance for swin?
I think the main cons for doing composition here is that the layout of the model will change and hence we need to change the pretrained weight as well. This might not be a good idea because it is not BC.

> tests failing
I have take a look on this, and seems like conda install nightly install the wrong version of torchvision (it install version 0.13.1 where it should be 0.15 for nightly). I have followed up and indeed there is a failure on conda upload recently that cause this issue. So this failure is not really related to the changes, but to be safe lets wait until it resolved.

YosuaMichael:
I create this issue on torchvision to keep track the root cause of the test failure: https://github.com/pytorch/vision/issues/7022

YosuaMichael:
@ankitade The issue regarding the torchvision version on conda has been resolved and it fix the test as well

ankitade (approved):

Issue #384, opened 2022-11-30
Use CLIP models with pretrained weights


konradkalita:
## Issue description

Hi, I wanted to ask if it is possible to load [openai/clip-vit-base-patch16](https://huggingface.co/openai/clip-vit-base-patch16) weights to `torchmultimodal.models.clip.model.CLIP` model provided by the library.

PR #397, opened 2023-01-23
Added OpenAI Pre-trained weights to CLIP Models

pbontrager:
Summary:
Added pre-trained flag to all CLIP models. This will pull pre-trained weights for any model there are open weights for, otherwise it'll provide a warning and ignore the flag. The flag pulls the OpenAI weights from S3 which are saved as state_dict and loads those weights. All of the weights have been tested to output the same outputs as their corresponding models in the OpenAI CLIP repo.

CLIP models that were configured differently from the OpenAI repo were updated to match; this was the ViT-l/14 model and all the ResNet models. Additionally, a parameter was added to the CLIP text encoder to be able to be configured to match the OpenAI configs.

Test plan:
I added new unit tests that call each CLIP model with a random image and tokens and check the first 5 features of the text and image embeddings against an expected output. Running `pytest tests/models/clip/test_clip.py -vv` will run the tests. I removed the cached models on the server before running the test to verify the pre-trained weights were correctly pulled from S3 during the test.
Example notebook to verify weights match: https://colab.research.google.com/drive/1ANLQ4tBP2ejuLWZNprLl_Z8iWqdBG_CX?usp=sharing

Fixes https://github.com/facebookresearch/multimodal/issues/384

### DIFF
--- a/torchmultimodal/models/clip/model.py
+++ b/torchmultimodal/models/clip/model.py
@@ -13,12 +13,25 @@ from torch import nn
 
 from torchmultimodal.models.clip.image_encoder import CLIPViTEncoder, ResNetForCLIP
 from torchmultimodal.models.clip.text_encoder import CLIPTextEncoder
+from torchmultimodal.utils.common import load_module_from_url
 from torchvision.models.resnet import Bottleneck, ResNet
 
 
 class CLIPOutput(NamedTuple):
     embeddings_a: torch.Tensor
     embeddings_b: torch.Tensor
+
+
+CLIP_MODEL_MAPPING = {
+    "vit_b16": "https://download.pytorch.org/models/multimodal/clip/clip_vit_b16.pt",
+    "vit_b32": "https://download.pytorch.org/models/multimodal/clip/clip_vit_b32.pt",
+    "vit_l14": "https://download.pytorch.org/models/multimodal/clip/clip_vit_l14.pt",
+    "rn50": "https://download.pytorch.org/models/multimodal/clip/clip_rn50.pt",
+    "rn101": "https://download.pytorch.org/models/multimodal/clip/clip_rn101.pt",
+    "rn50x4": "https://download.pytorch.org/models/multimodal/clip/clip_rn50x4.pt",
+    "rn50x16": "https://download.pytorch.org/models/multimodal/clip/clip_rn50x16.pt",
+    "rn50x64": "https://download.pytorch.org/models/multimodal/clip/clip_rn50x64.pt",
+}
 
 
 class CLIP(nn.Module):
@@ -63,23 +76,29 @@ class CLIP(nn.Module):
         return CLIPOutput(embeddings_a=embeddings_a, embeddings_b=embeddings_b)
 
 
-def clip_vit_b16() -> CLIP:
+def clip_vit_b16(pretrained: bool = False) -> CLIP:
     vision_encoder = CLIPViTEncoder(
         image_size=224, patch_size=16, layers=12, heads=12, width=768, embedding_dim=512
     )
     text_encoder = CLIPTextEncoder(embedding_dim=512)
-    return CLIP(vision_encoder, text_encoder)
+    clip = CLIP(vision_encoder, text_encoder)
+    if pretrained:
+        load_module_from_url(clip, CLIP_MODEL_MAPPING["vit_b16"])
+    return clip
 
 
-def clip_vit_b32() -> CLIP:
+def clip_vit_b32(pretrained: bool = False) -> CLIP:
     vision_encoder = CLIPViTEncoder(
         image_size=224, patch_size=32, layers=12, heads=12, width=768, embedding_dim=512
     )
     text_encoder = CLIPTextEncoder(embedding_dim=512)
-    return CLIP(vision_encoder, text_encoder)
+    clip = CLIP(vision_encoder, text_encoder)
+    if pretrained:
+        load_module_from_url(clip, CLIP_MODEL_MAPPING["vit_b32"])
+    return clip
 
 
-def clip_vit_l14() -> CLIP:
+def clip_vit_l14(pretrained: bool = False) -> CLIP:
     vision_encoder = CLIPViTEncoder(
         image_size=224,
         patch_size=14,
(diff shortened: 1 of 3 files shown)


torchmultimodal/models/clip/text_encoder.py
```diff
@@ -24,6 +24,7 @@ class CLIPTextEncoder(nn.Module):
         context_length (int): Maximum sequence length for Transforer.
         vocab_size (int): Vocab size.
         width (int): Embedding dimension for Transformer encoder.
+        dim_feedforward (int): Dimmension of the feedfoward networks.
```
RdoubleA, line 27:
```suggestion
        dim_feedforward (int): Dimension of the feedfoward networks.
```

RdoubleA (approved):
Overall looks good to me. Are there any references for the updated ResNet parameters? Or for any the CLIP variations. Would be great to add some info on where they're from/where they're used in a docstring

pbontrager:
> Overall looks good to me. Are there any references for the updated ResNet parameters? Or for any the CLIP variations. Would be great to add some info on where they're from/where they're used in a docstring

The updated parameters are reverse engineered from the saved checkpoints from the CLIP library. I'm not sure where the original values came from.

tests/models/clip/test_clip.py
```diff
@@ -120,3 +131,201 @@ def test_clip_vit_forward(self, context_length):
...
+        )
+
+    def test_clip_vit_l14(self, data):
+        text, image, *_ = data
+        model = pretrained.clip_vit_l14(True)
+        model.eval()
+        with torch.no_grad():
+            actual_a_embedding, actual_b_embedding = model(image, text)
+
+        expected_a_embedding = torch.tensor([-0.0066, 0.0428, 0.0369, 0.0357, -0.0001])
```
ankitade, line 190: why not just check for mean

tests/models/clip/test_clip.py
```diff
@@ -120,3 +131,201 @@ def test_clip_vit_forward(self, context_length):
...
+        expected_b_embedding = torch.tensor(
+            [-0.0415, -0.0375, -0.0474, 0.0057, -0.0353]
+        )
+
+        assert_expected(
```
ankitade, line 195: assert on size

torchmultimodal/models/clip/model.py
```diff
@@ -88,85 +108,121 @@ def clip_vit_l14() -> CLIP:
...
 
 
-def clip_rn50() -> CLIP:
+def clip_rn50(pretrained: bool = False) -> CLIP:
     vision_encoder = ResNetForCLIP(
         layers=(3, 4, 6, 3),
         output_dim=1024,
-        heads=1024,
-        width=2048,
+        heads=32,
```
ankitade, line 124: idk how were these values so off to begin with

torchmultimodal/models/clip/model.py
```diff
@@ -88,85 +108,121 @@ def clip_vit_l14() -> CLIP:
...
-    return CLIP(vision_encoder, text_encoder)
+    clip = CLIP(vision_encoder, text_encoder)
+    if pretrained:
+        load_module_from_url(clip, CLIP_MODEL_MAPPING["rn50x64"])
+    return clip
 
 
 # Note: these models use torchvision's ResNet
-def clip_rn50_tv() -> CLIP:
+def clip_rn50_tv(pretrained: bool = False) -> CLIP:
```
ankitade, line 201: remove the pretrained parameter to all tv ones, its confusing

torchmultimodal/models/clip/text_encoder.py
```diff
@@ -24,6 +24,7 @@ class CLIPTextEncoder(nn.Module):
         context_length (int): Maximum sequence length for Transforer.
         vocab_size (int): Vocab size.
         width (int): Embedding dimension for Transformer encoder.
+        dim_feedforward (int): Dimension of the feedfoward networks.
```
ankitade, line 27: did u check adding this isnt breaking anything internal in terms of args order

tests/models/clip/test_clip.py
```diff
@@ -120,3 +131,201 @@ def test_clip_vit_forward(self, context_length):
...
+            actual=actual_b_embedding[0, :5],
+            expected=expected_b_embedding,
+            rtol=0,
+            atol=1e-4,
+        )
+
+    def test_clip_rn50x64(self, data):
+        text, *_, image = data
```
ankitade, line 310: nit: can u call it same as the data name i.e. image_448 for easy reading

ankitade (commented):
can u add a notebook showing equivalence of fwd pass with open ai clip models.

tests/models/clip/test_clip.py
pbontrager: Not sure which one is more unique, but that probably would have been a bit more compact.

tests/models/clip/test_clip.py
pbontrager: Of the embedding or of all the parameters? If it's of the embedding, it seems unlikely the embedding size could change without changing the values.

torchmultimodal/models/clip/model.py
pbontrager: I was trying to keep the "api" consistent for all the model functions. If someone had some kind of model factory code or config I could also see it being annoying that only some models have the pretrained flag.

torchmultimodal/models/clip/text_encoder.py
pbontrager: I did a search on the codebase for CLIPTextEncoder and didn't find anywhere referencing the module that used positional arguments.

tests/models/clip/test_clip.py
ankitade: > unlikely the embedding size could change without changing the values.

don't quite follow. if you are going to assert on the mean value, its good to add a check on the shape of the output embedding (bsz x projection dim)?

torchmultimodal/models/clip/model.py
ankitade: in this case, I would say, functionality is more important than consistency. why would we expose an arg that does nothing (and gives an impression its doing something if the user misses the warning log)

ankitade (approved):

PR #406, opened 2023-02-13
[FLAVA] Fix for DDP with active checkpointing

### LINKED ISSUE #404: training flava with ddp and activation checkpointing gives runtime error
Hi, I'm following the [tutorial](https://pytorch.org/blog/scaling-multimodal-foundation-models-in-torchmultimodal-with-pytorch-distributed/) here and trying to train flava with torchmm. However, training with ddp and activation checkpointing gives the following error. 

`
  File "/data/root/anaconda3/envs/torchdata/lib/python3.8/site-packages/torch/utils/checkpoint.py", line 157, in backward
    torch.autograd.backward(outputs_with_grad, args_with_grad)
  File "/data/root/anaconda3/envs/torchdata/lib/python3.8/site-packages/torch/autograd/__init__.py", line 197, in backward
    Variable._execution_engine.run_backward(  # Calls into the C++ engine to run the backward pass
RuntimeError: Expected to mark a variable ready only once. This error is caused by one of the following reasons: 1) Use of a module parameter outside the `forward` function. Please make sure model parameters are not shared across multiple concurrent forward-backward passes. or try to use _set_static_graph() as a workaround if this module graph does not change during training loop.2) Reused parameters in multiple reentrant backward passes. For example, if you use multiple `checkpoint` functions to wrap the same part of your model, it would result in the same set of parameters been used by different reentrant backward passes multiple times, and hence marking a variable ready multiple times. DDP does not support such use cases in default. You can try to use _set_static_graph() as a workaround if your module graph does not change over iterations.
Parameter at index 493 has been marked as ready twice. This means that multiple autograd engine  hooks have fired for this particular parameter during this iteration. You can set the environment variable TORCH_DISTRIBUTED_DEBUG to either INFO or DETAIL to print parameter names for further debugging.
`
I also tried some other configurations as below, only ddp + checkpoint gives the above error. I'm wondering if there is some other change we need to make to the code. 

| launch | activation checkpoint | result |
------------ | -------------| -------------
ddp | &check; | &cross;
ddp | &cross; | &check;
fsdp | &check; | &check;

My Env: pytorch=1.13.1, cuda 11.7, torchmultimodal(via pip, from source) and I launched my script with
`torchrun --rdzv_endpoint="localhost:1234" --nproc_per_node=4 -m flava.native.train config=flava/native/configs/pretrain_debug.yaml`

Many thanks in advance!

enmalik:
Summary:
Fix for using DDP with active checkpointing. Have to change to NO_REENTRANT.

Test plan:
`NUM_GPUS=8; torchrun --nproc_per_node=$NUM_GPUS -m flava.native.train config=flava/native/configs/pretrain_debug.yaml `

Confirmed to work with DDP + AC and FSDP + AC.

Fixes #404

### DIFF
--- a/examples/flava/native/train.py
+++ b/examples/flava/native/train.py
@@ -146,15 +146,19 @@ class Trainer:
 
         if self.config.training.activation_checkpointing:
             check_fn = lambda submodule: isinstance(submodule, TransformerEncoderLayer)
+            checkpoint_impl = CheckpointImpl.REENTRANT
 
-            non_reentrant_wrapper = partial(
+            if strategy == "ddp":
+                checkpoint_impl = CheckpointImpl.NO_REENTRANT
+
+            checkpoint_wrapper_fn = partial(
                 checkpoint_wrapper,
                 offload_to_cpu=False,
-                checkpoint_impl=CheckpointImpl.REENTRANT,
+                checkpoint_impl=checkpoint_impl,
             )
             apply_activation_checkpointing(
                 model,
-                checkpoint_wrapper_fn=non_reentrant_wrapper,
+                checkpoint_wrapper_fn=checkpoint_wrapper_fn,
                 check_fn=check_fn,
             )
 
--- a/examples/flava/native/train.py
+++ b/examples/flava/native/train.py
@@ -148,6 +148,7 @@ class Trainer:
             check_fn = lambda submodule: isinstance(submodule, TransformerEncoderLayer)
             checkpoint_impl = CheckpointImpl.REENTRANT
 
+            # DDP gradient hooks have compatibility issues with REENTRANT autograd
             if strategy == "ddp":
                 checkpoint_impl = CheckpointImpl.NO_REENTRANT
 


examples/flava/native/train.py
```diff
@@ -146,15 +146,19 @@ def create_model(self) -> torch.nn.Module:
 
         if self.config.training.activation_checkpointing:
             check_fn = lambda submodule: isinstance(submodule, TransformerEncoderLayer)
+            checkpoint_impl = CheckpointImpl.REENTRANT
 
-            non_reentrant_wrapper = partial(
+            if strategy == "ddp":
```
ebsmothers, line 151: Add a comment here explaining why we use NO_REENTRANT with DDP

ebsmothers (commented):
Overall looks good modulo one minor comment. Re the test plan, did you confirm that DDP + AC works after your change?

enmalik:
> Overall looks good modulo one minor comment. Re the test plan, did you confirm that DDP + AC works after your change?

Yes, works with DDP + AC as well as FSDP + AC.

examples/flava/native/train.py
enmalik: Done.

ebsmothers (approved):

PR #410, opened 2023-03-02
Replace nn.GroupNorm with Fp32GroupNorm for mixed precision training

RdoubleA:
Summary: To support mixed-precision / fp16 training, replaced all GroupNorm layers with Fp32GroupNorm, transferred from [fairseq](https://github.com/facebookresearch/fairseq/blob/0338cdc3094ca7d29ff4d36d64791f7b4e4b5e6e/fairseq/modules/fp32_group_norm.py#L13). This ensure normalization in fp32, even if input is fp16.

Reviewed By: kartikayk

Differential Revision: D43581790

### DIFF
--- a/tests/modules/layers/test_normalizations.py
+++ b/tests/modules/layers/test_normalizations.py
@@ -5,11 +5,18 @@
 # LICENSE file in the root directory of this source tree.
 
 import torch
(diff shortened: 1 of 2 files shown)


pbontrager (approved):
Looks good

PR #424, opened 2023-04-27
Move mask creation to init + set is_causal flag for CLIP

ankitade:
Summary:
1. Mask was being created unnecessarily for each forward call. move it to init now
2. Set is_causal flag to ensure we always go down flash attn + no mask checking is done inside transformer 

Test plan:
pytest tests (ignore the circle ci failure, unrelated)

### DIFF
--- a/torchmultimodal/models/clip/text_encoder.py
+++ b/torchmultimodal/models/clip/text_encoder.py
@@ -69,6 +69,11 @@ class CLIPTextEncoder(nn.Module):
         self.ln_final = Fp32LayerNorm(width)
         self.projection = nn.Linear(width, embedding_dim, bias=False)
 
+        self.mask = torch.full(
+            (self.context_length, self.context_length),
+            float("-inf"),
+        ).triu(1)
+
         if use_clip_init:
             self.initialize_parameters()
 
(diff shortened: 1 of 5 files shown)


torchmultimodal/models/clip/text_encoder.py
```diff
@@ -69,6 +69,11 @@ def __init__(
         self.ln_final = Fp32LayerNorm(width)
         self.projection = nn.Linear(width, embedding_dim, bias=False)
 
+        self.mask = torch.full(
```
cpuhrsch, line 72: Once you use SDPA directly you should be able to avoid allocating this entirely. If you need to migrate checkpoints you could look into using a hook. For example the following code concatenates three separate weights into a single one for input projections of nn.MHA (courtesy of LLaMa).

```
    def __init__(
        self,
        dim: int,
        head_dim: int,
        n_heads: int,
        n_kv_heads: int,
    ):
        [...]
        self._register_load_state_dict_pre_hook(self.load_hook)

    def load_hook(
        self,
        state_dict,
        prefix,
        local_metadata,
        strict,
        missing_keys,
        unexpected_keys,
        error_msgs,
    ):
        if prefix + "wq.weight" in state_dict:
            wq = state_dict.pop(prefix + "wq.weight")
            wk = state_dict.pop(prefix + "wk.weight")
            wv = state_dict.pop(prefix + "wv.weight")
            state_dict[prefix + "wqkv.weight"] = torch.cat([wq, wk, wv])
```

PR #434, opened 2023-07-07
Fix AC order for FLAVA example

rohan-varma:
Summary: There have been some fixes to the non-reentrant AC and FLAVA model now works with it.

Differential Revision: D47218368


ankitade (commented):
please update the test plan with a cmd you ran

PR #442, opened 2023-08-04
Support returning hidden state in CLIPTextEncoder

22quinn:
Summary:
Added a `return_hidden_state` arg in CLIPTextEncoder. If set to True, forward will return a tuple of final embedding and the last hidden state.

Test plan:
pytest tests/models/clip/test_text_encoder.py

cc @abhinavarora Moved code to here for easier review and testing.

### DIFF
--- a/tests/models/clip/test_text_encoder.py
+++ b/tests/models/clip/test_text_encoder.py
@@ -28,7 +28,6 @@ class TestCLIPTextEncoder:
             context_length=context_length,
             heads=heads,
             width=width,
-            return_hidden_state=False,
         ):
             return CLIPTextEncoder(
                 embedding_dim=embedding_dim,
@@ -36,7 +35,6 @@ class TestCLIPTextEncoder:
                 context_length=context_length,
                 width=width,
                 heads=heads,
-                return_hidden_state=return_hidden_state,
             )
 
         return build_encoder, build_text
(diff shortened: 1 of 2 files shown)


torchmultimodal/models/clip/text_encoder.py
```diff
@@ -120,11 +127,13 @@ def forward(self, text: Tensor) -> Tensor:
...
+        hidden_state = self.ln_final(embeddings)
         # take features from the eot embedding (the highest number in each sequence)
         embeddings = self.projection(
-            embeddings[torch.arange(embeddings.shape[0]), text.argmax(dim=-1)]
+            hidden_state[torch.arange(hidden_state.shape[0]), text.argmax(dim=-1)]
         )
         # embeddings now has size [bs, embedding_dim]
 
+        if self.return_hidden_state:
+            return embeddings, hidden_state
```
ankitade, line 138: nit: use named tuple

tests/models/clip/test_text_encoder.py
```diff
@@ -117,6 +121,47 @@ def test_forward(self, start):
...
+        build_encoder, build_text = start
+        text = build_text(text_length=3)
+
+        text_encoder = build_encoder(
+            context_length=3, width=4, return_hidden_state=True
+        )
+        assert isinstance(text_encoder, torch.nn.Module)
+
+        actual_clip_init, actual_hidden_state = text_encoder(text)
+        print(actual_hidden_state)
```
ankitade, line 134: remove

tests/models/clip/test_text_encoder.py
```diff
@@ -117,6 +121,47 @@ def test_forward(self, start):
...
+        expected_clip_init = torch.Tensor(
+            [
+                [-0.366838, -1.596611, -0.330413, -0.593790],
+                [-0.790419, 0.876780, -0.970667, -0.727134],
+            ]
+        )
+        expected_hidden_state = torch.Tensor(
+            [
+                [
+                    [6.348165e-01, -4.137459e-02, -1.604239e00, 1.010798e00],
```
ankitade, line 144: nit: can we limit to 4 decimal places given atol is being set

torchmultimodal/models/clip/text_encoder.py
```diff
@@ -77,6 +82,8 @@ def __init__(
         if use_clip_init:
             self.initialize_parameters()
 
+        self.return_hidden_state = return_hidden_state
```
ankitade, line 85: you can make this an arg to fwd

torchmultimodal/models/clip/text_encoder.py
```diff
@@ -120,11 +127,13 @@ def forward(self, text: Tensor) -> Tensor:
 
         # [n_ctx, bs, transformer.width] -> [bs, n_ctx, transformer.width]
         embeddings = torch.permute(embeddings, (1, 0, 2))
-        embeddings = self.ln_final(embeddings)
+        hidden_state = self.ln_final(embeddings)
         # take features from the eot embedding (the highest number in each sequence)
         embeddings = self.projection(
```
abhinavarora, line 132: I would suggest renaming this to something like `projected_embeddings` to avoid re-using the `embeddings` name as it was used before.

22quinn:
Thanks both for the review! I've addressed all comments. buck test looks good.

> buck test @mode/dev-nosan //torchmultimodal/tests
Tests finished: Pass 317. Fail 0. Fatal 0. Skip 6. Build failure 0

ankitade:
can you import the diff and land it there

PR #446, opened 2023-08-15
add parallel_attention_blocks

lessw2020:
Summary:
This PR adds Parallel Attention blocks to Torch MultiModal. 
There are 3 main additions:
a - Rotary Embeddings 
b - RMS Norm
c - Parallel_Blocks 

Test plan:
Code has been tested separately in ViT and LLM applications and with rotary unit tests.  
However, a general unit test needs to be added along with rotary unit tests being migrated.

### DIFF
--- a/torchmultimodal/modules/layers/normalizations.py
+++ b/torchmultimodal/modules/layers/normalizations.py
@@ -5,7 +5,7 @@
 # LICENSE file in the root directory of this source tree.
 
 from typing import Any
-
+import torch
 from torch import nn, Tensor
 
 
@@ -45,3 +45,17 @@ class Fp32GroupNorm(nn.GroupNorm):
             self.eps,
         )
         return output.type_as(x)
+
+
+class RMSNorm(nn.Module):
+    def __init__(self, dim: int, eps: float = 1e-6):
+        super().__init__()
+        self.eps = eps
+        self.weight = nn.Parameter(torch.ones(dim))
+
+    def _norm(self, x):
+        return x * torch.rsqrt(x.pow(2).mean(-1, keepdim=True) + self.eps)
+
+    def forward(self, x):
+        output = self._norm(x.float()).type_as(x)
+        return output * self.weight
--- a/torchmultimodal/modules/layers/parallel_blocks.py
+++ b/torchmultimodal/modules/layers/parallel_blocks.py
@@ -1,0 +1,259 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+
+import torch
+import torch.nn as nn
+import torch.nn.functional as F
+from packaging import version
+from typing import Optional
+import math
+
+from normalizations import RMSNorm
+from position_embedding import RotaryEmbedding
+
+
+def repeat_kv(x: torch.Tensor, n_rep: int) -> torch.Tensor:
+    """expands kv_heads to match q num_heads
+    via
+    torch.repeat_interleave(x, dim=2, repeats=n_rep)"""
+
+    bs, slen, n_kv_heads, head_dim = x.shape
+
+    if n_rep == 1:
+        return x
+    return (
+        x[:, :, :, None, :]
+        .expand(bs, slen, n_kv_heads, n_rep, head_dim)
+        .reshape(bs, slen, n_kv_heads * n_rep, head_dim)
+    )
+
+
+class ParallelAttentionBlock(nn.Module):
+    """
+    Transformer layer multi-head attention and MLP, in a parallelized fashion rather than sequential,
+    with optional attention masking.
+    Inspired by PaLM:  https://arxiv.org/abs/2204.02311
+
+    * We use SwiGLU for the activation function
+    * SwiGLU will approximate same total num params as traditional MLP with GELU
+    * Cross Attention is not enabled here (but available)
+    * MQA and GQA are enabled - modify heads via 'num_heads_group_query_attn'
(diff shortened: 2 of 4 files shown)


torchmultimodal/modules/layers/normalizations.py
```diff
@@ -45,3 +45,17 @@ def forward(self, x: Tensor) -> Tensor:
             self.eps,
         )
         return output.type_as(x)
+
+
+class RMSNorm(nn.Module):
```
rohan-varma, line 50: add docstring

torchmultimodal/modules/layers/parallel_blocks.py
```diff
@@ -0,0 +1,259 @@
...
+
+class ParallelAttentionBlock(nn.Module):
+    """
+    Transformer layer multi-head attention and MLP, in a parallelized fashion rather than sequential,
+    with optional attention masking.
+    Inspired by PaLM:  https://arxiv.org/abs/2204.02311
+
+    * We use SwiGLU for the activation function
+    * SwiGLU will approximate same total num params as traditional MLP with GELU
+    * Cross Attention is not enabled here (but available)
```
rohan-varma, line 38: what does "but available" mean?

torchmultimodal/modules/layers/parallel_blocks.py
```diff
@@ -0,0 +1,259 @@
...
+    * MQA and GQA are enabled - modify heads via 'num_heads_group_query_attn'
```
rohan-varma, line 39: this seems for GQA, how would I use MQA?

torchmultimodal/modules/layers/parallel_blocks.py
```diff
@@ -0,0 +1,259 @@
...
+        attention_dropout: float = 0.0,
+        use_group_query_attention: bool = True,
+        num_heads_group_query_attention: int = 1,
+        use_in_projection_bias: bool = True,
+        use_out_projection_bias: bool = True,
+        use_weight_init: bool = True,
+        num_layers: int = 1,
+        use_rms_norm: bool = True,
+        use_rotary_embeddings: bool = False,
+        max_expected_seq_len: int = 2048,  # needed only if using rotary
```
rohan-varma, line 64: so if I use rotary embeddings, my sequence length is fixed for pretraining, finetuning, inference? cc @ankitade

rohan-varma (commented):
looking forward to the unittests, will do more thorough review after that

torchmultimodal/modules/layers/parallel_blocks.py
lessw2020: ah - we support cross attention in the main codebase for parallel attention.  However, it's not really useful for TMM, so I removed it for this PR. (hence the 'but available').  Let me expand that comment to note it's in the main codebase.

torchmultimodal/modules/layers/parallel_blocks.py
lessw2020: Easy (but I will expand this in the comments to clarify):
num_heads_group_query_attn = 1 and you now have MQA.
num_heads_group_query_attn > 1 and < Q_heads = GQA.  
I should probably also clarify that MQA heads needs to be a multiple of Q heads (there is an assert check so you ultimately can't miss it, but might be nicer to note in docstring).

torchmultimodal/modules/layers/parallel_blocks.py
lessw2020: yes fixed for pretraining. However,  you can do rotary embedding interpolation and expand it though with finetuning (apparently up to 32x).  
https://arxiv.org/pdf/2306.15595.pdf

torchmultimodal/modules/layers/position_embedding.py
```diff
@@ -169,3 +169,98 @@ def forward(self, t: Tensor) -> Tensor:
...
+        Args
+        ----
+        dim : int
+            Per-head embedding dimension
+        max_position_embeddings : int
+            Maximum expected sequence length for the model, if exceeded the cached freqs will be recomputed
+        ratio: int
+            The ratio for the geometric progression to compute the rotation angles
+        """
+        super(RotaryEmbedding, self).__init__()
```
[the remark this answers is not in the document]
lessw2020: good catch, fixed! (though breaking this whole PR into 3 different ones, so fixed in Rotary PR that's coming).

torchmultimodal/modules/layers/parallel_blocks.py
```diff
@@ -0,0 +1,259 @@
...
+        self.attention_dropout = nn.Dropout(attention_dropout)
+        self.mlp_dropout = nn.Dropout(projection_dropout)
+
+        # weight init
+        self.num_layers = num_layers
+        self.use_weight_init = use_weight_init
+        if self.use_weight_init:
+            assert (
+                self.num_layers > 1
+            ), f"Need to pass in global num layers for weight init"
```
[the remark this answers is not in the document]
lessw2020: it's a valid point - the design was trying to be as portable/encapsulated as possible initially, but with this upstream can perhaps provide a container class to go with it and remove this.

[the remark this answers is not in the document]
lessw2020: update - I started a wrappers.py class but atm Rohan does not feel we should add it.  Thus, will leave this out for now though will continue to build the wrappers.py class separately (as it has LLMWrapper, ViTWrapper, Encoder, Decoder).

PR #447, opened 2023-08-16
Add new transformer implementation to oss

ankitade:
Summary:
Cloned the existing transformer impl (will add the decoder in follow up diff). There is some odd discrepancy in sdpa results between fbcode and nightly so changed 1 test values. lets see if the test starts failing after somedays i.e. if fbcode sdpa gets picked up in nightly.

Test plan:
copied over existing tests

### DIFF
--- a/tests/modules/layers/test_transformer.py
+++ b/tests/modules/layers/test_transformer.py
@@ -1,0 +1,197 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+import pytest
+
+import torch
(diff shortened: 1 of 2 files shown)


tests/modules/layers/test_transformer.py
```diff
@@ -0,0 +1,197 @@
...
+
+    @pytest.fixture
+    def inputs(self):
+        return Tensor([[[1, 2], [4, 2], [1, 1]]])
+
+    @pytest.mark.parametrize(
+        "norm_first, expected_output",
+        [
+            (True, Tensor([[[15.0, 16.0], [18.0, 16.0], [15.0, 15.0]]])),
+            (False, Tensor([[[0.0, 2.0], [2.0, 0.0], [0.9414, 0.9414]]])),
```
ankitade, line 48: for reference, this is the test i changed (last row of the tensor)

ebsmothers (approved):

PR #448, opened 2023-08-16
Add RMSNorm  (part of parallel_attn_blocks)

lessw2020:
Summary:
This PR adds RMSNorm.  
It is used in parallel_attention_blocks but moving the earlier mono PR for that into smaller chunks for easier review and landing. 

Test plan:
added two units tests:
test_rms_norm_fp32
test_rms_core_algo  (compares F.norm version vs PR version)

Fixes:
Adds requested doc string to RMSNorm from earlier PR review.

### DIFF
--- a/tests/modules/layers/test_normalizations.py
+++ b/tests/modules/layers/test_normalizations.py
@@ -5,7 +5,15 @@
 # LICENSE file in the root directory of this source tree.
 
 import torch
-from torchmultimodal.modules.layers.normalizations import Fp32GroupNorm, Fp32LayerNorm
+import torch.nn.functional as F
+
+from tests.test_utils import gpu_test
+
+from torchmultimodal.modules.layers.normalizations import (
+    Fp32GroupNorm,
+    Fp32LayerNorm,
+    RMSNorm,
+)
 
 
 def test_fp32layernorm():
@@ -20,3 +28,45 @@ def test_fp32groupnorm():
     norm = Fp32GroupNorm(2, 4)
     output = norm(x)
     assert output.dtype == torch.float16
+
+
+def test_rms_norm_fp32return():
+    """verify type is returned as fp32"""
+    dims = 512
+    x = torch.empty(dims, dtype=torch.float16)
+    norm = RMSNorm(
+        dims,
+    )
+    output = norm(x)
+    assert output.dtype == torch.float32
+
+
+@gpu_test(1)
+def test_rms_norm_core_algo():
+    """compare RMSNorm with RMSNorm using F.norm version"""
+
+    dims = 1024
+    x = torch.empty(dims, dtype=torch.float16, device="cuda")
+    x_clone = x.clone().detach()
+
+    class RMSNormFunctional(torch.nn.Module):
+        def __init__(self, dim, eps=1e-6):
+            super().__init__()
+            self.scale = dim**0.5
+            self.weights = torch.nn.Parameter(torch.ones(dim))
+            self.eps = eps
+
+        def forward(self, x):
+            return F.normalize(x, p=2, dim=-1, eps=self.eps) * self.scale * self.weights
+
+    base_norm = RMSNorm(
+        dims,
+    ).to("cuda")
+    backup_norm = RMSNormFunctional(
+        dims,
+    ).to("cuda")
+
+    output_base_rms = base_norm(x)
+    output_backup_rms = backup_norm(x_clone)
+
+    assert torch.allclose(output_base_rms, output_backup_rms)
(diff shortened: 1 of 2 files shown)


rohan-varma (approved):
LG! I would just absorb the fp32 type check test into the other one for brevity.

torchmultimodal/modules/layers/normalizations.py
```diff
@@ -45,3 +46,25 @@ def forward(self, x: Tensor) -> Tensor:
             self.eps,
         )
         return output.type_as(x)
+
+
+class RMSNorm(nn.Module):
+    """Root Mean Square layer normalization
+    as proposed in: https://arxiv.org/abs/1910.07467
+
```
ankitade, line 54: i think this is copy paste of llama. can u add a link

torchmultimodal/modules/layers/normalizations.py
```diff
@@ -45,3 +46,25 @@ def forward(self, x: Tensor) -> Tensor:
...
+    params:
+    dim = model size
+    eps = epsilon
```
ankitade, line 57: nit: follow this format https://github.com/facebookresearch/multimodal/blob/main/torchmultimodal/models/dalle2/dalle2_decoder.py#L48 (type, default values etc)

tests/modules/layers/test_normalizations.py
```diff
@@ -20,3 +28,45 @@ def test_fp32groupnorm():
...
+    dims = 512
+    x = torch.empty(dims, dtype=torch.float16)
+    norm = RMSNorm(
+        dims,
+    )
+    output = norm(x)
+    assert output.dtype == torch.float32
+
+
+@gpu_test(1)
```
ankitade, line 44: why is gpu required? for fp16 support?

tests/modules/layers/test_normalizations.py
```diff
@@ -20,3 +28,45 @@ def test_fp32groupnorm():
...
+def test_rms_norm_core_algo():
+    """compare RMSNorm with RMSNorm using F.norm version"""
+
+    dims = 1024
+    x = torch.empty(dims, dtype=torch.float16, device="cuda")
+    x_clone = x.clone().detach()
+
+    class RMSNormFunctional(torch.nn.Module):
```
ankitade, line 52: this test is little weirdly written. u should just assert on output values directly (that u hand calculated or got from reference impl)

tests/modules/layers/test_normalizations.py
```diff
@@ -20,3 +28,45 @@ def test_fp32groupnorm():
     norm = Fp32GroupNorm(2, 4)
     output = norm(x)
     assert output.dtype == torch.float16
+
+
+def test_rms_norm_fp32return():
```
ankitade, line 33: why dont we just do 1 test with fp32 and assert on output values?

torchmultimodal/modules/layers/normalizations.py
lessw2020: added

torchmultimodal/modules/layers/normalizations.py
lessw2020: updated to use similar format

tests/modules/layers/test_normalizations.py
lessw2020: it's not required per se - however for finer precision / lower precision dtypes I have seen times where the PT cpu version impl does not match the gpu version impl.
Since we care about gpu, figured it's better to run this test on gpu. 
Can remove this if it creates to much test overhead?

tests/modules/layers/test_normalizations.py
lessw2020: my idea was to separate out the tests esp if we add more dtype support (ala fp8) but will merge this into one for brevity.

tests/modules/layers/test_normalizations.py
lessw2020: ok moved to fixed tensor tests instead with pre-calced results.

tests/modules/layers/test_normalizations.py
lessw2020: removed the gpu decorator since CI does not necessarily have gpu.

tests/modules/layers/test_normalizations.py
```diff
@@ -20,3 +26,38 @@ def test_fp32groupnorm():
     norm = Fp32GroupNorm(2, 4)
     output = norm(x)
     assert output.dtype == torch.float16
+
+
+def test_rms_norm_core_algo():
```
ankitade, line 31: nit: rename to test_rms_norm

ankitade (approved):

tests/modules/layers/test_normalizations.py
lessw2020: that makes sense.  I'll resolved in the fb diff.

PR #454, opened 2023-08-21
Return either hidden state or projected embeddings not both

abhinavarora:
Summary: This is to fix an error when torchscriptifying the transform.

Differential Revision: D48380633

### DIFF
--- a/tests/models/clip/test_text_encoder.py
+++ b/tests/models/clip/test_text_encoder.py
@@ -126,16 +126,7 @@ class TestCLIPTextEncoder:
         text_encoder = build_encoder(context_length=3, width=4)
         assert isinstance(text_encoder, torch.nn.Module)
 
-        out = text_encoder(text, return_hidden_state=True)
-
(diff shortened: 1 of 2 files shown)


tests/models/clip/test_text_encoder.py
```diff
@@ -150,12 +141,6 @@ def test_forward_return_hidden_state(self, start):
                 ],
             ]
         )
-        assert_expected(
```
rohan-varma, line 153: shouldn't we have a test that also checks the projected embeddings are as expected?

22quinn: There's another test case `test_forward` checking the projected embeddings are as expected when `return_hidden_state=False`. In this PR, we removed the feature returning both the hidden state and projected embeddings at the same time, hence removing this assert.

rohan-varma: thanks for the clarification!

PR #457, opened 2023-08-27
Add Parallel_Attention_Blocks (3 of 3)

lessw2020:
Summary:
This PR adds the main and final part for upstreaming Parallel Attention Blocks, specifically the actual Parallel Attn Blocks class itself. 
This supports both MHA, MQA and GQA attention head setups. 
RMSNorm has already landed and so can be used.  
Cross Attention has been removed as requested.  

Test plan:
Added 3 unit tests - one each for Parallel Attn Blocks using MHA, MQA and GQA.
Within each test, the number of query heads and number of KV heads are tested to ensure appropriate head counts (i.e. for GQA, num_KV heads is set to 2 and then verified.
From there, a Parallel Attn Layer runs a forward pass on a fixed single input tensor and the first row of the output is checked as is the attn_output shape.

### DIFF
--- a/tests/modules/layers/test_parallel_attention.py
+++ b/tests/modules/layers/test_parallel_attention.py
@@ -1,0 +1,100 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+import pytest
+
+# import torch
+from tests.test_utils import assert_expected  # , init_weights_with_constant
+
+# from torch import nn
+from torchmultimodal.modules.layers.parallel_attention import ParallelAttentionBlock
+
+
+class TestParallelAttentionBlocks:
+    @pytest.fixture
+    def embedding_dim(self):
+        return 64
+
+    @pytest.fixture
+    def total_layers(self):
+        return 1
+
+    @pytest.fixture
+    def mqa_num_heads(self):
+        return 1
+
+    @pytest.fixture
+    def gqa_num_heads(self):
+        return 2
+
+    @pytest.fixture
+    def num_heads(self):
+        return 16
+
+    @pytest.fixture
+    def mha_parallel_attention(self, embedding_dim, num_heads, total_layers):
+        print(f"{embedding_dim=}, {num_heads=}, {total_layers=}")
+        pab_mha = ParallelAttentionBlock(
+            emb_dimension=embedding_dim,
+            num_heads=num_heads,
+            use_group_query_attention=False,
+            num_layers=total_layers,
+            use_weight_init=True,
+        )
+        pab_mha.eval()
+        return pab_mha
+
+    @pytest.fixture
+    def gqa_parallel_attention(
+        self, embedding_dim, num_heads, total_layers, gqa_num_heads
+    ):
+        print(f"{embedding_dim=}, {num_heads=}, {total_layers=}")
+        pab_gqa = ParallelAttentionBlock(
+            emb_dimension=embedding_dim,
+            num_heads=num_heads,
+            use_group_query_attention=True,
+            num_heads_group_query_attention=gqa_num_heads,
+            num_layers=total_layers,
+            use_weight_init=True,
+        )
+        pab_gqa.eval()
+        return pab_gqa
+
+    @pytest.fixture
+    def mqa_parallel_attention(
+        self, embedding_dim, num_heads, total_layers, mqa_num_heads
+    ):
+        print(f"{embedding_dim=}, {num_heads=}, {total_layers=}")
+        pab_mqa = ParallelAttentionBlock(
+            emb_dimension=embedding_dim,
+            num_heads=num_heads,
+            use_group_query_attention=True,
+            num_heads_group_query_attention=mqa_num_heads,
+            num_layers=total_layers,
+            use_weight_init=True,
+        )
+        pab_mqa.eval()
+        return pab_mqa
+
+    def test_mha_parallel_attention(self, mha_parallel_attention, num_heads):
+        # confirm all K and V keys match Q (i.e. MHA)
+        assert_expected(num_heads, mha_parallel_attention.num_kv)
+
+    def test_mqa_parallel_attention(
+        self, mqa_parallel_attention, num_heads, mqa_num_heads
+    ):
+        print("in test")
+
+        # confirm all K and V keys match MQA num heads (i.e. MQA == 1)
+        assert_expected(mqa_num_heads, mqa_parallel_attention.num_kv)
+
+    def test_gqa_parallel_attention(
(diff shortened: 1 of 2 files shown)


tests/modules/layers/test_parallel_attention.py
```diff
@@ -0,0 +1,210 @@
...
+        num_heads,
+        embedding_dim,
+        max_seq_len,
+    ):
+        # confirm all K and V keys match Q (i.e. MHA)
+        assert_expected(num_heads, mha_parallel_attention.num_kv)
+        # confirm num Q matches num_heads
+        assert_expected(num_heads, mha_parallel_attention.num_heads)
+
+        # input_ones = torch.ones(dims, dtype=torch.float)
```
rohan-varma, line 98: nit: rm

tests/modules/layers/test_parallel_attention.py
```diff
@@ -0,0 +1,210 @@
...
+                165.9994,
+                193.9994,
+                93.0001,
+            ],
+            dtype=torch.float32,
+        )
+        fixed_output_shape = torch.Size([1, max_seq_len, embedding_dim])
+
+        assert_expected(fixed_result_firstrow, attn_output[0][0], rtol=0, atol=1e-4)
+        assert_expected(fixed_output_shape, attn_output.shape)
```
rohan-varma, line 127: @ebsmothers, do we / should we do any additional testing besides verifying the first row of outputs?

tests/modules/layers/test_parallel_attention.py
```diff
@@ -0,0 +1,210 @@
...
+                213.0028,
+                213.9989,
+                12.0008,
+                190.9990,
+            ],
+            dtype=torch.float32,
+        )
+        fixed_output_shape = torch.Size([1, max_seq_len, embedding_dim])
+        assert_expected(fixed_output_shape, attn_output.shape)
+        # print(f"{attn_output[0][0]}")
```
rohan-varma, line 168: nit: remove

torchmultimodal/modules/layers/parallel_attention.py
```diff
@@ -0,0 +1,269 @@
...
+    def forward(
+        self,
+        x: torch.Tensor,
+        cross_x: Optional[torch.Tensor] = None,
+        mask: Optional[torch.Tensor] = None,
+        cross_mask: Optional[torch.Tensor] = None,
+        rel_pos_bias: Optional[torch.Tensor] = None,
+        has_causal_mask: bool = False,
+    ) -> torch.Tensor:
+        """TODO:   No KV cache support yet"""
```
rohan-varma, line 186: shall we file an issue for this?

rohan-varma (commented):
LG overall, @ebsmothers do you mind taking a pass through this as well?

torchmultimodal/modules/layers/parallel_attention.py
```diff
@@ -0,0 +1,269 @@
...
+    * Parallel blocks have automated weight initialization via _init_weights.
+    * Please pass in num_layers of your model in num_layers for the weight initialization.
+    """
+
+    def __init__(
+        self,
+        emb_dimension: int,
+        num_heads: int,
+        head_dimension: int = None,
+        mlp_expansion_ratio: float = 2.6875,  # 8/3 is param matching
```
ebsmothers, line 61: nit: any reason not to just use an integer value here and skip the multiplication by ratio?

tests/modules/layers/test_parallel_attention.py
```diff
@@ -0,0 +1,210 @@
...
+    ):
+        # confirm all K and V keys match Q (i.e. MHA)
+        assert_expected(num_heads, mha_parallel_attention.num_kv)
+        # confirm num Q matches num_heads
+        assert_expected(num_heads, mha_parallel_attention.num_heads)
+
+        # input_ones = torch.ones(dims, dtype=torch.float)
+
+        x = torch.randint(0, 256, (1, max_seq_len, embedding_dim))  # bs =1,
+        attn_output = mha_parallel_attention(x)
```
ebsmothers, line 101: Would also consider adding a test case covering mask and/or rel_pos_bias args

tests/modules/layers/test_parallel_attention.py
```diff
@@ -0,0 +1,210 @@
...
+        max_seq_len,
+    ):
+        # confirm all K and V keys match Q (i.e. MHA)
+        assert_expected(num_heads, mha_parallel_attention.num_kv)
+        # confirm num Q matches num_heads
+        assert_expected(num_heads, mha_parallel_attention.num_heads)
+
+        # input_ones = torch.ones(dims, dtype=torch.float)
+
+        x = torch.randint(0, 256, (1, max_seq_len, embedding_dim))  # bs =1,
```
ebsmothers, line 100: Can probably just make this a fixture (since I think it's used in all test cases)

torchmultimodal/modules/layers/parallel_attention.py
```diff
@@ -0,0 +1,269 @@
...
+        self.use_in_projection_bias = use_in_projection_bias
+        self.use_out_projection_bias = use_out_projection_bias
+
+        # previous init params, moved to internal defaults for streamlining
+        normalization_layer = RMSNorm if use_rms_norm else nn.LayerNorm
+        self.mlp_activation = nn.SiLU()
+
+        self.num_q = 1
+
+        self.in_proj_dims = [
```
ebsmothers, line 130: nit: add a comment either here or on self.in_proj definition about the fusing you're doing. I think fusing of the MLP and gate could be a bit unusual for those unfamiliar with parallel attention

torchmultimodal/modules/layers/parallel_attention.py
lessw2020: oh there is a reason for this actually - the idea was to mimic the same param count as when most people do mlp_expansion =4, but keep it within a power of 8 regime.
A common mistake made is when people compare swiglu vs say GeLU they will claim swiglu is slower...but what's really happening is that since swiglu uses a gate you have more total params if you leave the same 4.0 multiplication factor and simply drop in swiglu in the mlp.
Therefore, this 2.6875 gives you ~ the same params as activation + mul factor 4, but within the power of 8  regime (this was from a paper that showed this gave you a slight efficiency gain...can't remember if it was power of 8 or 16 but something like that).  (Hence the comment that exact match is 8/3).

torchmultimodal/modules/layers/parallel_attention.py
```diff
@@ -0,0 +1,269 @@
...
+from typing import Optional
+
+import torch
+import torch.nn as nn
+import torch.nn.functional as F
+from packaging import version
+
+from torchmultimodal.modules.layers.normalizations import RMSNorm
+
+# from position_embedding import RotaryEmbedding
```
rohan-varma, line 17: remove this?

PR #460, opened 2023-09-05
add tensor_hash

lessw2020:
Summary:
This adds a tensor hash function for multi-dim tensors.  
It's purpose is to allow easy full tensor verification on forward tensors for other unit tests
by storing the hash of the answer tensor.  
This strikes a middle ground between simple verification (i.e. first row, mean of an axis) and every item in the tensor verification which would require storing the entire tensor in a unit test.

Usage = 
hash the answer tensor, save in your unit test.  
Run a forward in unit test, hash the output tensor. 
Compare hashed tensors to confirm unit test result.

Test plan:
Verified a + .01 change to single item in a multi-dim tensor is detected, verified identical tensors generate identical hashes.
Since this is a unit test function, not sure that unit test for a unit test function is needed as will not change unless function itself is modified.

Fixes #{issue number}

### DIFF
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -191,3 +191,34 @@ def assert_expected_namedtuple(
 def init_weights_with_constant(model: nn.Module, constant: float = 1.0) -> None:
     for p in model.parameters():
         nn.init.constant_(p, constant)
+
+
+def tensor_hash(x: torch.tensor, scaling=0.05, buckets=1000) -> torch.tensor:
+    """hashes a multi-dim tensor for unit test verification
+
+    usage: hash a forward tensor to serve as correct answer
+    compare unit test result hashes with hashed answer using allclose
+
+    example:
+    t1 = torch.randn(1,16,32)  # answer tensor
+    t2 = t1.clone()
+    t2[0][0][0] += .01  # clone but modified with a single .01 change
+
+    t1_hash = tensor_hash(t1)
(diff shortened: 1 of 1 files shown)


lessw2020:
unit test failure is not related to this PR

pbontrager:
Could you provide a little more context for this function? It seems to compare sums for the final dimension, does this provide more precision than mean? It's just one extra dimension of data. I guess this would be more likely to catch tensors that had the right data but the wrong shape, is that the primary use case? To add clarity to my question, I am wondering what the potential mistakes are that are not caught by mean or mean(-1) and if this implementation of a hash catches all of them.

lessw2020:
Hi @pbontrager,
My concern with the mean test is that it's an average, and multiple combinations of numbers could potentially end up at the same mean, so I felt the mean test alone is not sufficient to fully check a tensor. 

I can't guarantee that a hash will catch all issues and probably using both would result in the best coverage, but I show a simple example below where using the mean fails to detect the difference of .01 to an entry,  vs the hash does using the default all_close settings that are in the multi-modal tests.

Here adding .01 to a single entry is not detected by mean (see line 18 allclose = True), but is detected with the hashing. 

hash_and_mean

rohan-varma (approved):
just testing

rohan-varma (commented):
test

ebsmothers (approved):

PR #462, opened 2023-09-06
add Alibi positional embeddings

lessw2020:
Summary:
This PR adds Alibi positional embeddings class.  (per the Alibi paper https://arxiv.org/abs/2108.12409)
This generates the Alibi attn mask to be added post QKT/sqrt(k.dim) and replaces the usual sinusoidal type positional embeddings. 
The usage is designed to be instantiated outside the transformer block loop based on max_seq_length, and the layers retrieve the attn mask based on current seq length (thus only a single mask buffer needs to be created).

Test plan:
I tested by running in a 200M gpt2 model along with 10% of openwebtext to compare curves between learned embeddings (default in gpt2) and alibi. 
alibi_training_curves

I also added a unit test with three tests:
a - shape of the alibi mask
b - verify first head row entry
c - verify last head last row entry
Note that half the mask is -inf, but in trying to use allclose with -inf, they will not match...so I targeted entries that have only real numbers.


lessw2020:
unit test failure is not related.

lessw2020:
test failure is not related - appears to be rounding issue:
test_model.py::TestAudioMaskedAutoEncoder::test_audio_mae_train_masking - AssertionError: actual: 512.999755859375, expected: 513.

lessw2020:
I reran the training with the updates to confirm:
proper_causal_mask

torchmultimodal/modules/layers/position_embedding.py
```diff
@@ -169,3 +170,106 @@ def forward(self, t: Tensor) -> Tensor:
...
+
+        # paper authors note that they only trained models that have 2^a heads for some a.
+        # This has beneficial properties related to input being power of 2.
+        # Closest power of 2 below is workaround for when num of heads is not power of 2
+
+        closest_power_of_2 = 2 ** math.floor(math.log2(num_heads))
+        return (
+            get_slopes_power_of_2(closest_power_of_2)
+            + get_slopes_power_of_2(2 * closest_power_of_2)[0::2][
+                : num_heads - closest_power_of_2
```
daviswer, line 273: Looks good, my one strictly cosmetic gripe is that this returns slopes in a fairly unintuitive order for non-powers of 2: for 12 heads, for example, the scales are `2^(1,2,3,4,5,6,7,8,.5,1.5,2.5,3.5)`. Rather than doing:
```
a = get_slopes_power_of_2(closest_power_of_2)
b = get_slopes_power_of_2(2 * closest_power_of_2)[0::2][:num_heads - closest_power_of_2]
return a+b
```
You could return something like:
```
return [x for pair in zip(b,a) for x in pair] + a[len(b):]
```
which, while less legible, would return scales in order (i.e. for 12 heads this gives `2^(.5,1,1.5,2,2.5,3,3.5,4,5,6,7,8)`). 

Up to you if you think that tradeoff is worth it.

lessw2020: Thanks for the feedback @daviswer  - I agree, I think the consistency and symmetry of the heads (regardless of power of 2) has merit and will update.

lessw2020: I've updated as well as added unit test coverage for non power of 2 heads case. 
Thanks again!

torchmultimodal/modules/layers/position_embedding.py
```diff
@@ -169,3 +170,108 @@ def forward(self, t: Tensor) -> Tensor:
...
+        def get_slopes_power_of_2(n: int) -> List[float]:
+            start = 2 ** (-(2 ** -(math.log2(n) - 3)))
+            ratio = start
+            return [start * ratio**i for i in range(n)]
+
+        if math.log2(num_heads).is_integer():
+            return get_slopes_power_of_2(num_heads)
+
+        # paper authors note that they only trained models that have 2^a heads for some a.
+        # This has beneficial properties related to input being power of 2.
```
ebsmothers, line 266: Do you know what these properties are? Tbh I am confused by this because even if n is a power of 2 some of the ratios will not be rational for n > 8

torchmultimodal/modules/layers/position_embedding.py
```diff
@@ -169,3 +170,108 @@ def forward(self, t: Tensor) -> Tensor:
...
+        # Closest power of 2 below is workaround for when num of heads is not power of 2
+        # Slopes are returned in ordered sequence to keep symmetry.
+
+        closest_power_of_2 = 2 ** math.floor(math.log2(num_heads))
+
+        a = get_slopes_power_of_2(closest_power_of_2)
+        b = get_slopes_power_of_2(2 * closest_power_of_2)[0::2][
+            : num_heads - closest_power_of_2
+        ]
+        return [x for pair in zip(b, a) for x in pair] + a[len(b) :]
```
ebsmothers, line 277: Imo this is hard to parse. Agree with @daviswer's comment about returning values in order but could we just do sorted(a+b)? (Maybe I'm missing a tricky case.. if so a comment explaining this would suffice instead)

torchmultimodal/modules/layers/position_embedding.py
```diff
@@ -169,3 +170,108 @@ def forward(self, t: Tensor) -> Tensor:
...
+            ratio = start
+            return [start * ratio**i for i in range(n)]
+
+        if math.log2(num_heads).is_integer():
+            return get_slopes_power_of_2(num_heads)
+
+        # paper authors note that they only trained models that have 2^a heads for some a.
+        # This has beneficial properties related to input being power of 2.
+
+        # Closest power of 2 below is workaround for when num of heads is not power of 2
```
ebsmothers, line 268: Their method of interpolating is a bit unusual. Maybe explicitly explain that for $num \textunderscore heads=2^N + k$ they are splicing the geometric series with ratio $2^{-\frac{8}{N}}$ with the first $2k$ elements of the geometric series with ratio $2^{-\frac{8}{N+1}}$ (assuming I am even understanding it correctly 😅)

PR #465, opened 2023-09-09
add SimpleRMSNorm

lessw2020:
Summary:
This PR implements SimpleRMSNorm, as proposed in:
Scaling TransNormer to 175 Billion Parameters
and is:
"
In TransNormerLLM, we replace the RMSNorm with a new simple normalization
function called SimpleRMSNorm, abbreviated as SRMSNorm:
SRMSNorm(x) = x / ∥x∥2/√d

We empirically find that using SRMSNorm does not lead to any performance loss, as demonstrated in
the ablation study [below]:

Norm Type Params Updates Loss PPL
SRMSNorm 385M 100K 2.247 4.765
RMSNorm 385M 100K 2.247 4.766
LayerNorm 385M 100K 2.247 4.765
"

note that their architecture is not a TransFormer but a TransNormer...therefore, I tested this on gpt2 transformer and saw equivalent results between LayerNorm and SimpleRMSNorm as below:

simpleRMS_gpt2

In addition, SimpleRMSNorm is ~ 34% faster vs regular RMSNorm (eager mode comparison). 

Test plan:
Tested on GPT2 training as shown above, and have added 4 unit tests (2 for BF16 and 2 for FP32 dtypes).

### DIFF
--- a/tests/modules/layers/test_normalizations.py
+++ b/tests/modules/layers/test_normalizations.py
@@ -11,6 +11,7 @@ from torchmultimodal.modules.layers.normalizations import (
     Fp32GroupNorm,
     Fp32LayerNorm,
     RMSNorm,
+    SimpleRMSNorm,
 )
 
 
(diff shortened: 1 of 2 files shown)


torchmultimodal/modules/layers/normalizations.py
```diff
@@ -72,3 +72,24 @@ def _norm(self, x: Tensor) -> Tensor:
     def forward(self, x: Tensor) -> Tensor:
         x_normed = self._norm(x.float()).type_as(x)
         return x_normed * self.scale
+
+
+class SimpleRMSNorm(nn.Module):
+    """Simple RMSNorm
+
+    SRMSNorm(x) = x / ∥x∥2/√d
```
rohan-varma, line 80: Can we rewrite this with maybe paranthesis to show more clarity?

tests/modules/layers/test_normalizations.py
```diff
@@ -61,3 +62,63 @@ def test_rms_norm_core_algo():
...
+            120.2589,
+        ],
+        dtype=torch.float32,
+    )
+
+    expected_output_bf16_ones = torch.tensor(
+        [1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0],
+        dtype=torch.bfloat16,
+    )
+    expected_output_fixed = torch.tensor(
```
rohan-varma, line 95: curious, how do we get these expected outputs? If we get them by running the function itself, I'm not sure how that provides confidence on the correctness? cc @ebsmothers

rohan-varma (approved):
stamped, but is that self.scaling supposed to be in the denom or the top? I can't quite tell based on how the equation is written in the docstr

torchmultimodal/modules/layers/normalizations.py
lessw2020: added parens to make it clear.

tests/modules/layers/test_normalizations.py
lessw2020: This is a valid point, but note that for previous test cases where I did this (effectively compute something in two different ways where the end result should be the same) got pushback that doing that was 'weird' and should just use constant inputs and expected outputs. 

Regardless - In this case I've established it works as expected via a model training run (i.e actual use testing), and from there locked in the expected outputs using two known inputs.

ebsmothers: Sorry, just getting to this now. In this case I would say that we can actually hand compute the expected output for some simple inputs. Not sure that can be done super easily with the existing `input_fixed_fp32` but for something like e.g. [3.0, 4.0, 3.0, 4.0, 3.0, 4.0, 3.0, 4.0] it'd be nice and easy. Tbh not a big deal either way

torchmultimodal/modules/layers/normalizations.py
```diff
@@ -72,3 +72,25 @@ def _norm(self, x: Tensor) -> Tensor:
     def forward(self, x: Tensor) -> Tensor:
         x_normed = self._norm(x.float()).type_as(x)
         return x_normed * self.scale
+
+
+class SimpleRMSNorm(nn.Module):
+    """Simple RMSNorm
+
+    SRMSNorm(x) = (x / ∥x∥2) /√d
```
ebsmothers, line 80: Should be this, right?
```suggestion
    SRMSNorm(x) = x / (∥x∥2 /√d)
```

ebsmothers (approved):
LGTM, my only (small) comment of substance is fixing the parentheses in the docstring

PR #477, opened 2023-10-02
add AnyPrecision optimizer

lessw2020:
Summary:
This PR brings the AnyPrecison_AdamW optimizer over from torch.distx lib.  Users have complained that torch.distx install requires building c++ for meta tensor and that lib is now very out of date...thus they were blocked from installing and using AnyPrecision, unless hand importing. 

For reference - AnyPrecision optimizer adds a couple features over AdamW:
a - direct control over the dtype for storing the variance and momentum states of AdamW
b - adds Kahan summation option - when active a compenstation buffer is maintainted such that loss of precision from using  bfloat16 is rolled into the compensation buffer and eventually added back during a future weight update.  Thus, this enables closer to fp32 precision for weight updates while still maintaining optimizer states in bf16. 

Test plan:
There are two unit tests - one that runs AnyPrecision with same states as AdamW to simple ensure algorithmic correctness.
The second tests runs AnyPrecision with bfloat16 for variance and momentum and confirms that AnyPrecision states are being run in bf16. 
Note that the second test requires GPU due to a bug in PyTorch CPU not supporting bfloat for one of the operations. 

Experiments with AnyPrecision:
I did a comparison training run using AnyPrecision with variance in Bfloat16 and compared that to AdamW using the normal fp32 for variance. (gpt2, 200M).
As largely expected, AnyPrecision produced equal/slightly better results while reducing memory usage by 2%+ due to the variance being in bf16 (this % will increase with a larger model size).
anyprecision_var16


tests/modules/layers/test_anyprecision_optimizer.py
```diff
@@ -0,0 +1,98 @@
...
+            for p1, p2 in zip(model.parameters(), model_clone.parameters()):
+                assert_expected(p1, p2)
+
+    def test_adam_equivalence(self, device="cpu"):
+        """
+        Tests that AnyPrecisionAdamW is equivalent to AdamW when
+        kahan summation and different dtypes for momentum, variance,
+        and compensation buffer are turned off (i.e. all float32).
+        """
+        if device == "cuda" and not torch.cuda.is_available():
```
ebsmothers, line 62: Sorry maybe I'm missing something: when would device ever not be cpu in the current setup?

tests/modules/layers/test_anyprecision_optimizer.py
```diff
@@ -0,0 +1,98 @@
...
+            anyprecision_adam.zero_grad()
+            inp = torch.randn(5, 5, device=next(simple_model.parameters()).device)
+            simple_model(inp).sum().backward()
+            anyprecision_adam.step()
+
+        for group in anyprecision_adam.param_groups:
+            for p in group["params"]:
+                state = anyprecision_adam.state[p]
+                assert state["exp_avg"].dtype == torch.bfloat16
+                assert state["exp_avg_sq"].dtype == torch.bfloat16
```
ebsmothers, line 98: It might also be useful to add some test case related to Kahan summation. Not a huge deal either way, but as is it's not well-covered

ebsmothers (commented):
Thanks for adding this! Overall looks great, a few (mostly minor) comments

torchmultimodal/modules/optimizers/anyprecision.py
```diff
@@ -0,0 +1,185 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+# AnyPrecisionAdamW: a flexible precision AdamW optimizer
```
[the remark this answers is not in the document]
lessw2020: agree - moved it to the docstring.

tests/modules/layers/test_anyprecision_optimizer.py
lessw2020: I was hoping gpu would be added based on triton kernels being included (which requires gpu). 
I was also hitting an issue running on cpu due to an upcast error. 
I've reworked the tests to now function on cpu and avoid the upcasting issue, and opened this issue in PT to get the cpu upcast issue resolved for the future:
https://github.com/pytorch/pytorch/issues/110905

I've thus removed the gpu_test from this one and the new kahan_summation test, so we are now 100% running on cpu. 
anyPrecision_3Pass

tests/modules/layers/test_anyprecision_optimizer.py
lessw2020: ha - I was literally thinking this last week that I wasn't expressly checking the kahan aspect, so definitely agree.
I was hitting cpu issues with gradients and bfloat16 / fp32 interactions which was another reason I had held off, but I've worked around that now and opened an issue to get the cpu upcast fixed. 

Anyway, now has an express kahan summation test that activates kahan summation, runs an iteration and checks the internal compensation buffer for expected results.

torchmultimodal/modules/optimizers/anyprecision.py
```diff
@@ -0,0 +1,185 @@
...
+                # adjust using bias1
+                bias_correction1 = 1 - beta1**step
+
+                step_size = lr / bias_correction1
+
+                # adjust using bias2
+                denom_correction = (1 - beta2**step) ** 0.5  # avoids math import
+
+                centered_variance = (exp_avg_sq.sqrt() / denom_correction).add_(
+                    eps, alpha=1
```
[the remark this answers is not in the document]
lessw2020: yes it is - I believe I was trying to be explicit with the alpha here, but I see that earlier I used add_ and didn't call alpha out, so min it should be consistent across all use, and thus I have removed the alpha=1.

torchmultimodal/modules/optimizers/anyprecision.py
```diff
@@ -0,0 +1,185 @@
...
+                parameter groups
+            lr (float, optional): learning rate (default: 1e-3)
+            betas (Tuple[float, float], optional): coefficients used for computing
+                running averages of gradient and its square (default: (0.9, 0.999))
+            eps (float, optional): term added to the denominator to improve
+                numerical stability (default: 1e-8)
+            weight_decay (float, optional): weight decay coefficient (default: 1e-2)
+
+            # Any Precision specific
+            use_kahan_summation = creates auxiliary buffer to ensure high precision
```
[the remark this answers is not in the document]
lessw2020: good idea.  I've added a link and the very first few lines summarizing how Kahan summation works so it's visible right in the doc string:
"       Kahan summation overview:  
        https://en.wikipedia.org/wiki/Kahan_summation_algorithm

        "In numerical analysis, the Kahan summation algorithm, also known as compensated 
        summation, significantly reduces the numerical error in the total obtained by 
        adding a sequence of finite-precision floating-point numbers, compared to the 
        obvious approach. This is done by keeping a separate running compensation 
        (a variable to accumulate small errors), in effect extending the precision of 
        the sum by the precision of the compensation variable.  "

torchmultimodal/modules/optimizers/anyprecision.py
```diff
@@ -0,0 +1,185 @@
...
+        lr: float = 1e-3,
+        betas: Tuple[float, float] = (0.9, 0.999),
+        eps: float = 1e-8,
+        weight_decay: float = 0.0,
+        use_kahan_summation: bool = False,
+        momentum_dtype: torch.dtype = torch.float32,
+        variance_dtype: torch.dtype = torch.bfloat16,
+        compensation_buffer_dtype: torch.dtype = torch.bfloat16,
+    ) -> None:
+        """
```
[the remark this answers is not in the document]
lessw2020: good point, this will help ensure people don't think it's two different impls. 
I've added a note and pointer to the original TorchDistX so people know it's the same, just moved for easier access.

tests/modules/layers/test_anyprecision_optimizer.py
```diff
@@ -0,0 +1,165 @@
...
+            weight_decay=weight_decay,
+        )
+
+        # Verify params are equal initially
+        model_orig_params = [p.clone() for p in model.parameters()]
+        for p1, p2 in zip(model_clone.parameters(), model_orig_params):
+            assert_expected(p1, p2)
+
+        for i in range(6):
+            adam_opt.zero_grad()
```
rohan-varma, line 45: nit: can we test both settings of `set_to_none`?

tests/modules/layers/test_anyprecision_optimizer.py
```diff
@@ -0,0 +1,165 @@
...
+                assert_expected(p1, p2)
+
+    def test_adam_equivalence(self, device="cpu"):
+        """
+        Tests that AnyPrecisionAdamW is equivalent to AdamW when
+        kahan summation and different dtypes for momentum, variance,
+        and compensation buffer are turned off (i.e. all float32).
+        """
+        if device == "cuda" and not torch.cuda.is_available():
+            # raise unittest.SkipTest("CUDA not available")
```
rohan-varma, line 68: nit: remove

tests/modules/layers/test_anyprecision_optimizer.py
```diff
@@ -0,0 +1,165 @@
...
+            pytest.skip(reason="CUDA not available")
+
+        model = nn.Sequential(nn.Linear(5, 5), nn.Linear(5, 5), nn.Linear(5, 5))
+        if device == "cuda":
```
rohan-varma, line 72: Can we have CUDA test for now even without triton kernel, just to ensure things work on GPU where this will normally be used?

tests/modules/layers/test_anyprecision_optimizer.py
```diff
@@ -0,0 +1,165 @@
...
+            nn.Linear(5, 10, dtype=torch.bfloat16),
+            nn.GELU(),
+            nn.Linear(10, 2, dtype=torch.bfloat16),
+        )
+        simple_model.to(torch.bfloat16)
+        simple_model.to(device)
+
+        anyprecision_adam = AnyPrecisionAdamW(
+            simple_model.parameters(),
+            variance_dtype=torch.bfloat16,
```
rohan-varma, line 91: What happens when these dtypes mismatch the model dtype? Do we throw the appropriate error?

tests/modules/layers/test_anyprecision_optimizer.py
```diff
@@ -0,0 +1,165 @@
...
+            momentum_dtype=torch.bfloat16,
+            use_kahan_summation=True,
+            compensation_buffer_dtype=torch.bfloat16,
+        )
+
+        expected_kahan_buffer = torch.tensor(
```
rohan-varma, line 126: how did we get this buffer?

tests/modules/layers/test_anyprecision_optimizer.py
```diff
@@ -0,0 +1,165 @@
...
+                [2.2888e-05, 9.5367e-04, 3.0518e-05, -9.4604e-04, -2.2888e-05],
+                [-9.5367e-04, -9.5367e-04, 2.2888e-05, -9.5367e-04, -3.0518e-05],
+                [9.5367e-04, -9.5367e-04, -9.4604e-04, 9.5367e-04, 2.2888e-05],
+                [9.4604e-04, -2.2888e-05, 9.4604e-04, 9.5367e-04, 2.2888e-05],
+            ],
+            dtype=torch.bfloat16,
+            device=device,
+        )
+
+        for i in range(1):
```
rohan-varma, line 143: why do we need a loop if we're just iterating once?

tests/modules/layers/test_anyprecision_optimizer.py
```diff
@@ -0,0 +1,165 @@
...
+            )
+            simple_model(inp).sum().backward()
+            anyprecision_adam.step()
+
+            for group in anyprecision_adam.param_groups:
+                for index, p in enumerate(group["params"]):
+                    state = anyprecision_adam.state[p]
+                    pcomp = state["compensation"]
+                    assert pcomp.dtype == torch.bfloat16
+                    if index == 0:
```
rohan-varma, line 159: why only for index 0? what happens with the other indices?

torchmultimodal/modules/optimizers/anyprecision.py
```diff
@@ -0,0 +1,199 @@
...
+        self,
+        params: params_t,
+        lr: float = 1e-3,
+        betas: Tuple[float, float] = (0.9, 0.999),
+        eps: float = 1e-8,
+        weight_decay: float = 0.0,
+        use_kahan_summation: bool = False,
+        momentum_dtype: torch.dtype = torch.float32,
+        variance_dtype: torch.dtype = torch.bfloat16,
+        compensation_buffer_dtype: torch.dtype = torch.bfloat16,
```
rohan-varma, line 28: only A100 and more recent support bf16, so should we choose defaults that are more hardware agnostic?

torchmultimodal/modules/optimizers/anyprecision.py
```diff
@@ -0,0 +1,199 @@
...
+        """
+        AnyPrecisionAdamW: a flexible precision AdamW optimizer
+        with optional Kahan summation for high precision weight updates.
+        Allows direct control over momentum, variance and auxiliary compensation
+        buffer dtypes.
+        Optional Kahan summation is used to offset precision reduction for
+        the weight updates. This allows full training in BFloat16 (can be equal or
+        better than FP32 results in many cases) due to high precision weight upates.
+
+        This optimizer is the same AnyPrecision that was previously residing
```
rohan-varma, line 39: nit: I don't think this part of the doc is really needed, plus we don't wanna advertise torchdistX much.

rohan-varma (commented):
LGTM overall, some minor comments. Will stamp once @ebsmothers comments are also resolved, thanks for doing this!

torchmultimodal/modules/optimizers/anyprecision.py
ebsmothers: Ah sorry this change is actually based on a [previous comment](https://github.com/facebookresearch/multimodal/pull/477#discussion_r1349370152) of mine 😅 . Personally I am fine either way here

tests/modules/layers/test_anyprecision_optimizer.py
ebsmothers: Yeah a short comment like "values determined by comparison to reference implementation" (assuming that's what you did) should suffice

ebsmothers (commented):
Thanks for addressing my comments, looks good from my side! Given that @rohan-varma has a few more comments here, will leave the final sign-off to him.

torchmultimodal/modules/optimizers/anyprecision.py
lessw2020: I'd prefer to leave this comment mostly to avoid confusion (i.e. if someone finds there are two AnyPrecision within the larger PT repo or external references ala HF to the TorchDistX AnyPrecision).

torchmultimodal/modules/optimizers/anyprecision.py
lessw2020: hmm - BF16 is pretty mainstream now.  
The Ampere architecture supports native BF16, and has been around since 2020 and that includes datacenter A2, A10, A100 as well as consumer 2050 (mobile), 3050 - 3090...4090 range.   
Hopper of course supports it as well though that is not common atm.
For the compensation buffer and variance though, FP16 should be fine as we don't really need the full range of BF16 and the extra precision of FP16 could be handy.

Ok will switch over to FP16 as the default then, but will need to update the unit tests to match.

lessw2020: well nvm - we need to stay with BF16.
In running the unit tests using FP16 and inspecting the compensation buffer as fp16, started seeing 'nans' which I did not expect but will defeat the purpose of the compensation buffer.
Anyway, will stick with BF16 and based on how mainstream BF16 is at this point don't think it should be an issue.

tests/modules/layers/test_anyprecision_optimizer.py
lessw2020: the other buffers were smaller but I have extended the check to all 4 compensation buffers so no longer have this index==0.

tests/modules/layers/test_anyprecision_optimizer.py
lessw2020: I've extended the test do to 2 loops (which was the original reason I had the loop) and then check the buffers and dtype afterwards. 
so the loop is relevant now. :)

tests/modules/layers/test_anyprecision_optimizer.py
lessw2020: added comment that these are pre-computed buffers determined by comparison to reference impl.

tests/modules/layers/test_anyprecision_optimizer.py
lessw2020: removed.

tests/modules/layers/test_anyprecision_optimizer.py
lessw2020: The variance dtype and momentum dtype are totally independent of the model dtype. 
Thus, by design, nothing happens :)  
There's no linkage at all between the model/param dtypes and the optimizer states dtypes.

tests/modules/layers/test_anyprecision_optimizer.py
lessw2020: this is a great catch - yes I have updated the test to toggle based on even/odd of the index to use set_to_none = True/False. 
All passing.

tests/modules/layers/test_anyprecision_optimizer.py
lessw2020: I've expanded the tests to all have both cpu and gpu (if available) equivalents.   Note that there are slight rounding diffs between gpu and cpu so I had to make different reference Kahan comp buffers to keep the assert_expected tolerances tight. 
cpu_gpu_6_anyprecision

rohan-varma (approved):
great, thanks so much for resolving all our comments!

PR #479, opened 2023-10-04
Add pretrained MAE weights, option to load checkpoints in ViT builder

ebsmothers:
Summary:
For MAE fine-tuning, fine-tuning occurs just on the encoder (ViT). This change allows easy loading of MAE pretrained weights directly into our ViT class.

Test plan:
```
python -m pytest -v tests/models/*
...
========== 207 passed, 25 warnings in 424.67s (0:07:04) ===========================

python -m pytest -v tests/modules/*
...
======================== 192 passed, 2 skipped, 22 warnings in 10.75s ==========================
```

Test instantiating ViT using MAE pretrained weights for each of the 3 checkpoints:

Screenshot 2023-10-05 at 6 39 02 PM

### DIFF
--- a/torchmultimodal/models/masked_auto_encoder/model.py
+++ b/torchmultimodal/models/masked_auto_encoder/model.py
@@ -13,6 +13,11 @@ from torchmultimodal.models.masked_auto_encoder.position_embeddings import (
     get_2d_sin_cos_embeddings,
 )
 from torchmultimodal.models.masked_auto_encoder.swin_decoder import SwinTransformer
+from torchmultimodal.modules.encoders.vision_transformer import (
+    VisionTransformer,
+    vit_b_16,
+    vit_l_16,
+)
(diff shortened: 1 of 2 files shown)


torchmultimodal/models/masked_auto_encoder/model.py
```diff
@@ -20,6 +20,13 @@
 )
 
 
+MAE_MODEL_MAPPING = {
```
ankitade, line 23: isnt it nicer to expose as vit_mae* wrapper with pretrained=True like clip in here itself vs making the user pass it around

ebsmothers: Yeah that's fine too. I thought it was a bit weird cause we are then mixing builders across files (i.e. we either have a builder for MAE inside vision_transformer.py or a builder returning ViT inside MAE model.py). But maybe the second option (I think that's what you're suggesting?) isn't so bad

PR #488, opened 2023-10-13
Refactor ADM classes

ebsmothers:
* #490
* #489
* __->__ #488

Differential Revision: [D50288361](https://our.internmc.facebook.com/intern/diff/D50288361)

### DIFF
--- a/torchmultimodal/diffusion_labs/models/adm_unet/adm.py
+++ b/torchmultimodal/diffusion_labs/models/adm_unet/adm.py
@@ -96,7 +96,7 @@ class ADMUNet(nn.Module):
             Expected shape of tensors are [b, c], where c is the embedding dim of the Tensor.
     """
 
-    DEFAULT_EMBED_NAME = "clip_image"
+    DEFAULT_EMBED_NAME = "context"
 
     def __init__(
         self,
--- a/torchmultimodal/diffusion_labs/models/adm_unet/adm.py
+++ b/torchmultimodal/diffusion_labs/models/adm_unet/adm.py
@@ -24,8 +24,10 @@ from torchmultimodal.modules.layers.position_embedding import (
     SinusoidalPositionEmbeddings,
 )
 
+DEFAULT_EMBED_NAME = "clip_image"
 
-class ADM(nn.Module):
+
+class ADMUNet(nn.Module):
     """Ablated Diffusion Model as described in "Diffusion Models Beat GANs on Image Synthesis" (https://arxiv.org/abs/2105.05233)
 
     ADM consists of three major components:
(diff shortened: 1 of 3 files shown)


ebsmothers:
@ebsmothers has imported this pull request.  If you are a Meta employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D50288361).

ebsmothers:
@ebsmothers has imported this pull request.  If you are a Meta employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D50288361).

torchmultimodal/diffusion_labs/models/adm_unet/adm.py
```diff
@@ -24,25 +24,55 @@
     SinusoidalPositionEmbeddings,
 )
 
+DEFAULT_EMBED_NAME = "clip_image"
```
pbontrager, line 27: I think this should be a parameter for ADMUnet instead of a global variable. If you look at the transform and adapter classes they all take the conditional input key as "*_field", for this case it could be "conditional_field". Also I don't think the default name should be based on CLIP as that's a dalle input and not generic to ADM. It should be something generic like "context" or "condition".

torchmultimodal/diffusion_labs/models/adm_unet/adm.py
```diff
@@ -68,132 +98,6 @@ class ADM(nn.Module):
...
-
-    Args:
-        x (Tensor): input Tensor of shape [b, in_channels, h, w]
-        res_conditional_embedding (Tensor): conditioning embedding vector of shape [b, dim_res_cond].
-        attn_conditional_embedding (Tensor): inputs to be used in cross attention of shape [b, n, dim_attn_cond] where n is the
-            number of tokens.
-    """
-
     def __init__(
         self,
```
pbontrager, line 102: We should add "*" for forced named parameters

torchmultimodal/diffusion_labs/models/adm_unet/adm.py
```diff
@@ -378,7 +346,65 @@ def forward(
             if hidden_states:
                 h = torch.cat([h, hidden_states.pop()], dim=1)
             h = block(h, res_conditional_embedding, attn_conditional_embedding)
-        return h
+
+        # If model is predicting variance, then it should be configured to output double the channels as input
```
pbontrager, line 350: This should probably be broken out as it's own method (_get_variance_value) to clean up the forward method a bit. If predict_variance_value is false it can return None. Then you can just return DiffusionOutput with both prediction and variance_value as the default variance_value is None anyway.

torchmultimodal/diffusion_labs/models/adm_unet/adm.py
```diff
@@ -378,7 +346,65 @@ def forward(
...
+            return DiffusionOutput(
+                prediction=prediction,
+                variance_value=variance_value,
+            )
+        else:
+            return DiffusionOutput(
+                prediction=h,
+            )
+
+    def _get_conditional_projections(
```
pbontrager, line 369: nit I think it's better to have forward as the last method in a nn.Module so it's easier to find.

torchmultimodal/diffusion_labs/models/adm_unet/adm.py
```diff
@@ -453,7 +479,7 @@ def adm_unet(
     time_embed_dim: int = 512,
```
pbontrager, line 479: Do all of these parameters make sense here anymore? Since a user can just call ADMUnet if they want to change many defaults. Maybe just expose "embedding_name", "image_channels", and "predict_variance"

torchmultimodal/diffusion_labs/models/adm_unet/adm.py
```diff
@@ -453,7 +479,7 @@ def adm_unet(
...
     cond_embed_dim: int = 2048,
     clip_embed_dim: int = 768,
```
pbontrager, line 481: These two variables should be just "embed_*" without clip

torchmultimodal/diffusion_labs/models/adm_unet/adm.py
```diff
@@ -453,7 +479,7 @@ def adm_unet(
...
-    clip_embed_name: str = "clip_image",
+    clip_embed_name: str = DEFAULT_EMBED_NAME,
     predict_variance_value: bool = True,
     # ADMUNet args
     image_channels: int = 4,
```
pbontrager, line 485: Can't comment lower than this line. But on line 490 I think this should be: constructs a conditional ADM U-Net

torchmultimodal/diffusion_labs/models/adm_unet/adm.py
```diff
@@ -202,10 +106,40 @@ def __init__(
         use_attention_for_layer: List[bool],
         dim_res_cond: int,
         dim_attn_cond: Optional[int] = None,
+        embed_dim: Optional[int] = None,
         in_channels: int = 3,
         out_channels: int = 3,
+        time_embed_dim: Optional[int] = None,
+        timestep_encoder: Optional[nn.Module] = None,
```
pbontrager, line 113: I'm not sure if it makes sense to make these three encoders/projectors a parameter. It makes the init much more messy. I'd be in favor of removing that option.

pbontrager (commented):
Looks great overall and structurally good. I left some comments around name and parameters.

torchmultimodal/diffusion_labs/models/adm_unet/adm.py
pbontrager: I just mean, that if the purpose of this builder is just to house the defaults used in ADM then I don't think we should allow the user to override all of the defaults. If they need custom parameters they can use ADMUnet. If they want the ADM defaults then they can use this functions with just the three parameters exposed I mentioned. I'm also okay with moving these defaults to ADMUnet and dropping this builder.

torchmultimodal/diffusion_labs/models/adm_unet/adm.py
ebsmothers: After our offline discussion, I plan to leave these in for this version to unblock other changes. But let's remove in a follow-up

torchmultimodal/diffusion_labs/models/adm_unet/adm.py
ebsmothers: I see. Maybe let's do that in a follow-up then? I think dalle2_decoder is using a bunch of these params rn so making this change would have implications for that

torchmultimodal/diffusion_labs/models/adm_unet/adm.py
ebsmothers: Updated to context, left clip_image in the builder. Lmk if that's what you have in mind

ebsmothers:
@ebsmothers has imported this pull request.  If you are a Meta employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D50288361).

torchmultimodal/diffusion_labs/models/adm_unet/adm.py
pbontrager: CLIP is a dalle2 input and is not specific to ADM. So I think we should remove the name from everywhere. Also I still don't believe this should be a global variable.

torchmultimodal/diffusion_labs/models/adm_unet/adm.py
pbontrager: These defaults match dalle2 decoder. But you're right we should update dalle2 decoder then to be simpler.

torchmultimodal/diffusion_labs/models/adm_unet/adm.py
```diff
@@ -471,8 +507,8 @@ def adm_unet(
     Args:
         time_embed_dim (int): desired dimensionality of timestep embedding
         cond_embed_dim (int): desired dimensionality of conditional input embeddings
-        clip_embed_dim (int): expected dimensionality of CLIP image embeddings
-        clip_embed_name (str): name of CLIP embedding conditional input
+        embed_dim (int): expected dimensionality of CLIP image embeddings
```
pbontrager, line 510: The description here shouldn't say CLIP either. It's the expected dim of the conditional image embedding

ebsmothers:
@ebsmothers has imported this pull request.  If you are a Meta employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D50288361).

torchmultimodal/diffusion_labs/models/adm_unet/adm.py
ebsmothers: Ditched the global variable and used "context" for default in both the ADMUNet class and the builder

pbontrager (approved):
It looks good now. Thanks for merging these two classes!

PR #494, opened 2023-10-17
[bugfix] Fix missing self in res/attn cond proj definitions

ebsmothers:
Summary:
Fixing a bug where attn_cond_proj and res_cond_proj were not being set. Also fix a bug where we were not passing embed_name from adm_unet builder

Test plan:
Run diffusion_labs unit tests


pbontrager (approved):
Thanks for fixing this bug.

PR #504, opened 2023-10-24
Remaining fixes to unblock nightly build

ebsmothers:
Two more changes to fix our nightly builds:

1) Add \_\_init\_\_.py to `torchmultimodal/models/video_gpt/` to fix import failures.
2) Define a new test util to skip unit tests if ffmpeg is not installed. Per discussions with @mthrok torchaudio does not come with ffmpeg installed, though when we use conda install it will be included with torchvision. As a result, the pip install command in our nightly builds does not have ffmpeg, which causes some of our audio MAE transform tests to fail. We also do not want to install it directly with our library due to licensing. As a result, we add a test utility to skip these tests if ffmpeg is not installed.

Test plan:
Create a new conda env and follow the steps in nightly_build.yaml to install our deps with pip (so no ffmpeg).

```
conda create -n tmm-10-24-23-no-ffmpeg python=3.8
conda activate tmm-10-24-23-no-ffmpeg
python -m pip install --pre torch torchvision torchaudio --extra-index-url https://download.pytorch.org/whl/nightly/cpu
python -m  pip install -r requirements.txt
python -m pip install pytest pytest-mock pytest-cov
python -m pytest -v tests/transforms
...
============ 58 passed, 7 skipped in 2.79s =========
```

Also confirmed that by removing one of the `skip_if_no_mmpeg` decorator from one of the tests in the same environment causes it to fail.

### DIFF
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -11,6 +11,8 @@ from contextlib import contextmanager
 from pathlib import Path
 from typing import Any, Dict, NamedTuple, Optional, Tuple, Union
 
+import numpy as np
+
 import pytest
 import torch
 import torch.distributed as dist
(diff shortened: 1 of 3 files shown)


rohan-varma (approved):

tests/test_utils.py
```diff
@@ -286,3 +286,10 @@ def fixed_init_model(
                 dtype=param.dtype,
             )
         )
+
+
+def skip_if_no_ffmpeg(message="Requires ffmpeg"):
+    import importlib
+
+    ffmpeg_spec = importlib.util.find_spec("ffmpeg")
+    return pytest.mark.skipif(ffmpeg_spec is None, reason=message)
```
mthrok, line 295: `ffmpeg` is not a Python package, so this will return True even when a proper FFmpeg installation is found in the system.

PR #506, opened 2023-11-01
Add CoCa model

ebsmothers:
* #507
* __->__ #506

Differential Revision: [D51332627](https://our.internmc.facebook.com/intern/diff/D51332627)

### DIFF
--- a/torchmultimodal/models/coca/coca_model.py
+++ b/torchmultimodal/models/coca/coca_model.py
@@ -415,7 +415,8 @@ class CoCaForPretraining(nn.Module):
         super().__init__()
         self.model = model
         self.contrastive_loss = ContrastiveLossWithTemperature(
-            logit_scale_min=logit_scale_min, logit_scale_max=logit_scale_max
+            logit_scale_min=contrastive_logit_scale_min,
+            logit_scale_max=contrastive_logit_scale_max,
         )
         self.caption_loss = nn.CrossEntropyLoss(ignore_index=pad_idx)
 
--- a/torchmultimodal/models/coca/coca_model.py
+++ b/torchmultimodal/models/coca/coca_model.py
@@ -4,6 +4,8 @@
 # This source code is licensed under the BSD-style license found in the
 # LICENSE file in the root directory of this source tree.
 
+import math
+from functools import partial
 from typing import Callable, Dict, List, NamedTuple, Optional, Tuple, Union
 
 import torch
@@ -83,6 +85,10 @@ class CoCaModel(nn.Module):
         # Extract image embeddings
         if isinstance(vision_encoder_outs, TransformerOutput):
             image_embeddings = vision_encoder_outs.last_hidden_state
+        elif isinstance(vision_encoder_outs, tuple):
+            vision_encoder_outs = vision_encoder_outs[0]
+            assert isinstance(vision_encoder_outs, Tensor)
+            image_embeddings = vision_encoder_outs
         else:
             assert isinstance(vision_encoder_outs, Tensor)
             image_embeddings = vision_encoder_outs
@@ -142,7 +148,6 @@ def coca_vit(
     fusion_n_layer: int,
     fusion_n_head: int,
     fusion_dim_feedforward: int,
-    fusion_output_dim: int,
     # Required attention pooler args
     pooler_input_embed_dim: int,
     pooler_output_embed_dim: int,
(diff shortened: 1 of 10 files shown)


torchmultimodal/models/coca/multimodal_decoder.py
```diff
@@ -0,0 +1,101 @@
...
+import torch
+from torch import nn, Tensor
+from torchmultimodal.modules.layers.transformer import TransformerDecoder
+from torchmultimodal.utils.attention import get_causal_attention_mask
+
+
+class CoCaMultimodalDecoder(nn.Module):
+    """
+    Multimodal decoder with cross-attention for CoCa model.
+    Based on the implementation in open_clip: https://tinyurl.com/mn35vdmd
```
pbontrager, line 18: It would be useful to have a brief description of the architecture and use cases for this model here. Is the CoCA decoder a variant on a more general architecture?

torchmultimodal/models/coca/text_decoder.py
```diff
@@ -0,0 +1,252 @@
...
+import torch
+import torch.nn.functional as F
+from torch import nn, Tensor
+from torchmultimodal.modules.layers.transformer import TransformerDecoder
+from torchmultimodal.utils.attention import get_causal_attention_mask
+
+
+class CoCaTextEmbeddings(nn.Module):
+    """
+    Text embeddings for CoCa model. Includes token embeddings, positional embeddings,
```
pbontrager, line 18: Same comment as multimodal decoder though probably not as necessary here

torchmultimodal/modules/layers/attention_pooler.py
```diff
@@ -0,0 +1,100 @@
...
+from torch import nn, Tensor
+from torchmultimodal.modules.layers.multi_head_attention import (
+    MultiHeadAttentionWithCache,
+)
+
+
+class AttentionPooler(nn.Module):
+    """
+    Attention pooling layer: pools inputs to sequence length n_queries by performing
+        cross-attention with learned query embeddings. Based on the CoCa implementation
```
pbontrager, line 19: Did CoCA invent this or just use it? It'd be better if we could point to the original abstract if this is meant as a general layer.

torchmultimodal/models/coca/coca_model.py
```diff
@@ -0,0 +1,503 @@
...
+        """
+
+        # Image encoder
+        vision_encoder_outs = self.vision_encoder(images)
+
+        # Extract image embeddings
+        if isinstance(vision_encoder_outs, TransformerOutput):
+            image_embeddings = vision_encoder_outs.last_hidden_state
+        elif isinstance(vision_encoder_outs, tuple):
+            vision_encoder_outs = vision_encoder_outs[0]
```
pbontrager, line 89: This seems very risky. In general isinstance is very risky, but here especially since the user doesn't know without looking at the code that it's assumed that output 0 is the only thing being used.

pbontrager (approved):
This looks good as an implementation for CoCA and as a tested internal model. I think it would be helpful to include more context for this model and that it was already tested and built internally in the PR description. Otherwise no issues with how it's being moved to open. I left a few nit comments that would be nice to have but not blocking.

torchmultimodal/modules/layers/attention_pooler.py
ebsmothers: Good point. I think it is originally from the set transformer paper, will add a link.

torchmultimodal/models/coca/multimodal_decoder.py
ebsmothers: It is pretty much just a generic transformer decoder with cross-attention and optional output projection. I can add more detail here though

torchmultimodal/models/coca/coca_model.py
ebsmothers: Yeah fair point. This is because a common return type from transformer encoders is a tuple (last_hidden_state, all_hidden_states). Maybe we could add an assert here? Something like 
```suggestion
            assert len(vision_encoder_outs) == 2 and isinstance(vision_encoder_outs[0], Tensor) and isinstance(vision_encoder_outs[1], List)
            vision_encoder_outs = vision_encoder_outs[0]
```

ebsmothers:
@pbontrager thanks for the review. Agree it's probably worth to add more context; actually this is why I created #507 to demonstrate parity with the open_clip implementation on ImageNet zero-shot (though I probably should have mentioned it on this PR 😅).

ebsmothers:
@ebsmothers has imported this pull request.  If you are a Meta employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D51332627).

PR #509, opened 2023-11-02
Fix contrastive loss with temperature test

ebsmothers:
For some reason when we run pytest on a subdirectory of tests/ PyTorch is unable to find GPUs if they are present. But when we run just `pytest tests/` PyTorch does see the GPU. 

This means that in the latter case the contrastive_loss_with_temperature test fails due to tensors on different devices (previously only model params were moved to the current device but the embeddings were not). But random initialization on GPU is different than on CPU so the test will give different results across different types of hardware.

Since the contrastive loss distributed tests are intended to only run on GPU, we leave those as is. But for the remaining test cases we explicitly move everything to CPU so we can ensure consistent results.

### DIFF
--- a/tests/modules/losses/test_contrastive_loss_with_temperature.py
+++ b/tests/modules/losses/test_contrastive_loss_with_temperature.py
@@ -25,7 +25,6 @@ from torch import distributed as dist, Tensor
 from torchmultimodal.modules.losses.contrastive_loss_with_temperature import (
     ContrastiveLossWithTemperature,
 )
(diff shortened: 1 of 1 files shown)


kartikayk (commented):
Have we spoken to dev infra about this behavior difference between pytest and pytests tests/

ebsmothers:
@kartikayk no we haven't. But this only manifests locally and not in our CI (since we don't have GPUs for our CI runs currently). So it may be outside their purview (but still it'd be interesting to understand why this happens)

PR #512, opened 2023-11-15
update readme

ebsmothers:


laurencer:
This is a great start. I would take a look at https://github.com/pytorch/rl and https://github.com/huggingface/peft as examples of READMEs here for inspiration on structure, style, etc.

One thing to consider is the user onboarding flow here. E.g. when someone lands on the repo what are they trying to do and how do you point them in the right direction. There's probably a couple of different flows:

1. I land and I want to quickly see the models and play around with them in a notebook
2. I want to just get an overview of what models and components are available to see if this is useful for me
3. <probably a lot more I'm not thinking about>

It may be useful to include some code examples (or whatever is going to be the primary way people use the library). E.g. I assume the most common case will be importing a component?

On a different note, the diffusion_labs callout feels a weird/uncohesive. It seems like we should just say we support diffusion and all the components/models are under diffusion_labs.

ebsmothers:
@laurencer I took another pass based on your comments, please let me know what you think.

README.md
```diff
@@ -1,11 +1,136 @@
...
+[![Downloads](https://static.pepy.tech/personalized-badge/torchmultimodal-nightly?period=total&units=international_system&left_color=blue&right_color=orange&left_text=Downloads%20(nightly))](https://pepy.tech/project/torchmultimodal-nightly)
+
 # TorchMultimodal (Beta Release)
 
+[**Models**](#models) | [**Example scripts**](#example-scripts) | [**Getting started**](#getting-started) | [**Code overview**](#code-overview) | [**Installation**](#installation) | [**Contributing**](#contributing) | [**License**](#license)
+
 ## Introduction
-TorchMultimodal is a PyTorch library for training state-of-the-art multimodal multi-task models at scale. It provides:
+**TorchMultimodal** is a PyTorch library for training state-of-the-art multimodal multi-task models at scale, including both content understanding and generative models. TorchMultimodal contains:
 - A repository of modular and composable building blocks (models, fusion layers, loss functions, datasets and utilities).
```
pbontrager, line 11: nit, but usually a list should be at least three or structured differently. It might be worth breaking out the full models  (e.g. CLIP, FLAVA) vs the building blocks as they're different offerings.

README.md
```diff
@@ -1,11 +1,136 @@
...
+    loss.backward()
+    optimizer.step()
+    lr_scheduler.step()
+  ```
+  </details>
+
+## Code overview
+
+### [torchmultimodal/diffusion_labs](https://github.com/facebookresearch/multimodal/tree/main/torchmultimodal/diffusion_labs)
+diffusion_labs contains components for building diffusion models and end-to-end training of those models. For more details on these components, see [diffusion_labs/README.md](https://github.com/facebookresearch/multimodal/blob/main/torchmultimodal/diffusion_labs/README.md).
```
pbontrager, line 124: It doesn't currently offer end to end training, and it's still an open question as to whether it will, so I'd remove that part.

pbontrager (approved):
Thank you for updating this. I love the new code examples. I just left two small comments for change but otherwise it's good to go.

